PR #254, opened 2026-05-05
test: bump charset-torture corpus pin to v0.12.0

jakebromberg:
## Summary

Re-vendors the WX-1 charset-torture corpus from `@wxyc/shared` v0.10.0 (sha `75a3395b…`) to v0.12.0 (sha `41a18c5c0a92d129ec4b575827b6874196bfb7591e4bdf237a918a5da2de7b66`) and bumps the drift-guard workflow's pinned values to match. The upstream corpus changes deliver the WX-2.2.1 `to_storage_form` charter (WXYC/wxyc-shared#100, WXYC/wxyc-etl#89).

Two upstream corpus changes:

- **`normalization`** — the two NFD entries (`café` decomposed, `ñ` decomposed) now expect the NFC precomposed form in `expected_storage`. LML's WX-1 detector tests storage round-trip using `entry["input"]` directly, so SQLite continues to preserve the input bytes verbatim — no behavior change here. The new `expected_storage` field codifies what `to_storage_form` is expected to do at the future write boundary; LML does not yet wire `to_storage_form` into `LibraryDB` writes (that's WX-3 scope).

- **`bidi_marks`** — the three Cf-bearing entries (U+200E LRM, U+200F RLM, U+202E RLO + U+202C PDF) now have non-null `expected_match_form` values (`helloworld`, `helloworld`, `reversed`). LML's existing detectors do not assert against `expected_match_form` — the SQLite leg (`tests/integration/test_charset_torture_library_db.py`) is a pure storage-byte round-trip + FTS5/LIKE search round-trip on the raw input, and the PG leg (`tests/integration/test_charset_torture_entity_identity.py`) is a pure `entity.identity` TEXT round-trip. All three bidi_marks entries pass cleanly under the new corpus, so no xfail entries are added to `SEARCH_XFAIL_INPUTS` or `PG_TEXT_XFAIL_INPUTS`. When LML grows a normalize-vs-match-form check (WX-2.2.2 wiring of `to_match_form` into `LibraryDB.search`), this audit gets revisited and `[lml:no-match-form]` xfails get added then.

## Drift-guard CI failure is expected

The Charset Corpus Drift workflow on this PR will fail until `@wxyc/shared` v0.12.0 publishes to npm — that's the contract of the drift guard, and it self-resolves once upstream lands. Do not try to fix it in this PR.

## Test plan

- [x] `uv run pytest` — full default suite (unit + integration, excluding `pg`/`external_api`): 1851 passed, 88 deselected, 8 xfailed (the existing 6 search-roundtrip xfails for emoji/ZWJ + 2 unrelated), 0 unexpected failures, 0 XPASSes.
- [x] `uv run pytest tests/integration/test_charset_torture_library_db.py -v` — all 3 bidi_marks search-roundtrip cases PASS, all 4 normalization cases (NFC + NFD for café and ñ) PASS, all 6 emoji/ZWJ cases XFAIL as expected.
- [x] `uv run ruff check .` — clean.
- [x] `uv run ruff format --check .` — clean.

## Refs

- Upstream corpus PR: WXYC/wxyc-shared#100
- WX-2.2.1 `to_storage_form` Rust impl: WXYC/wxyc-etl#89


jakebromberg:
Status update (tracker #386): the upstream blocker `@wxyc/shared` PR #100 has now merged (state MERGED — v0.12.0 corpus published). The drift-guard CI failure documented in this PR's body should self-resolve once v0.12.0 hits npm. Worth a re-run on this PR.

Issue #397, opened 2026-05-28
[Tracker] Metadata correctness on free-text flowsheet entries (now-playing)


jakebromberg:
Tracker for a cluster of metadata-correctness bugs on free-text flowsheet entries — rows with `album_id IS NULL` whose iOS now-playing card shows wrong or missing streaming links / release year. Four distinct failure modes observed this session (2026-05-27 → 2026-05-28); all touch project #32 critical surfaces (`lookup/orchestrator.py`, `apps/backend/services/metadata/`, V2 flowsheet shape, iOS V2 playlist).

| # | Case | Pattern | Status |
|---|---|---|---|
| 1 | "Joyous" by Pleasure → Sheryl Crow / "The Jolly Beggars" | iTunes Search non-deterministic; `_fetch_apple_music_url` blindly returned `results[0]` with no artist/track verification | Closed by #390 (merged + deployed 2026-05-28); #389 |
| 2 | "Hebebeb (Zrag)" by Noura Mint Seymali / Yenbett → Apple link goes to Tzenni; release_year shown as 2014 | Right artist + right track on **wrong album**; same track name exists on multiple of the artist's releases (Tzenni 2014, Yenbett 2025), iTunes ranks the older more-indexed one first, and #390 doesn't verify album | Closed by #398 (merged + deployed 2026-05-28); #396 |
| 3 | "The Four Sleeping Princesses" (Julianna Barwick & Mary Lattimore / Tragic Magic) → all 5 streaming buttons greyed despite release being on Apple Music | Spotify+Apple structurally unreachable on a Discogs/library MISS (BS's `metadata.service.ts` fills them only from LML's `artwork` block; `SearchUrlProvider` has no Spotify/Apple branch; LML's own in-`artwork` search-URL fallback is dead code on a total miss). iOS inline-v2 path additionally suppresses the proxy fallback when `hasV2Metadata` is true with null streaming columns. | WXYC/Backend-Service#1184 (open) |
[...]

Issue #368, opened 2026-05-24
Bulk `/api/v1/lookup` endpoint for amortizing cold-cache cost across batches


jakebromberg:
## Problem

`POST /api/v1/lookup` is single-item. Backend-Service's [flowsheet-metadata-backfill-cron](https://github.com/WXYC/Backend-Service/blob/main/jobs/flowsheet-metadata-backfill/orchestrate.ts) (BS#1011) and the live enrichment-worker call it one row at a time, each paying a fresh cold-cache penalty per #337. With the BS#995 static gate (`BACKFILL_LML_MAX_CONCURRENT=1`), drain throughput is ~14 rows/min — months to clear a 1.6M-row residual.

The 2026-05-23 drain-acceleration pass on Backend-Service flipped 309,379 of 1.9M pending rows in 80 minutes via a pure-SQL JOIN against `album_metadata`. The residual decomposes:

| bucket | count | unique LML calls needed |
|---|---:|---:|
| `album_id IS NULL` (DJ free-form) | 744,449 | 744,449 (each unique) |
| `album_id IS NOT NULL`, album never enriched | 857,364 | **35,692** (dedup by album_id) |

A bulk endpoint amortizes per-row cost in two ways:
1. **One HTTP roundtrip per N items** instead of N roundtrips
2. **In-process cache (`TTLCache`/`LRUCache`) stays warm across all N items in a batch** — N-1 of N items hit warm if they share an artist/album. With per-row calls and a 1-deep BS gate, those caches go cold between calls.

## Desired end state

```
POST /api/v1/lookup/bulk
Authorization: Bearer <LML_API_KEY>
Content-Type: application/json

{ "items": [
    {"artist": "Will Oldham", "album": "The Isota Singles Club", "track": "Barcelona"},
    ...
] }
```

Response (results in input order):

```json
{ "results": [
    {"index": 0, "status": "match",     "discogs_url": "...", "artwork_url": "...", "release_year": 1996, ...},
    {"index": 1, "status": "no_match"},
    {"index": 2, "status": "error",     "message": "..."}
] }
```

Handler internals:
- `asyncio.gather` with bounded semaphore (mirror #278's pattern for `bulk-resolve-libraries`)
- Share the in-process caches naturally — single process, single event loop
- Max 100 items per request (return 400 above)
- Honor `X-Caller-Budget-Ms` if/when #345 lands; for now, an env-configured per-batch ceiling

## Where

[...]

PR #410, opened 2026-05-29
fix(orchestrator): album-match floor on artist-fallback path (closes #400)

### LINKED ISSUE #400: LML free-text lookup falls back to single Discogs release on album miss, inheriting wrong-album metadata onto ~184k flowsheet rows
## Problem

When LML's `/api/v1/lookup` handles a free-text request (artist + DJ-typed album + song) and the typed album doesn't resolve, the lookup pipeline falls back to **a single Discogs release for the artist** and returns it as the top match. Backend-Service's enrichment then inherits the matched release's `release_year`, `apple_music_url`, `spotify_url`, `discogs_url`, and `artwork_url` and writes them onto the flowsheet row — regardless of which album the DJ actually typed. The result: every free-text play of, say, Miles Davis shows the same album artwork, the same year, the same Spotify link, and the same Discogs page across 65 distinct DJ-typed albums.

#390 and #398 (both merged + deployed to prod on 2026-05-28) tightened the **iTunes-result** verification — #389 (wrong-artist `results[0]`) and #396 (right-track-on-wrong-album when album is provided). Those guards verify the *iTunes match* against the request, but they don't gate the **LML lookup result itself** — the upstream Discogs release that BS pulls year/artwork/spotify/discogs from doesn't go through any album-match floor. So even with #390 + #398 deployed, free-text album misses keep producing wrong-album `release_year` / `spotify_url` / `discogs_url` / `artwork_url`. (#398 does help on the Apple side for the specific Yenbett shape by rejecting same-named tracks on a different album collection.)

## Evidence (prod RDS, 2026-05-28)

Free-text rows = `flowsheet WHERE album_id IS NULL`. Of **1,412,114** total:
- **199,647** have a non-NULL `discogs_url`
- **199,149** have a non-NULL `spotify_url`
- **182,447** have a non-NULL `release_year`
- **28,976** have a non-NULL `apple_music_url`

Of those with `discogs_url`, **16,532 distinct `discogs_url` values** are each attached to multiple distinct `(artist_name, album_title)` pairs — i.e. the same Discogs release was inherited as metadata for multiple different DJ-typed albums. Those 16,532 URLs touch **183,822 rows** carrying **105,290 distinct `(artist, album)` variants** — ~6.4 album variants per contaminated release URL.

Top 10 offenders (`discogs_url` → `(rows, distinct albums)` → representative DJ-typed variants):

| Discogs release | Rows | Distinct albums | Example DJ-typed variants |
|---|---:|---:|---|
| `/release/15815070` | 674 | 431 | "Unknown" / V/A fallback used as catch-all (Cambodian Cassette Archives, 16mm School Soundtracks, 500 Lock Grooves, …) |
| `/release/1402136` | 419 | 150 | Nina Simone — every entry (`Anthology`, `At Newport`, `At the Village Gate`, `Baltimore`, `A Single Woman`, `Bad Habits`, …) |
| `/release/12654` | 301 | 239 | V/A fallback (239 different compilations) |
| `/release/31134770` | 279 | 34 | Horaflora + Secret Boyfriend (split-release leak across both artists' catalogues) |
| `/release/1390987` | 278 | 100 | Os Mutantes — 100 distinct album entries |
| `/release/21625933` | 253 | 80 | Ariel Pink — same shape |
| `/release/2133766` | 251 | 65 | Miles Davis — 65 album entries |
| `/release/1171368` | 246 | 110 | Outkast — 110 album entries |
| `/release/14773378` | 236 | 130 | Hank Williams — 130 album entries |
| `/release/1643524` | 224 | 95 | Johnny Cash — 95 album entries |

Each row in those groups except (at most) one is **wrong-album metadata**.

Diagnostic SQL (read-only, reproducible):

```sql
SET search_path TO wxyc_schema, public;

-- (1) Upper bound: free-text rows with any inherited Discogs-side metadata.
SELECT
  COUNT(*) FILTER (WHERE discogs_url IS NOT NULL)     AS with_discogs_url,
  COUNT(*) FILTER (WHERE apple_music_url IS NOT NULL) AS with_apple_url,
  COUNT(*) FILTER (WHERE spotify_url IS NOT NULL)     AS with_spotify_url,
  COUNT(*) FILTER (WHERE release_year IS NOT NULL)    AS with_release_year,
  COUNT(*)                                            AS total_freetext_rows
FROM flowsheet
WHERE album_id IS NULL;

-- (2) Contamination signal: discogs_urls used across multiple fs_album values.
SELECT COUNT(*) AS contaminated_discogs_urls,
       SUM(row_count) AS total_rows_on_contaminated_urls,
       SUM(distinct_albums) AS distinct_album_variants_across_those_urls
FROM (
  SELECT discogs_url,
         COUNT(*) AS row_count,
         COUNT(DISTINCT (artist_name, album_title)) AS distinct_albums
  FROM flowsheet
  WHERE album_id IS NULL AND discogs_url IS NOT NULL
  GROUP BY discogs_url
  HAVING COUNT(DISTINCT (artist_name, album_title)) > 1
) t;

-- (3) Top 20 worst offenders (eyeball the variants).
SELECT discogs_url,
       COUNT(*) AS rows,
       COUNT(DISTINCT (artist_name, album_title)) AS distinct_albums,
       STRING_AGG(DISTINCT (artist_name || ' / ' || album_title), '; ' ORDER BY (artist_name || ' / ' || album_title)) AS variants
FROM flowsheet
WHERE album_id IS NULL AND discogs_url IS NOT NULL
GROUP BY discogs_url
HAVING COUNT(DISTINCT (artist_name, album_title)) > 1
ORDER BY COUNT(*) DESC
LIMIT 20;
```

## Where the fallback comes from

Almost certainly the search-strategy pipeline in `lookup/orchestrator.py` — when `ARTIST_PLUS_ALBUM` finds no album match, the cascade falls through to artist-fallback / `TRACK_ON_COMPILATION` / `SONG_AS_ARTIST` modes that return a top result for the artist regardless of which album was requested. The downstream `enrich_artwork_results` then attaches that release's `artwork`, and BS's `extractAlbumMetadata` projects every album-derived field. Investigation should pin the exact strategy:

- https://github.com/WXYC/library-metadata-lookup/blob/4ff35d436d4a647852e14c805cae54f1b316538e/lookup/orchestrator.py (search strategies + `enrich_artwork_results`)
- https://github.com/WXYC/library-metadata-lookup/blob/4ff35d436d4a647852e14c805cae54f1b316538e/core/search.py (declarative strategy cascade)

The BS write surface that *inherits* every album-derived field is the no-album-match-but-result-returned path in `apps/backend/services/metadata/metadata.service.ts` (`extractAlbumMetadata` at the top-1 artwork block) — the same surface called out in WXYC/Backend-Service#1184.

## Desired end state

LML should not return a Discogs-release-anchored top-1 result whose album differs materially from the requested album when album was provided. Two reasonable shapes (decide during implementation):

1. **In LML** — reject artist-fallback results when `request.album` is non-empty and the matched release's title doesn't clear an 80-floor `token_set_ratio` against the requested album (mirroring the streaming-availability batch matcher and the #390/#398 floor). The lookup either returns a hit whose album matches, or none.
2. **In the LookupResponse contract** — surface a per-result `album_match_confidence` so BS's `extractAlbumMetadata` can conditionally omit the album-derived fields when confidence is below a floor. Apple / Spotify / year / artwork / Discogs URL get dropped, free-text row stays uncontaminated.

Either way the goal is the same: stop persisting `release_year` / `apple_music_url` / `spotify_url` / `discogs_url` / `artwork_url` derived from a different album. Approach 2 is more flexible because BS can fall back to `SearchUrlProvider` for YT/BC/SC search URLs while skipping the strict-album fields — mirroring the shape WXYC/Backend-Service#1184 needs for the Discogs-miss case, so the two tickets would slot together naturally.

## Constraints

- #32 critical surfaces (`lookup/orchestrator.py`, `apps/backend/services/metadata/`, V2 flowsheet shape) — bolt-tightening, in-scope for the freeze.
- Reuse the existing match primitives (`wxyc_etl.text.to_match_form` + `rapidfuzz.fuzz`) — no coupling to `scripts/` (same deferral as #389/#396).
- API contract: if approach 2 (surfaced confidence) is chosen, that's a `wxyc-shared/api.yaml` change — needs the codegen pass and consumer updates on BS / iOS.

## Acceptance criteria

- [ ] Unit test exercising the failure mode: a free-text request `(artist="Miles Davis", album="Kind of Blue", song="<known track>")` against a mocked LML library catalog that has *no* matching Kind of Blue → lookup returns no album-derived match (or a result flagged as low album confidence) instead of an arbitrary Miles Davis Discogs release.
- [ ] Integration test against real Discogs cache for at least one of the contaminated cases above (Nina Simone / Miles Davis / Outkast).
- [ ] BS-side: when LML returns no `artwork` OR album-confidence below floor, `extractAlbumMetadata` is not called and `result.album` carries only the YT/BC/SC search URLs (same shape WXYC/Backend-Service#1184 targets for the Discogs-miss case).
- [ ] Re-running the prod contamination diagnostic above on **newly enriched** rows after the fix shows zero contamination growth.

## Follow-up — backfill (do NOT file from this ticket)

After the upstream fix lands and is deployed, file a one-shot backfill job to re-enrich the existing ~184k contaminated rows. Likely shape: a sibling of `flowsheet-metadata-backfill` that selects `album_id IS NULL AND discogs_url IS NOT NULL`, re-runs LML lookup for each `(artist_name, album_title, track_title)`, and re-applies enrichment via the post-fix code path. Will need pacing + a circuit-breaker (per the WXYC/Backend-Service#994/#995 lessons from the LML backfill monopolization incident on 2026-05-21). File when the upstream fix is closer to landing.

## Related

- #389 (closed-via-#390) and #396 (closed-via-#398) — the iTunes-result verification predecessors. Both merged + deployed to prod on 2026-05-28.
- WXYC/Backend-Service#1184 — the sibling structural-reachability bug (Spotify/Apple unreachable on a Discogs miss). Approach 2 above (surfaced confidence + BS conditional projection) would let BS#1184's fix slot in naturally alongside this one.
- Parent tracker: #397.

jakebromberg:
## Problem

When `search_library_with_fallback` in `lookup/orchestrator.py` falls through to its artist-only branch (or its artist+song keyword branch returns a row whose title doesn't relate to the typed album), the cascade historically returned an arbitrary library row for the typed artist. Downstream `enrich_artwork_results` then attached the matched Discogs release's `release_year` / `apple_music_url` / `spotify_url` / `discogs_url` / `artwork_url` and Backend-Service wrote those onto the flowsheet row — tagged with a completely different album.

Prod evidence in WXYC/library-metadata-lookup#400: ~184k free-text flowsheet rows carry this shape, attached to 16,532 distinct Discogs release URLs each bound to multiple `(artist, album)` pairs. Top offenders include Nina Simone (`/release/1402136` — 150 distinct DJ-typed albums across 419 rows), Outkast (110 distinct), Hank Williams (130 distinct), Miles Davis (65 distinct).

WXYC/library-metadata-lookup#390 and WXYC/library-metadata-lookup#398 tightened the iTunes-result verification (artist match floor; track-on-album-collection check). This PR tightens the LML lookup result itself — the upstream Discogs release that BS pulls year / artwork / spotify / discogs from.

## Approach (locked: approach 1)

Reject artist-fallback candidates whose library titles don't clear `rapidfuzz.fuzz.token_set_ratio >= 80` against `parsed.album` when `parsed.album` is non-empty. On sub-floor, every candidate drops and the function returns the unknown-artist shape (`[], song_not_found_flag`) — same shape LML already returns for an unknown artist. No `LookupResponse` contract change. No `wxyc-shared/api.yaml` ripple. No BS consumer update needed.

Why approach 1 over approach 2 (surface `album_match_confidence`):
- Hermetic to LML — single-repo PR.
- Matches the existing "no-result for unknown shape" behavior at every consumer that already handles it.
- Reversible: if the floor turns out to be too strict, bumping `_ALBUM_MATCH_FLOOR` is a one-line knob.
- Floor mirrors the values in WXYC/library-metadata-lookup#390 / WXYC/library-metadata-lookup#398 — keeps the verification surface coherent.

## Implementation

New helper at `lookup/orchestrator.py:742`:

```python
_ALBUM_MATCH_FLOOR = 80.0

def _filter_results_by_album_match(
    results: list[LibraryItem],
    album: str | None,
) -> list[LibraryItem]:
    if not album or not album.strip():
        return results
    from rapidfuzz import fuzz
    norm_album = normalize_for_comparison(album)
    kept = [
        item for item in results
        if fuzz.token_set_ratio(norm_album, normalize_for_comparison(item.title or "")) >= _ALBUM_MATCH_FLOOR
    ]
    if len(kept) < len(results):
        logger.info(
            f"Album-match floor dropped {len(results) - len(kept)} of {len(results)} "
            f"artist-fallback candidates against typed album '{album}' (#400)"
        )
    return kept
```

Called at both cascade legs (`f"{parsed.artist} {parsed.song}"` keyword path and the artist-only path) in `search_library_with_fallback`.

Pure CPU, no I/O. Lazy `from rapidfuzz import fuzz` matches the 5 other lazy-import sites in this file. Reuses `normalize_for_comparison` already imported at module scope. Safe to call in the lookup hot path without compounding the LML cascade semaphore-queue accumulation (the BS#1064 / BS#1078 failure mode documented in the memory and in `project_lml_cascade_mechanism.md`).

One existing test modified: `test_falls_back_to_artist_only_when_no_discogs_albums` had `album="Unknown Album"` which would now be rejected by the floor (Queen titles don't match "Unknown Album"). Dropped the album field (still exercises the artist-only fallback shape) and added a docstring note pointing readers to the new test file.

Nina Simone fixture added to `tests/integration/conftest.py` (call letters `SI`, "Pastel Blues" — top-2 prod offender from the #400 evidence table).

## Tests

| Suite | Result |
|---|---|
| `tests/unit/test_album_match_floor.py` (9 new) | **9 / 9 passing** |
| `tests/integration/test_album_match_floor.py` (2 new) | **2 / 2 passing** |
| `tests/unit/test_orchestrator_helpers.py` (1 modified) | passing |
| **`tests/unit/` (full suite)** | **1931 / 1931 passing** in 82.13s |

Unit cases:
- `test_floor_constant_is_80` — pinned at the same value as WXYC/library-metadata-lookup#390 / WXYC/library-metadata-lookup#398's `_APPLE_MUSIC_MATCH_FLOOR`.
- `test_artist_fallback_rejected_when_typed_album_misses_library_titles` — Miles Davis / Kind of Blue against a library with Bitches Brew + On the Corner only. Both candidates fail the floor → empty result.
- `test_artist_fallback_keeps_candidate_when_title_clears_floor` — Kind of Blue (50th Anniversary Edition) clears via `token_set_ratio`.
- `test_floor_skipped_when_typed_album_empty` — regression pin against the 808 State / Flow Coma case in `test_orchestrator_helpers.py`.
- `test_floor_skipped_when_typed_album_whitespace_only` — Outkast with `album="   "` still surfaces.
- `test_floor_rejects_outkast_top_offender` — concrete prod offender (`/release/1171368`, 110 distinct albums across 246 rows).
- `test_floor_rejects_hank_williams_top_offender` — concrete prod offender; includes a live `rapidfuzz.fuzz.token_set_ratio` assertion so the test isn't relying on opaque ratio internals.
- `test_floor_handles_diacritics_via_normalization` — Hermanos Gutiérrez / ASCII variant via `wxyc_etl.text.to_match_form`.
- `test_artist_plus_song_branch_also_filtered` — Nina Simone Sinnerman: the FTS hit "Pastel Blues" gets dropped when the DJ typed "Wild Is the Wind".

Integration cases (real `LibraryDB`, Discogs mocked to empty so the library-side floor is the unit-under-test):
- `test_nina_simone_wild_is_the_wind_does_not_surface_pastel_blues` — wrong-album case; Pastel Blues is in the library but DJ typed "Wild Is the Wind" → floor drops it.
- `test_nina_simone_pastel_blues_typed_correctly_still_surfaces` — green companion; typed-correctly request returns Pastel Blues.

## Local pre-flight

```
pytest tests/unit/                # 1931 passed in 82.13s
ruff check                        # All checks passed!
ruff format --check               # 265 files already formatted
mypy . --ignore-missing-imports   # Success: no issues found in 265 source files
```

## Out of scope (followups)

- **One-shot backfill of the 184k contaminated rows** — separate ticket to be filed after this PR deploys to prod and soaks. Needs pacing + circuit-breaker per WXYC/Backend-Service#994 / WXYC/Backend-Service#995. Will leverage the bulk endpoint (WXYC/library-metadata-lookup#368) for compression.
- **Discogs-side contamination shapes** — this PR drops the library row that would have surfaced the wrong Discogs release. `enrich_artwork_results` is keyed on the matched library row, so suppressing the wrong library row cascades to suppress the wrong Discogs enrichment. A follow-up could add an end-to-end integration test where Discogs DOES return a candidate release to confirm the suppression works through both layers.
- **Approach-2 confidence-band variant** — explicitly declined for this PR. If LML ever needs to surface "we found a candidate but album confidence is below floor" as a positive signal (vs the current "no result"), that's a `wxyc-shared/api.yaml` change with codegen + BS consumer updates. Today there's no consumer that distinguishes "no result" from "low confidence" — both fall back to the same `SearchUrlProvider` path on BS. Re-evaluate if a consumer needs the distinction.

Closes #400.

Refs: WXYC/library-metadata-lookup#397 (Tragic Magic cluster tracker), WXYC/library-metadata-lookup#390, WXYC/library-metadata-lookup#398 (sibling iTunes-result verification floors). Also relates to WXYC/Backend-Service#1184 (Tragic Magic structural reachability — this PR is the "stop producing wrong-album metadata upstream" complement to BS#1184's "make the client adapt to missing metadata" iOS-side fix in WXYC/wxyc-ios-64#303).

### DIFF
--- a/lookup/orchestrator.py
+++ b/lookup/orchestrator.py
@@ -739,6 +739,49 @@ def _release_matches_library_row(release: ReleaseInfo, item: LibraryItem) -> boo
     return False
 
 
+# Minimum fuzzy score (0-100) for accepting a library-row title as a genuine
+# album match for the DJ-typed album. Mirrors `_APPLE_MUSIC_MATCH_FLOOR` and
+# the streaming-availability batch matcher. When the artist-fallback branches
+# of `search_library_with_fallback` surface a row whose title doesn't clear
+# this floor against the typed album, the row would otherwise carry the
+# matched Discogs release's `release_year` / `apple_music_url` / `spotify_url`
+# / `discogs_url` / `artwork_url` onto a flowsheet row tagged with a
+# completely different album — the contamination shape documented in #400
+# (~184k rows; 16,532 distinct Discogs URLs each attached to many distinct
+# DJ-typed `(artist, album)` pairs). #390 / #398 tightened the iTunes-result
+# verification; this tightens the LML lookup result itself.
+_ALBUM_MATCH_FLOOR = 80.0
+
+
+def _filter_results_by_album_match(
+    results: list[LibraryItem],
+    album: str | None,
+) -> list[LibraryItem]:
+    """Drop library rows whose title doesn't clear `_ALBUM_MATCH_FLOOR` against
+    the typed album. No-ops when `album` is empty or whitespace-only.
+
+    Pure CPU (`rapidfuzz.fuzz.token_set_ratio` + `to_match_form`); no I/O —
+    safe to call in the lookup hot path without compounding the LML cascade
+    semaphore-queue accumulation (the BS#1064 / BS#1078 failure mode).
+    """
+    if not album or not album.strip():
+        return results
+    from rapidfuzz import fuzz
+
+    norm_album = normalize_for_comparison(album)
+    kept: list[LibraryItem] = []
+    for item in results:
+        title_norm = normalize_for_comparison(item.title or "")
+        if fuzz.token_set_ratio(norm_album, title_norm) >= _ALBUM_MATCH_FLOOR:
+            kept.append(item)
+    if len(kept) < len(results):
+        logger.info(
+            f"Album-match floor dropped {len(results) - len(kept)} of {len(results)} "
+            f"artist-fallback candidates against typed album '{album}' (#400)"
+        )
+    return kept
+
+
 async def search_library_with_fallback(
     db: LibraryDB,
     parsed: ParsedRequest,
@@ -839,6 +882,7 @@ async def search_library_with_fallback(
         query = f"{parsed.artist} {parsed.song}"
         results = await db.search(query=query, limit=_FETCH_LIMIT)
         results = filter_results_by_artist(results, parsed.artist)
+        results = _filter_results_by_album_match(results, parsed.album)
 
         if results:
             song_lower = parsed.song.lower()
@@ -852,6 +896,7 @@ async def search_library_with_fallback(
         logger.info(f"No results for albums {albums}, trying artist only: '{parsed.artist}'")
         results = await db.search(query=parsed.artist, limit=_FETCH_LIMIT)
         results = filter_results_by_artist(results, parsed.artist)
+        results = _filter_results_by_album_match(results, parsed.album)
         if results:
             return results, True
 
--- a/lookup/orchestrator.py
+++ b/lookup/orchestrator.py
@@ -759,10 +759,6 @@ def _filter_results_by_album_match(
 ) -> list[LibraryItem]:
     """Drop library rows whose title doesn't clear `_ALBUM_MATCH_FLOOR` against
     the typed album. No-ops when `album` is empty or whitespace-only.
-
-    Pure CPU (`rapidfuzz.fuzz.token_set_ratio` + `to_match_form`); no I/O —
-    safe to call in the lookup hot path without compounding the LML cascade
-    semaphore-queue accumulation (the BS#1064 / BS#1078 failure mode).
     """
     if not album or not album.strip():
         return results
@@ -777,7 +773,7 @@ def _filter_results_by_album_match(
     if len(kept) < len(results):
         logger.info(
             f"Album-match floor dropped {len(results) - len(kept)} of {len(results)} "
-            f"artist-fallback candidates against typed album '{album}' (#400)"
+            f"artist-fallback candidates against typed album '{album}'"
         )
     return kept
 
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -87,6 +87,12 @@ SEED_ITEMS = [
     (39, "Confield", "Autechre", "EL", 16, 1, "Electronic", "CD"),
     (40, "1st Class", "Large Professor", "HH", 1, 1, "Hiphop", "CD"),
     (41, "...Destroys The Space Invaders", "Prince Jammy", "RE", 1, 1, "Reggae", "Vinyl"),
+    # Nina Simone: top-2 prod offender in the #400 evidence table
+    # (/release/1402136 → 150 distinct DJ-typed albums across 419 rows). The
+    # library carries Pastel Blues but NOT every Nina Simone album a DJ might
+    # type — the artist-fallback contamination floor (#400) must drop
+    # candidates whose titles don't clear 80 against the typed album.
+    (42, "Pastel Blues", "Nina Simone", "SI", 1, 1, "Jazz", "Vinyl"),
     # Junior Kimbrough: title-album and same-artist comp that both contain the
     # song "Meet Me in the City". Exercises song-title-vs-primary-album
     # ranking when several library candidates by the same artist all pass
--- a/tests/integration/test_album_match_floor.py
+++ b/tests/integration/test_album_match_floor.py
@@ -33,20 +33,14 @@ class TestAlbumMatchFloorIntegration:
     @pytest.mark.asyncio
     async def test_nina_simone_wild_is_the_wind_does_not_surface_pastel_blues(self, library_db):
         """The exact prod-evidence failure mode: artist matches the library but
-        the typed album doesn't, and the artist-fallback cascade would
-        otherwise contaminate the flowsheet row with a wrong-album Discogs
-        release.
+        the typed album doesn't, and the artist-fallback cascade would otherwise
+        contaminate the flowsheet row with a wrong-album Discogs release.
 
         Library seed has Nina Simone / Pastel Blues only. Request types album
-        "Wild Is the Wind" + song "Sinnerman". Without the floor:
-        artist-fallback surfaces Pastel Blues, enrichment attaches its
-        release metadata, BS persists `release_year` / `apple_music_url` /
-        `spotify_url` / `discogs_url` onto a row tagged "Wild Is the Wind".
-
-        With the floor: token_set_ratio("Wild Is the Wind", "Pastel Blues")
-        is well below 80 (no shared significant tokens), the candidate is
-        dropped, and no album-derived enrichment leaks. The response mirrors
-        the unknown-artist shape.
+        "Wild Is the Wind" + song "Sinnerman" —
+        token_set_ratio("Wild Is the Wind", "Pastel Blues") sits well below 80
+        (no shared significant tokens), so the candidate is dropped and the
+        response mirrors the unknown-artist shape.
         """
         from wxyc_fastapi.observability import init_cache_stats
 
--- a/tests/integration/test_album_match_floor.py
+++ b/tests/integration/test_album_match_floor.py
@@ -1,0 +1,134 @@
+"""Integration test for the artist-fallback album-match floor (LML#400).
+
+End-to-end exercise of ``perform_lookup`` against a real (in-memory)
+``LibraryDB`` with the seeded Nina Simone fixture (top-2 prod offender from
+the #400 evidence table — `/release/1402136`, 150 distinct DJ-typed albums
+across 419 contaminated rows).
+
+When a DJ types a Nina Simone album that the WXYC library doesn't physically
+carry (e.g. "Wild Is the Wind") plus a song that Nina Simone DOES have on a
+DIFFERENT album in the library ("Sinnerman" → "Pastel Blues"), the
+artist-fallback cascade historically surfaced the wrong-album library row,
+and ``enrich_artwork_results`` then attached Pastel Blues' Discogs release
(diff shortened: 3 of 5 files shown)


jakebromberg:
### Code review

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #445, opened 2026-05-31
fix: add logging to _fetch_apple_music_url when iTunes API returns non-200

### LINKED ISSUE #444: _fetch_apple_music_url silently swallows non-200 responses (no log, no Sentry); add observability so iTunes outages stop being invisible
## Problem

[`_fetch_apple_music_url`](https://github.com/WXYC/library-metadata-lookup/blob/f25889ad0f8cb9a53818f73c0cc722486f223daf/lookup/orchestrator.py#L1618-L1680) returns `None` on every non-200 iTunes response with no log line and no Sentry capture:

```python
resp = await http_client.get(url)
if resp.status_code != 200:
    return None      # ← no logger.warning, no sentry_sdk.capture_message
```

And the broad except at the bottom is just as quiet:

```python
except Exception:
    return None      # ← same: silent
```

This means iTunes-side issues (rate limit, IP block, DNS failure, certificate error, timeout, malformed response) are observationally indistinguishable from "no match found." The only signal is downstream: `apple_music_url` quietly going null for every consumer.

## Why it matters (current incident)

Apple Music URLs have been persisting as null for ~100% of new flowsheet enrichments since 2026-05-28 due to iTunes returning uniform 403 Forbidden on every call from LML's Railway IP. The outage was invisible to LML's operators because of the silent-failure pattern above. Filed as the sibling incident ticket. This ticket is the defensive fix that would have surfaced that incident days earlier.

The pattern recurs (the 2026-05-10 to 2026-05-16 BS `apple_music_url` collapse window is likely the same shape, also invisible at the time).

## Desired end state

Every non-200 iTunes response and every caught exception in `_fetch_apple_music_url` emits both a `logger.warning` and a `sentry_sdk.capture_message`. Specifically:

```python
if resp.status_code != 200:
    logger.warning(
        "iTunes Search returned %d for artist=%r song=%r",
        resp.status_code, artist, song,
    )
    sentry_sdk.capture_message(
        f"iTunes Search returned {resp.status_code}",
        level="warning",
        extras={"artist": artist, "song": song, "status_code": resp.status_code},
    )
    return None
```

And the except block:

```python
except Exception as exc:
    logger.warning(
        "iTunes Search exception for artist=%r song=%r: %s",
        artist, song, exc,
    )
    sentry_sdk.capture_exception(exc)
    return None
```

This converts every non-200 into a Sentry event and a log line — at a 100% failure rate (the current incident) the Sentry alert would fire within minutes; at a 1% failure rate (transient flakes) the event volume stays manageable.

## Constraints

- Don't change the return-`None`-on-failure contract — callers depend on it.
- Don't raise — the broad except is intentional per LML#241's FD-leak hygiene.
- Sentry rate-limiting may need consideration if a long iTunes outage produces thousands of capture_message calls. The existing Sentry SDK config probably handles this; verify before merging.
- The `logger.warning` calls should use lazy formatting (`%s` / `%r` placeholders, not f-strings) to avoid the format cost when logging is filtered out.

## Acceptance criteria

- [ ] Both failure branches (non-200 and exception) in `_fetch_apple_music_url` emit `logger.warning` with artist/song context.
- [ ] Both failure branches emit a Sentry `capture_message` or `capture_exception` with structured extras.
- [ ] Unit test: mock iTunes returning 403, assert log line + Sentry capture both fire.
- [ ] Unit test: mock iTunes raising httpx.TimeoutException, assert log line + Sentry capture both fire.
- [ ] Existing happy-path and verification tests still pass.
- [ ] During the next iTunes outage (sibling incident ticket), the failure surfaces in Sentry within minutes rather than days.

## Related

- Sibling: incident ticket — the active iTunes 403 outage this would have caught earlier. (Filing concurrently; cross-link.)
- LML#241 — established the broad-except pattern for FD-leak safety. This ticket adds observability *inside* that pattern without changing its semantics.
- LML#389 / LML#390 / LML#396 — touched `_fetch_apple_music_url` for verification logic; none added observability.
- Related pattern check worth doing: the same silent-failure shape probably exists in [`clients/streaming/apple_music.py:84-91`](https://github.com/WXYC/library-metadata-lookup/blob/f25889ad0f8cb9a53818f73c0cc722486f223daf/clients/streaming/apple_music.py#L84-L91) (logs `logger.warning` but no Sentry) and in `spotify.py`, `deezer.py`, `bandcamp.py`. Consider folding all four streaming clients into a single observability pass, or file as separate follow-up.

kouyouqi123:
Adds logging when iTunes API returns non-200. Fixes #444

### DIFF
--- a/lookup/orchestrator.py
+++ b/lookup/orchestrator.py
@@ -1647,6 +1647,7 @@ async def _fetch_apple_music_url(
         url = f"https://itunes.apple.com/search?term={query}&entity=song&media=music&limit=5"
         resp = await http_client.get(url)
         if resp.status_code != 200:
+            logger.warning("iTunes API returned %d for %s %s", resp.status_code, artist, song)
             return None
         results = resp.json().get("results", [])
 
@@ -1676,7 +1677,8 @@ async def _fetch_apple_music_url(
                     continue
             return track_url
         return None
-    except Exception:
+    except Exception as exc:
+        logger.error("iTunes API request failed: %s", exc)
         return None
 
 


jakebromberg:
Triage finding (2026-06-07): #444 closed via the #462 timeout telemetry path. This PR's narrower logging concern may already be covered by that Sentry instrumentation. Recommend either rebasing onto current main and confirming the logging gap still exists, or closing as obsolete.

PR #514, opened 2026-06-07
test(discogs): pin fetched_at writer invariant and bulk-path stub semantics

### LINKED ISSUE #511: test(discogs): pin fetched_at writer invariant and bulk-path stub semantics
## Problem

`get_artist_details`'s fallthrough discriminator (`discogs/service.py:939`) depends on the invariant that any artist cache row written by LML has `fetched_at` set, and any row with `fetched_at IS NULL` was written by the out-of-repo `discogs-etl` stub-from-`release_artist` path. The invariant currently holds — `write_artist_details` (`discogs/cache_service.py:1152-1166`) is the only LML-resident writer and always sets `fetched_at` — but nothing enforces it. A future writer that omits `fetched_at` will silently break the discriminator: real rows get re-fetched from Discogs on every call, no test fails, the bug manifests only as quiet cache thrash.

Adjacent: test fixture at `tests/integration/test_search_aliases_bulk.py:175-179` seeds rows with `INSERT INTO artist (id, name, profile, image_url)` — no `fetched_at`. The rows it seeds are conceptually \"asked Discogs at least once\" (Stereolab with children, Juana Molina as the \"ran the leg, found nothing\" case — see comment at line 170), but the production write shape stamps `fetched_at = now()` on both. As written, the fixture looks like rebuild stubs to anyone reading it, contradicting the invariant this issue is trying to pin.

The bulk-path stub semantics (`get_artist_details_bulk` not projecting `fetched_at`) is being fixed separately in #520. This issue's bulk test depends on that landing first.

## Scope

Tests + fixture only. Production code change for the bulk-projection foot-gun is in #520.

### 1. Writer-invariant test (integration)

New file or extend an existing integration file: `@pytest.mark.pg`, uses `pg_pool` fixture. Real PG round-trip — not an SQL-string substring match — so the test catches \"writer that stamps `NULL` via a parameter\" as well as \"writer that drops the column.\"

Two cases, parameterized over the two SQL paths inside `write_artist_details`:

- **INSERT path**: fresh artist id → call `write_artist_details` → `SELECT fetched_at` → assert non-NULL.
- **ON CONFLICT UPDATE path**: pre-seed a row with `fetched_at = '2000-01-01'` → call `write_artist_details` → `SELECT fetched_at` → assert non-NULL **and** strictly greater than the seeded value.

Single writer today, so no parametrize-over-writers harness. If a second writer is added later, that writer's author adds a parallel test.

### 2. Bulk-path stub semantics test (integration, depends on #520)

Extend `tests/integration/test_search_aliases_bulk.py` with a class that exercises `get_artist_details_bulk` directly. Seed two artists:

- **Stub-shape row**: `(id, name)` only, `fetched_at IS NULL`, no children.
- **Hydrated row**: `(id, name, profile, image_url, fetched_at = now())` plus one alias / one name variation / one member.

Call `get_artist_details_bulk([stub_id, hydrated_id])`. Pin the chosen semantics:

- **Both** ids present in the returned dict — bulk is a faithful cache read, does not filter stubs.
- Stub entry → `profile=None`, `aliases=[]`, `name_variations=[]`, `members=[]`, `urls=[]`, `fetched_at=None`, `cached=True`.
- Hydrated entry → populated children, `fetched_at` non-NULL.

Rationale for pinning \"present-but-empty\" over \"filtered as miss\": cache layer stays a faithful read; the \"is this hydrated?\" judgment is service-layer policy (and lives on the singular path's `is_pg_hit` predicate). Composer (`artists/composer.py:144`) is unaffected either way.

### 3. Fixture cleanup

Update `tests/integration/test_search_aliases_bulk.py:175-179` to stamp `fetched_at = now()` on both Stereolab and Juana Molina rows. Both are conceptually \"asked Discogs\" rows (Juana's docstring at line 170 says as much); the production write shape gives them non-NULL `fetched_at`. The fixture should match.

Easier to land after #520 because then the existing endpoint tests in that file also exercise the corrected `fetched_at` shape end-to-end.

## Sequencing

1. Land #520 (bulk SELECT projects `fetched_at`).
2. Land this issue:
   - Writer-invariant test (independent of #520, could land first, but cleaner to bundle).
   - Bulk-path stub test (depends on #520 — without it, every bulk return has `fetched_at=None` and the test can't distinguish stub from hydrated).
   - Fixture cleanup.

## Acceptance

- [ ] Writer-invariant test fails if `fetched_at = now()` is removed or replaced with `NULL` in either branch of `write_artist_details`'s SQL.
- [ ] Writer-invariant test fails if a future change stops stamping `fetched_at` on the ON CONFLICT UPDATE path (real PG round-trip catches this; SQL substring grep would not).
- [ ] Bulk-path stub test pins: stub → `fetched_at IS NULL`, hydrated → non-NULL, both present in result dict.
- [ ] `test_search_aliases_bulk.py` fixture matches production write shape (`fetched_at = now()` on Stereolab + Juana).

## Related

- #503 — introduced the `fetched_at` discriminator that this invariant guards.
- #520 — fixes the bulk-projection asymmetry; the bulk test here depends on it.
- WXYC/Backend-Service#1361 — relies on the discriminator holding.

jakebromberg:
## Summary

Tests-only follow-up to #503 (the singular path's `fetched_at` cache-hit discriminator) and #520 (the bulk SELECT now projects `fetched_at`, fixed in #524). Pins the invariants both depend on so future refactors cannot silently regress them.

### Writer invariant (unit + integration)

Two layers cover this — substring on SQL text plus a real PG round-trip — because each catches a regression class the other misses.

- **Unit** (`tests/unit/test_cache_service.py::TestWriteArtistDetailsFetchedAtInvariant`): parametrised over the INSERT and ON CONFLICT branches of `write_artist_details`'s UPSERT. Asserts each branch carries `fetched_at` and stamps it server-side via `now()`. Drop the column from either branch → two parametrisations fail; replace `now()` → `test_artist_upsert_stamps_fetched_at_with_now` fails. Verified locally by mutation.
- **Integration** (`tests/integration/test_cache_service_artist_writer.py`, new file, `@pytest.mark.pg`): two cases against real PG, per #511's acceptance criteria.
  - INSERT path → write a fresh id → assert `fetched_at IS NOT NULL`.
  - ON CONFLICT path → seed with `fetched_at = '2000-01-01'` → write → assert `fetched_at > seed`. Catches a writer that binds `fetched_at` as a parameter and is handed a stale value — the substring grep would pass but the row would stay stuck in the past.

### Bulk-path stub semantics

Pins the contract `get_artist_details_bulk` exposes post-#524: the bulk SELECT (`cache_service.py:1154`) projects `fetched_at`, and callers can use the same `fetched_at is None` discriminator the singular path uses (`is_pg_hit` at `discogs/service.py:1014`).

Unit-level pins in `tests/unit/test_cache_service.py::TestGetArtistDetailsBulkStubSemantics`:

- `test_stub_row_surfaces_with_null_fetched_at` — a row with `fetched_at IS NULL` comes back with `fetched_at=None` (not filtered; bulk stays a faithful cache read).
- `test_stub_and_hydrated_distinguishable_by_fetched_at` — in a mixed batch, both rows surface but the caller can tell them apart.
- `test_bulk_select_projects_fetched_at` — substring check on the projection so a future change that drops the column is forced to update the semantics suite.

End-to-end coverage against real PG already lives in `TestGetArtistDetailsBulkFetchedAt` (added in #524); no duplication here.

### Fixture cleanup

`tests/integration/test_search_aliases_bulk.py` now seeds both Stereolab and Juana Molina with `fetched_at = now()`, matching the production write shape (#511 acceptance criterion #4). The stub-row edge case is exercised by `TestGetArtistDetailsBulkFetchedAt::test_stub_row_returns_null_fetched_at` and `test_mixed_batch_preserves_per_row_fetched_at`, which now NULL the column per-test. The fixture defaults to the common case so a new endpoint test added here inherits production shape, not the rebuild-stub edge case it would have silently picked up before.

## Test plan

- [x] New unit tests pass: `uv run pytest tests/unit/test_cache_service.py::TestWriteArtistDetailsFetchedAtInvariant tests/unit/test_cache_service.py::TestGetArtistDetailsBulkStubSemantics -v` (6 passed)
- [x] Affected PG integration suite passes: `uv run pytest -m pg tests/integration/test_search_aliases_bulk.py tests/integration/test_cache_service_artist_writer.py -v` (12 passed — endpoint suite + bulk-path fetched_at suite + writer-invariant integration)
- [x] Mutation verified: removing `fetched_at` from both UPSERT branches in `discogs/cache_service.py` fails the two substring parametrisations; replacing `now()` fails `test_artist_upsert_stamps_fetched_at_with_now`.
- [x] Full unit suite green: `uv run pytest tests/unit/`
- [x] Ruff lint + format clean on the full repo
- [ ] CI green on this branch

## Refs

- Closes #511
- Builds on #503 (singular-path discriminator) and #524 (bulk SELECT now projects `fetched_at`, addresses issue #520)
- Downstream: WXYC/Backend-Service#1361 relies on the discriminator behaviour pinned here

### DIFF
--- a/tests/integration/test_cache_service_artist_writer.py
+++ b/tests/integration/test_cache_service_artist_writer.py
@@ -1,0 +1,186 @@
+"""Integration tests for ``DiscogsCacheService.write_artist_details``'s
+``fetched_at`` invariant against real PostgreSQL.
+
+The cache-hit discriminator in ``DiscogsService.get_artist_details``
+(discogs/service.py) keys on ``ArtistDetails.fetched_at`` -- any row written
+by LML has a non-NULL ``fetched_at``; rows with ``fetched_at IS NULL`` are
+rebuild-created stubs that need to be re-fetched. The unit-level pin in
+``tests/unit/test_cache_service.py::TestWriteArtistDetailsFetchedAtInvariant``
+catches the column being dropped from the SQL text, but a future writer
+that binds the value as a parameter (and lets a caller pass ``None``) or
+swaps ``now()`` for a functionally-identical-but-textually-different
+alternative would slip past a substring check.
+
+This module asserts the post-write database state instead:
+
+* INSERT path: a fresh id is written with a non-NULL ``fetched_at``.
+* ON CONFLICT UPDATE path: a pre-seeded row with ``fetched_at = '2000-01-01'``
+  has its timestamp advanced past the seed value after the write.
+
+Run with: ``pytest -m pg -v tests/integration/test_cache_service_artist_writer.py``
+"""
+
+from __future__ import annotations
+
+import os
+from datetime import UTC, datetime
+
+import asyncpg
+import pytest
+import pytest_asyncio
+
+from discogs.cache_service import DiscogsCacheService
+from discogs.models import ArtistDetails
+
+DATABASE_URL = os.getenv(
+    "DATABASE_URL_TEST",
+    "postgresql://discogs:discogs@localhost:5433/discogs",
+)
+
+
+@pytest_asyncio.fixture
+async def pg_pool():
+    try:
+        pool = await asyncpg.create_pool(DATABASE_URL, min_size=1, max_size=3)
+    except Exception as e:
+        pytest.skip(f"Cannot connect to test PostgreSQL: {e}")
+        return
+    yield pool
+    await pool.close()
+
+
+@pytest_asyncio.fixture(autouse=True)
+async def fresh_artist_schema(pg_pool):
+    """Bring up the artist parent + child tables LML writes through.
+
+    Mirrors the production discogs-cache shape: ``fetched_at`` is nullable
+    (rebuild stubs land here as NULL) and ``not_found`` defaults to FALSE
+    (LML#510 tombstone column). The writer is responsible for populating
+    ``fetched_at`` on every write -- this fixture deliberately does NOT
+    default it server-side so the test catches any writer that fails to
+    stamp it.
+    """
+    async with pg_pool.acquire() as conn:
+        for child in (
+            "artist_alias",
+            "artist_name_variation",
+            "artist_member",
+            "artist_url",
+        ):
+            await conn.execute(f"DROP TABLE IF EXISTS {child} CASCADE")
+        await conn.execute("DROP TABLE IF EXISTS artist CASCADE")
+        await conn.execute(
+            """
+            CREATE TABLE artist (
+                id         integer PRIMARY KEY,
+                name       text NOT NULL,
+                profile    text,
+                image_url  text,
+                fetched_at timestamptz,
+                not_found  boolean NOT NULL DEFAULT FALSE
+            )
+            """
+        )
+        await conn.execute(
+            """
+            CREATE TABLE artist_alias (
+                artist_id  integer NOT NULL REFERENCES artist(id) ON DELETE CASCADE,
+                alias_id   integer,
+                alias_name text NOT NULL
+            )
+            """
+        )
+        await conn.execute(
+            """
+            CREATE TABLE artist_name_variation (
+                artist_id integer NOT NULL REFERENCES artist(id) ON DELETE CASCADE,
+                name      text NOT NULL
+            )
+            """
+        )
+        await conn.execute(
+            """
+            CREATE TABLE artist_member (
+                artist_id   integer NOT NULL REFERENCES artist(id) ON DELETE CASCADE,
+                member_id   integer NOT NULL,
+                member_name text NOT NULL,
+                active      boolean DEFAULT true
+            )
+            """
+        )
+        await conn.execute(
+            """
+            CREATE TABLE artist_url (
+                artist_id integer NOT NULL REFERENCES artist(id) ON DELETE CASCADE,
+                url       text NOT NULL
+            )
+            """
+        )
+    yield
+    async with pg_pool.acquire() as conn:
+        for t in (
+            "artist_alias",
+            "artist_name_variation",
+            "artist_member",
+            "artist_url",
+            "artist",
+        ):
+            await conn.execute(f"DROP TABLE IF EXISTS {t} CASCADE")
+
+
+@pytest.mark.pg
+@pytest.mark.asyncio
+async def test_insert_path_stamps_fetched_at_non_null(pg_pool):
+    """A fresh artist write lands with ``fetched_at`` non-NULL.
+
+    Catches any writer that omits the column or binds it as a nullable
+    parameter -- the substring unit test would still see ``fetched_at``
+    in the SQL, but the row would have ``NULL`` in the column.
+    """
+    cache = DiscogsCacheService(pg_pool)
+    await cache.write_artist_details(ArtistDetails(artist_id=2154, name="Stereolab"))
+
+    async with pg_pool.acquire() as conn:
+        fetched_at = await conn.fetchval("SELECT fetched_at FROM artist WHERE id = 2154")
+
+    assert fetched_at is not None, (
+        "INSERT branch of write_artist_details must stamp fetched_at; the "
+        "cache-hit discriminator in DiscogsService.get_artist_details (#503) "
+        "treats fetched_at IS NULL as a rebuild stub and re-fetches every call."
+    )
+
+
+@pytest.mark.pg
+@pytest.mark.asyncio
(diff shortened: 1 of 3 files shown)


jakebromberg (commented):
A few divergences from #511's spec that are worth a second pass — one fine, one mildly weakening, one that I think blocks.

### Writer invariant: unit SQL-substring vs the integration assertion #511 specified (mild)

The unit tests here catch:
- Dropping `fetched_at` from either branch — good.
- Replacing `now()` with a constant on the UPDATE branch — caught by the `count(\"now()\") >= 2` check, good.

What's lost compared to #511's acceptance criterion #2 (\"pre-seed `fetched_at = '2000-01-01'`, write, assert strictly greater\"):
- A future writer that does `SET fetched_at = '2000-01-01'` (or binds it as a parameter that happens to be None) leaves the substring `fetched_at` in the SQL, so the substring test passes — but no row's timestamp ever moves forward. That's exactly the \"stamps NULL via a parameter\" case #511 explicitly called out wanting to catch.
- Swapping `now()` for `current_timestamp` (functionally identical in PG) would false-fail.

Not a blocker — most realistic regressions are still caught — but I'd add one `@pytest.mark.pg` integration test for the seed-and-advance assertion. The substring tests can stay as a fast CI mirror; belt-and-suspenders.

### Bulk semantics pin the opposite invariant (blocker, I think)

This is the one I'd hold the PR on. #511's acceptance criterion #3 is:

> Bulk-path stub test pins: stub → `fetched_at IS NULL`, hydrated → non-NULL, both present in result dict.

`TestGetArtistDetailsBulkStubSemantics` pins the opposite — \"stub and real are indistinguishable; both come back `fetched_at=None`\" — and the docstring frames it as a deliberate end state (\"the bulk path surfaces stubs as cache hits... that's deliberate\"). That directly contradicts #520, which exists to fix this exact asymmetry as a foot-gun.

Concrete merge-ordering consequence:

- If #520 lands first, `test_stub_and_real_indistinguishable_by_caller` fails immediately when the bulk SELECT starts projecting `fetched_at`.
- If this PR lands first, #520 can't land without rewriting these tests — which means #520 acquires a test rewrite in its diff and the \"tests-only follow-up that pins the right shape\" goal of #511 is inverted: we'd be regression-guarding the bug.

Either way the test name (\"indistinguishable_by_caller\") will get cited later as canonical intent and confuse the next reader.

Options:

1. Rebase this PR behind #520 and flip the bulk class to the post-#520 shape (stub → `fetched_at IS NULL`, hydrated → non-NULL, both present).
2. Strip the bulk class out of this PR — land writer invariant + fixture cleanup now — and re-file the bulk class as a separate tests-only PR after #520.
3. Keep the class but `@pytest.mark.skip(reason=\"depends on #520\")` with the assertions written for the post-#520 shape, so it flips on automatically when #520 lands.

(2) is probably cleanest given #511 explicitly sequenced this work behind #520.

### Placement in `test_cache_service.py` instead of `test_search_aliases_bulk.py` (not a concern)

#511 sketched extending the integration file, but cache-layer assertions belong in the cache-service test file. Endpoint file is about search-aliases composition, not bulk cache reads. This placement is arguably better than what the issue specified — keep as-is.

### Fixture cleanup

Matches the spec, no notes.

jakebromberg:
Addressed in 23c3895 (rebased onto #520 in main):

- **Bulk semantics flipped.** `TestGetArtistDetailsBulkStubSemantics` now pins the post-#520 shape: stub → `fetched_at IS NULL`, hydrated → non-NULL, both present in the result dict. Old "indistinguishable" framing (and the inverted `test_bulk_select_does_not_project_fetched_at`) is gone.
- **Writer seed-and-advance pg test added.** New `tests/integration/test_cache_service_artist_writer.py` (`@pytest.mark.pg`) — INSERT path asserts `fetched_at` is non-NULL; ON CONFLICT path pre-seeds `fetched_at = '2000-01-01'` and asserts strictly greater after the write. Catches the "binds NULL as a parameter" / `current_timestamp` cases the substring unit tests miss. Substring unit tests stay as the fast CI mirror (belt-and-suspenders).
- **Fixture reverted to NULL on both rows.** #520 added `TestGetArtistDetailsBulkFetchedAt` to `test_search_aliases_bulk.py` and depends on the stub shape; writer-side stamping is now pinned in the dedicated writer integration file instead.
- **Placement kept** in `test_cache_service.py` per your read — cache-layer assertion belongs with cache tests.

Issue #518, opened 2026-06-09
fix(orchestrator): guard synthesized release_id=0 sentinel before Discogs API call


jakebromberg:
## Problem

`lookup/orchestrator.py` passes a synthesized `release_id=0` sentinel to `discogs_service.get_release(0)` at two call sites, hitting the Discogs API for `https://api.discogs.com/releases/0` which returns 404. Sentry (90d, project `wxyc/library-metadata-lookup`) shows `release/0` requested ~15× per 18h on the existing bare-except path at `discogs/service.py:838`. The request is structurally invalid; the API should never be called.

## Root cause

The synthesis pattern was added for WXYC/Backend-Service#1184 (releases on Apple Music not in the WXYC/Discogs catalog still surface streaming buttons on iOS). `lookup/orchestrator.py:1988` and `:1122` produce a `DiscogsSearchResult(release_id=0, release_url="", ...)` carrying only the streaming-URL fields. The `(release_id=0, release_url="")` pair is a documented cross-service contract: BS#1185 keys off it to skip `extractAlbumMetadata`. See the docstring at `lookup/orchestrator.py:1720-1734`.

Two downstream consumers don't guard against the sentinel before calling Discogs:

1. **`lookup/orchestrator.py:1668`** -- `fallback = await _resolve_fallback_artwork(discogs_service, result.release_id)`. `result` here can be the synthesized sentinel. `_resolve_fallback_artwork` at `:1569-1594` immediately calls `release = await discogs_service.get_release(release_id)` with no zero-check.
2. **`lookup/orchestrator.py:1780`** -- `release = await discogs_service.get_release(top_artwork.release_id)` inside `fetch_top1_release_details`. `top_artwork` can be the synthesized sentinel.

Both sites then check `if not release: return None` after the API call, so the failure is silent -- the symptom is only the wasted Discogs API call (rate-limit budget) and the Sentry log noise.

The orchestrator already has correct gating in several other places (`:745`, `:1205`, `:1466`: `if release.release_id:` / `if best_result.release_id:` before passing to downstream calls). The two flagged sites are inconsistent with the surrounding pattern.

## Fix

Add the same `if not release_id` / `if not result.release_id` gate that the surrounding code already uses.

- `_resolve_fallback_artwork` at `lookup/orchestrator.py:1571`: guard the body with `if not release_id: return None` as the first line. Cheaper than gating the caller because there's only one call site to update and it makes the helper robust against any future caller.
[...]

PR #522, opened 2026-06-09
feat(discogs): tombstone Discogs 404s in PG cache (LML#510)

### LINKED ISSUE #510: feat(discogs): tombstone 404s for get_artist_details + get_release via not_found discriminator
## Problem

`get_artist_details` (`discogs/service.py:864-941`) and `get_release` (`discogs/service.py:737-862`) both re-hit the Discogs API on every call for an artist or release id that 404s. The `_api_fetch` inner functions wrap `response.raise_for_status()` in a bare `except Exception` (`:918`, `:838`) and return `None` on any error. The fallthrough seam (`discogs/fallthrough.py:276`) deliberately skips `pg_write` when the API leg returns `None`, so no cache row records "we asked Discogs, the answer was nothing." Every subsequent call burns rate-limit budget on the same 404.

This bites three callers:
- WXYC/Backend-Service#1361 (rotation backfill -- hundreds of artist ids per run; tightest loop).
- `lookup/orchestrator.py:1789` (per-track enrichment -- one-shot per request, but the cross-request cache miss accumulates).
- Any future job that walks `release_artist` ids.

## Reframe vs. original proposal

The first revision of this issue proposed wiring the seam's `pg_negative_check` / `pg_negative_record` / `is_empty` / `on_negative_hit` quadrant (mirroring `search_releases_by_track:587-599`) and ~250 LOC across a new `negative_artist` table, two new `cache_service` methods, and the four fallthrough hooks at every call site.

That's the wrong shape. Commit 3e54907 (the `#503` `fetched_at` discriminator) already turned `is_pg_hit` into the negative-cache mechanism -- it just isn't being fed a sentinel yet, because `_api_fetch` collapses 404 to `None` and the seam refuses to write `None`. The principled fix is one layer, not three: discriminate 404 inside `_api_fetch`, return a tombstone-shaped row, let the existing `pg_write` + `is_pg_hit` machinery do the rest.

## Scope

Apply the same fix to both `get_artist_details` and `get_release` in one PR -- they have the same bug and the same solution. (The scope-audit confirmed: `get_release`'s existing `artwork_checked_at` discriminator only handles the *back-fill* path -- ETL stubs without artwork -- not the 404 path; on 404 it still skips the write and re-hits the API on call N+1.)

`get_master` (`discogs/service.py:992`) and `get_label` (`:969`) have the same `_api_fetch`-collapses-404-to-`None` shape but no PG cache backing, so there's no tombstone surface to land in without standing up new schema. Defer; reconsider when either method grows a cache leg or shows up in a tight-loop caller. Not a follow-up issue today -- document the deferral in the PR description.

Out of scope: `search` (`discogs/service.py:1163`) and `validate_track_on_release` (`discogs/service.py:1248`) are read-only by design per the `fallthrough.py:24-25` policy table -- the negative case there has no schema to land in. Document the design constraint in that table comment as part of this PR; do not file follow-ups.

## Cross-repo scope

Three repos coordinate on this change:

1. **WXYC/discogs-etl** -- alembic migration adding `not_found BOOLEAN NOT NULL DEFAULT FALSE` to `artist` and `release`. One migration each, mirroring the shape of `0008_release_artwork_checked_at.py`. Two precondition checks during the migration:
   - `artist.name` and `release.title` must accept the empty-string tombstone write. Spot-confirmed against `discogs-etl/migrations/02_enrich_data.sql:37` and `schema/create_database.sql:153`: both are `text NOT NULL` with no `CHECK (<col> <> '')`, so the migration goes through unchanged; re-verify in the migration's `upgrade()` so a future-added constraint trips the test rather than the prod write.
   - The rebuild pipeline's UPSERTs (and the dedup-swap path) must include `not_found = FALSE` in the `EXCLUDED` set / SQL `INSERT` so fresh data clears any prior tombstone. Without this, a tombstoned id survives a pipeline refresh and stays unreachable until the admin endpoint (below) hits it. Same gotcha applies to the rebuild-from-scratch path: the swap-in shouldn't pre-seed `not_found = TRUE` for any id.
2. **WXYC/wxyc-shared** -- add `not_found: bool` to `DiscogsReleaseMetadata` in `api.yaml`. The existing `artwork_checked_at` on that schema (line 1130 of the generated `api_models.py` in this repo) is precedent. `ArtistDetails` is local (`discogs/models.py:137`, intentionally not aliased to the generated form per the comment there) and grows the field directly in this repo -- no api.yaml change for the artist side.
3. **WXYC/library-metadata-lookup** -- this PR. Bump the wxyc-shared pin, run `scripts/generate_api_models.sh`, commit the regenerated `generated/api_models.py`, then the code changes below.

Land in this order: (1) discogs-etl alembic + rebuild-path UPSERT change merge to staging first; (2) wxyc-shared `api.yaml` PR + release; (3) this LML PR pins the new wxyc-shared release, runs codegen, and consumes the new column. The LML PR's integration tests need the new column live in the PG staging DB to pass.

## Change

Three layers in this repo.

1. **404 discrimination in `_api_fetch`.** In both methods, inspect `response.status_code == 404` *before* calling `raise_for_status()`. On 404, return the model populated with `not_found=True` and the identifier-derivable fields, instead of falling into the `except` branch:
   - `ArtistDetails(artist_id=artist_id, name="", not_found=True)`
   - `ReleaseMetadataResponse(release_id=release_id, title="", artist="", release_url=f"https://www.discogs.com/release/{release_id}", not_found=True)` -- `title` and `artist` are required `str` on the model so populate them with empty strings; `release_url` is identifier-derived so it's safe even for a tombstone; the remaining list/optional fields take their defaults.

   The existing bare `except Exception` keeps the transient-failure behavior (returns `None`, seam skips write-back) -- no change there.

2. **Boundary translation at two call sites.**
   - **Public service methods.** In `DiscogsService.get_artist_details` / `get_release`, translate `v.not_found` back to `None` in both return paths (the seam's return value and the `cache is None` early-return) so the public contract (`ArtistDetails | None` / `ReleaseMetadataResponse | None`) is preserved for callers like `lookup/orchestrator.py:1584` and `:1789`.
   - **`CachedOnlyResolver` (`discogs/markup_parser.py:466-470` and `:473-478`).** Both methods call `cache_service.get_artist_details` / `get_release` *directly*, not via `DiscogsService`, and rely on `None` to mean "not in cache." After the cache-service short-circuit (point 3 below) starts returning tombstones with `name = ""` / `title = ""`, these need explicit `if details is None or details.not_found: return None` guards -- otherwise `resolve_artist` returns `""` and the markup parser renders an `artistLink` token with `display_name=""` instead of falling through as plain text. (`scripts/track_streaming/track_extractor.py:52` is the only other direct caller of `cache_service.get_release`; it already gates on `release.tracklist` being non-empty so a tombstone falls through to title-parsing naturally -- no change needed.)

3. **Cache-service plumbing.** Three sub-touches on `cache_service.py`:
   - **Read-path SELECT + short-circuit.** `get_release` SQL at `:498-502` and `get_artist_details` SQL at `:989-992` project `not_found`. When `not_found = TRUE`, short-circuit before the child-table `asyncio.gather` fetches (`:508-540` and `:543-562` for releases; `:998-1015` for artists). No artists/labels/tracks exist for a tombstone -- saves 4-7 PG round-trips per tombstone hit.
   - **Write-path UPSERT, two branches with exact SQL.** The tombstone branch goes at the top of each method, runs a *narrow* UPSERT, and `return`s before the existing child-table cascade. The non-tombstone branch is the existing UPSERT plus one new column in its `ON CONFLICT DO UPDATE SET` clause.

     Release tombstone branch (early-return before the cascade at `:711-799`):
     ```sql
     INSERT INTO release (id, title, release_year, artwork_url, released,
                          artwork_checked_at, not_found)
     VALUES ($1, '', NULL, NULL, NULL, now(), TRUE)
     ON CONFLICT (id) DO UPDATE SET
         artwork_checked_at = EXCLUDED.artwork_checked_at,
         not_found = TRUE
     -- intentionally omitted: title, release_year, artwork_url, released
     -- so a previously-hydrated row's identifier columns survive a later 404
     ```

     Artist tombstone branch (early-return before the cascade at `:1168-onward`):
     ```sql
     INSERT INTO artist (id, name, profile, image_url, fetched_at, not_found)
     VALUES ($1, '', NULL, NULL, now(), TRUE)
     ON CONFLICT (id) DO UPDATE SET
         fetched_at = now(),
         not_found = TRUE
     -- intentionally omitted: name, profile, image_url
     ```

     The existing non-tombstone branches gain `not_found = FALSE` in their `ON CONFLICT DO UPDATE SET` clause so a recovered 200 (after the admin endpoint cleared the row, or a concurrent in-flight 200 races with the admin DELETE) re-hydrates and clears the flag in one statement. Without this, a tombstoned id whose Discogs upstream recovers stays unreachable forever -- the next `_api_fetch` returns a non-tombstone model, but the `ON CONFLICT DO UPDATE` path leaves `not_found = TRUE` untouched.

     Critical: the existing cascade unconditionally `DELETE`s `release_artist` / `release_label` / `release_track` / `release_track_artist` / `release_genre` / `release_style` / `release_video` (`:711-799`) and `artist_alias` / `artist_name_variation` / `artist_member` / `artist_url` (`:1168-onward`). The early-`return` makes tombstone writes purely additive on the row level: the parent gets a state marker, the children stay intact in case the tombstone is later cleared and the row is rehydrated from a recovered API response. Under a Discogs 404 storm without this branch we'd lose rich catalog metadata that the etl pipeline can only repopulate on its next run (and only for ids in `library.db`).
   - **`get_artist_details_bulk` intentionally untouched.** Mirrors 3e54907's decision for stubs: *"`get_artist_details_bulk` is intentionally left alone: it's cache-only by design and doesn't traverse the fallthrough seam, so stub rows return as empty `ArtistDetails` to callers."* Tombstones get the same treatment -- returned as empty-fields `ArtistDetails`. The one current caller (`artists/composer.py:144-169`) handles "in bundle with empty aliases" identically to "missing from bundle" (the `_variant_*` generators yield nothing), so filter vs. no-filter is observationally equivalent. Stay symmetric with the prior decision.

The fallthrough seam needs no changes. The SQL `now()` in `write_artist_details` / `write_release` stamps `fetched_at` / `artwork_checked_at` on the tombstone. The existing `is_pg_hit` predicates (`v.fetched_at is not None` and `v.artwork_url is not None or v.artwork_checked_at is not None`) already treat the tombstone as a hit on subsequent reads.

## Why this shape over the original proposal

- **Pattern alignment.** This is the third instance of the same discriminator-column shape -- `release.artwork_checked_at` (WXYC#414), `artist.fetched_at` (#503), now `not_found` on both. The codebase has converged on "ambiguity resolved by a state column on the existing row" twice in 90 days; doing it a third time deepens the pattern rather than introducing a parallel `negative_*` table.
- **Bulk-read safety.** A sentinel encoded as `name = ""` would leak empty `ArtistDetails` into `get_artist_details_bulk` (`cache_service.py:1119-1129`) with no way to distinguish from a real empty-aliased artist. An explicit `not_found` column keeps the bulk path's behavior unambiguous even though it doesn't act on the flag today.
- **No new seam plumbing.** The seam's `pg_negative_check` / `pg_negative_record` quadrant exists for *search* negatives, where the empty answer is structurally different from the positive answer (a `DiscogsSearchResponse` with `results=[]` vs. one with results). For a single-entity 404, the positive answer is `T | None` and `None` IS the negative answer -- translating it through a tombstone column is the right level.

## False-tombstone recovery

False tombstones happen. The most plausible source isn't a typo'd id -- it's a Discogs incident returning 404s for valid ids; under a brief outage we could tombstone hundreds or thousands of entities in quick succession.

The original revision of this issue claimed `?refresh=true` (commit 5ef8023 / #498) provided a recovery path. That's incorrect: refresh evicts L1 only, and the tombstone lives in L2 (PG). On `?refresh=true` against a tombstoned id, the seam reads the tombstone from PG, `is_pg_hit` returns True (because `fetched_at`/`artwork_checked_at` is set), the tombstone is returned, and `_api_fetch` is never invoked. `?skip_cache=true` doesn't recover either -- the seam at `fallthrough.py:276` skips `pg_write` when `skip` is set, so even on a recovered 200 it wouldn't overwrite the tombstone. Without a recovery path, an incident's tombstones stay until the next discogs-etl pipeline rebuild overwrites them -- which only covers ids that are in `library.db`.

Recovery path for this PR: a small admin endpoint, `DELETE /admin/discogs/tombstone/{type}/{id}`, that issues `DELETE FROM artist WHERE id = $1 AND not_found = TRUE` (or `release` likewise) and also evicts the L1 entry via the existing `evict_cached` primitive from #498. Manual, narrow, no surprise behavior changes to existing query params, and pairs naturally with the telemetry counter below -- if the post-deploy 404 rate spikes during a Discogs incident, the on-call has a single endpoint to call to undo a batch. The `AND not_found = TRUE` guard means the endpoint physically cannot delete a real row. Auth via the existing `ADMIN_TOKEN` (`routers/admin.py:_validate_auth`) -- distinct from `LML_API_KEY` so every caller doesn't gain incident-grade write access. Considered extending `?refresh=true` to also clear L2 tombstones; rejected on blast-radius grounds (every `LML_API_KEY` holder vs on-call only) and because it would invert #498's deliberate L1-only scope decision. Add a one-line cross-reference to the tombstone endpoint in `_REFRESH_DESCRIPTION` (`discogs/router.py:29-34`) so operators discover the right primitive without burning a triage cycle.

Response codes (distinguishable so a recovery script can branch without parsing log volume):
- `200 {deleted: true, id, type}` -- a tombstoned row matched and was deleted; the next request for that id will re-fetch from Discogs.
- `404 {detail: "row exists but is not a tombstone", id, type}` -- `id` exists with `not_found = FALSE`; the operator probably typo'd, or the tombstone was already cleared.
- `404 {detail: "no row for this id", id, type}` -- `id` doesn't exist at all.

Bulk variant (`DELETE /admin/discogs/tombstones?type=artist&since=<timestamp>`) is intentionally deferred until the telemetry shows we need it.

## TTL deferred

Adding a TTL to negatives but not positives breaks the symmetry without serving a clear failure mode now that the admin recovery endpoint is in place. Discogs artist and release deletions are effectively permanent in the common case; the admin endpoint handles the incident-driven false-tombstone case. Reconsider on the merits if the post-deploy telemetry shows sustained false-tombstone churn.

## Tests

- 404 returns the tombstone, fallthrough writes it via `pg_write`, public method returns `None` (integration).
- Subsequent call hits the cache row, `_api_fetch` not invoked (unit, with mocked `_api_fetch` counter).
- Transient failure (5xx, `ConnectionError`) returns `None` from `_api_fetch`, seam skips `pg_write`, no tombstone written (unit). Critical: must not poison the cache with a false tombstone during an upstream outage.
- `cache_service.get_release` / `get_artist_details` skip the child-table fetches on `not_found = TRUE` (unit, with a PG query counter or asyncpg `connection_recorder`).
- **Cascade preservation.** Write a row + rich children via the bulk-loader path; write a tombstone for the same id via `write_release` / `write_artist_details`; assert every child row (`release_artist`, `release_label`, `release_track`, `release_track_artist`, `release_genre`, `release_style`, `release_video`; `artist_alias`, `artist_name_variation`, `artist_member`, `artist_url`) survives. Regression test for the cascade-wipe hazard the write-path branch in point 3 above addresses.
- **Hydrated parent fields survive tombstone overwrite.** Separate from cascade-preservation. Insert a hydrated parent row (`name="Stereolab"`, `profile`, `image_url`; or `title="Aluminum Tunes"`, `release_year=1998`, `artwork_url`, `released`); call `write_*` with a tombstone model for the same id; read back and assert the identifier columns survived and only `not_found` flipped to TRUE plus `_checked_at` advanced. The cascade-preservation test would pass against a buggy `EXCLUDED.name`/`EXCLUDED.title` UPSERT shape that wipes parent identifiers; this one pins the narrow update set.
- **Tombstone clears on recovered 200.** Write a tombstone; then call `write_*` with a hydrated payload (`not_found=False`); assert `not_found` is `FALSE` and identifier columns reflect the new payload. Pins the non-tombstone branch's `not_found = FALSE` clause -- without it, a tombstoned id whose Discogs upstream recovers stays unreachable forever.
- **`CachedOnlyResolver` boundary.** With a tombstoned row in PG, `CachedOnlyResolver.resolve_artist` / `resolve_release` return `None` (not the empty-string identifier the cache short-circuit hands back). Pins the second translation site so a future refactor that drops the guard fails a test rather than silently corrupting bio token rendering.
- `get_artist_details_bulk` returns tombstones as empty-fields `ArtistDetails`, parallel to stubs (unit -- pins the "left alone" policy from 3e54907).
- Admin endpoint deletes only `not_found = TRUE` rows and refuses to touch real data (unit). Parameterize across the three response codes (`200`, `404 row-exists-not-tombstone`, `404 no-row`). After deletion, the next call re-runs `_api_fetch`; on a recovered 200 the row is overwritten via the UPSERT path (integration).

## Telemetry

Three counters, all flat (no tag dimensions) -- the existing `record(key)` surface auto-creates keys (`pg_negative_hits` is the precedent), and the projection in `lookup/router.py:63-89` and `lookup/orchestrator.py:2283` ships any new counter through Sentry transaction `data` + PostHog `lookup_completed` automatically. No wxyc_fastapi PR needed.

- **Write-side, at the 404 discrimination point in `_api_fetch`:** `record("artist_404_tombstone_written")` / `record("release_404_tombstone_written")`. Flat, not tagged by status. Sentry 90d evidence on the existing bare-except path: zero `"410 Gone"` events, 95% of `get_release` failures are 404, and the Discogs developer docs (`https://www.discogs.com/developers/`) document 404 but do not document 410 at all -- a `{404, 410}` tag would have 100% cardinality on `404` forever. Re-evaluate if a future Sentry scrub for `message:"410 Gone"` surfaces non-zero counts on the bare-except path.
- **Read-side, at the boundary-translation point in the public service methods (`discogs/service.py:~925` for artists, `~845` for releases):** `record("tombstone_returned")`. Keeps `pg_cache_hit` semantically "any L2 hit"; dashboard math for real-data hit rate is `pg_cache_hit - tombstone_returned`. Lives in the public method, not the fallthrough seam, because the seam is type-opaque on purpose (`is_pg_hit` is just `Callable[[T | None], bool]`) and one consumer doesn't justify widening that surface.
- **Adjacent: `release_5xx_passthrough` at the same 404-discrimination point.** Sentry shows ~5% of `get_release` bare-except hits are 5xx, currently lumped with everything else. Splitting them out keeps Discogs-outage signal legible separately from tombstone activity. Cheap; ships with this PR.

Observability gap to fix in this PR: `discogs/service.py:919` uses `logger.warning(...)` for artist failures while `:839` (release) uses `logger.error(...)`. Sentry's default Python integration captures `error+`, skipping warnings unless explicitly configured. The "zero artist 404s in Sentry 90d" finding is therefore a lower-bound artifact, not a true absence -- BS#1361's hot loop is still hitting artist 404s; we just can't count them. Promote the artist bare-except to `logger.error` in this PR so the write-side counter has a Sentry corroboration surface post-deploy.

L1 caveat for traffic shaping: `memory_cache.py:217-219` does not cache `None` results, so within the L1 TTL window every call for a tombstoned id still pays one PG round-trip (no API call). That's the main win vs. status quo, but it's not a "free" L1 hit -- the cache_stats `pg_cache_hit` rate will rise as tombstones land (which is why the read-side `tombstone_returned` counter exists -- to keep `pg_cache_hit` interpretable). If the PG round-trip cost shows up in traces, the follow-up is to relax `@async_cached` to cache `None` for these methods, which is out of scope here.

## Priority

Not load-bearing for WXYC/Backend-Service#1361 -- that issue's BS-side mitigation (persist a "tried, got 404" set across runs) is sufficient for its loop's correctness. Ship #1361 with that mitigation, then ship this as the server-side cleanup that benefits every other caller. Schedule accordingly.

## Related

- WXYC/Backend-Service#1361 -- calls `get_artist_details` in a tight loop; immediate beneficiary of the server-side fix.
- #503 -- `fetched_at` discriminator on `get_artist_details`; this issue composes with it directly (the predicate stays the same; the tombstone is what feeds it). Also establishes the "bulk left alone" policy that this issue inherits.
- #498 / commit 5ef8023 -- `?refresh=true` L1 eviction primitive; explicitly does NOT cover L2 tombstones, motivating the admin endpoint above.
- WXYC#414 -- `artwork_checked_at` discriminator on `get_release`; same pattern, this issue extends it to cover the 404 case.

## Follow-ups out of scope

- **Upstream `release/0` validation -- filed as #518.** Sentry shows `release/0` being requested ~15×/18h. Root cause is two unguarded LML orchestrator call sites (`lookup/orchestrator.py:1668` and `:1780`) that pass the synthesized `(release_id=0, release_url="")` sentinel from BS#1184 through to `discogs_service.get_release`. Tombstoning here would silence the Sentry signal without fixing the bug. Land #518 first so the Sentry noise is fixed at the right layer; otherwise this PR will mask the orchestrator bug and #518 can only be diagnosed by reading code.
- **Cache `None` in L1 for tombstoned ids.** Currently every tombstone hit pays one PG round-trip (L1 doesn't cache `None`). Relaxing `@async_cached` to cache `None` for the two affected methods would eliminate the PG round-trip but widens the decorator's contract. Reconsider if `pg_cache_hit - tombstone_returned` traces show the PG round-trip cost is material.

jakebromberg:
## Summary

- `_api_fetch` in `DiscogsService.get_release` / `get_artist_details` discriminates 404 from other failures and returns a tombstone-shaped model (`not_found = True`, `title=""` / `name=""` as identifier sentinels) instead of falling into the bare `except`. The existing fallthrough seam writes the tombstone via the same `pg_write` it already uses.
- Subsequent reads short-circuit on `not_found = TRUE` at the `cache_service` layer (no child-table `asyncio.gather` round-trips for a tombstoned id). The public `DiscogsService` methods translate the tombstone back to `None` so callers' `T | None` contract is unchanged.
- New tombstone-write branch in `write_release` / `write_artist_details` preserves the parent's identifier columns and the entire child cascade (the `ON CONFLICT DO UPDATE SET` clause omits identifier columns; the early `return` skips the child `DELETE+INSERT` cascade), so a 404 after a 200 doesn't wipe rich catalog metadata. Symmetric: the hydrated-write branch adds `not_found = FALSE` to its UPDATE SET so a recovered 200 clears any prior tombstone in one statement.
- `DELETE /admin/discogs/tombstone/{type}/{id}` recovery endpoint for incident-driven false tombstones. `AND not_found = TRUE` guard at the SQL level means a real row can never be deleted through this surface. Three response codes (`deleted` / `exists_not_tombstone` / `not_found`) so an on-call recovery script can branch without parsing logs. Auth via `ADMIN_TOKEN` (distinct from `LML_API_KEY` — so routine callers don't gain incident-grade write access). Cross-referenced from `?refresh=true`'s description so operators discover the right primitive.
- `CachedOnlyResolver` (in `discogs/markup_parser.py`) reads `cache_service` directly and gets explicit `if details is None or details.not_found: return None` guards — without these, the markup parser would render `artistLink` tokens with `display_name=""` for any tombstoned id.
- `discogs/service.py` logger promotion (`warning` → `error`) on the artist bare-except path: Sentry's default `error+` filter previously masked artist 404s (90-day Sentry scrub showed zero artist 404 events). The new `artist_404_tombstone_written` counter gets a Sentry corroboration surface post-deploy.

## Telemetry

- `release_404_tombstone_written` / `artist_404_tombstone_written` — write-side, at the 404 discrimination point.
- `tombstone_returned` — read-side, at the boundary translation. `pg_cache_hit - tombstone_returned` is the real-data hit rate.
- `release_5xx_passthrough` — splits Discogs-outage signal from tombstone activity, ships in this PR per the Sentry 90d scrub (5% of `get_release` bare-except hits are 5xx, currently lumped).

All flat (no tag dimensions) per the existing `record(key)` precedent (`pg_negative_hits` is the existing example).

## Cross-repo coordination

Sequenced after:
1. WXYC/discogs-etl#277 — alembic migration adding `release.not_found` + `artist.not_found` with `DEFAULT FALSE` (backward-compatible with prod LML running pre-510 code) plus the rebuild + dedup-swap + prune-swap path updates so a tombstoned id can't survive a pipeline refresh.
2. WXYC/wxyc-shared#174 — adds `not_found` to `DiscogsReleaseMetadata` in 1.12.0. `ArtistDetails` is LML-local (per the `discogs/models.py:137` comment) so the artist side grows the field directly in this repo without an api.yaml change.

LML PR pins the new wxyc-shared release, runs `scripts/generate_api_models.sh`, then consumes the new column.

Sequenced after #518 / #519 (release_id=0 sentinel guard at the orchestrator) so this PR's tombstoning doesn't mask the orchestrator bug; that PR is in main as of `3b45379`, this PR rebases onto it cleanly.

## Why this shape (vs. an alternative)

This is the third instance of the same discriminator-column shape — `release.artwork_checked_at` (WXYC#414), `artist.fetched_at` (#503), now `not_found` on both. The codebase has converged on "ambiguity resolved by a state column on the existing row" twice in 90 days; doing it a third time deepens the pattern rather than introducing a parallel `negative_*` table + 4 fallthrough hooks at every call site (~250 LOC alternative the issue's first revision proposed and the reframe rejected).

A sentinel encoded as `name = ""` would leak empty `ArtistDetails` into `get_artist_details_bulk` with no way to distinguish from a real empty-aliased artist. An explicit `not_found` column keeps the bulk path's behavior unambiguous (it still returns tombstones as empty-fields parallel to #503's stub policy, but the policy is now explicit at the schema level).

The seam's `pg_negative_check` / `pg_negative_record` quadrant exists for *search* negatives (where the empty answer is structurally different from the positive answer). For a single-entity 404, `None` IS the negative answer; routing it through a state column on the existing row is the right level.

## Deferred (with reasons)

- `get_master` / `get_label` (`discogs/service.py:992`, `:969`) have the same `_api_fetch`-collapses-404-to-None shape but no PG cache backing — no schema to land tombstones in. Reconsider when either method grows a cache leg or shows up in a tight-loop caller. Documented in the `fallthrough.py:24-25` policy table comment as the design constraint.
- TTL on negatives — admin endpoint covers the incident-driven false-tombstone case; reconsider if telemetry shows sustained churn.
- L1 caching `None` for tombstoned ids — every tombstone hit still pays one PG round-trip (the main win vs. status quo, but not a "free" L1 hit); relax `@async_cached`'s `None`-skip if `pg_cache_hit - tombstone_returned` traces show the PG round-trip cost is material.

## Test plan

- [x] 15 unit tests in `tests/unit/test_discogs_404_tombstone.py`: 404 discrimination, telemetry counters, 5xx passthrough no-tombstone, log level promoted to error, public boundary translation, CachedOnlyResolver guards.
- [x] 10 unit tests in `tests/unit/test_admin_tombstone_router.py`: auth, three-way response code mapping, L1 eviction, 503 on missing pool, 422 on unknown entity type.
- [x] 14 PG integration tests in `tests/integration/test_cache_service_tombstones.py`: read-path short-circuit on `not_found = TRUE`, hydrated-parent identifier preservation across tombstone overwrite, child-cascade preservation, recovery on 200 (in both directions), bulk-path returns tombstones as empty-fields per #503's stub policy, admin DELETE three-outcome behavior.
- [x] Existing `tests/unit/test_cache_service.py` updated for the new SELECT projection — all 71 still green.
- [x] `ruff check` + `ruff format --check` clean across all touched files.
- [x] Full LML unit sweep: 2248 passed (the one flaky `Event loop is closed` error is unrelated, passes in isolation).
- [ ] Staging integration tests pass once discogs-etl migration is live in staging PG (this PR's integration tests need the column live).

Closes #510.

### DIFF
--- a/discogs/cache_service.py
+++ b/discogs/cache_service.py
@@ -496,13 +496,34 @@ class DiscogsCacheService:
         """
         try:
             release_row = await self.pool.fetchrow(
-                "SELECT id, title, release_year, artwork_url, released, artwork_checked_at "
+                "SELECT id, title, release_year, artwork_url, released, "
+                "artwork_checked_at, not_found "
                 "FROM release WHERE id = $1",
                 release_id,
             )
 
             if release_row is None:
                 return None
+
+            # LML#510: tombstone short-circuit. The parent row's `not_found`
+            # flag is authoritative — there are no child rows for a
+            # tombstone (the tombstone branch in write_release skips the
+            # cascade), so the 4-7 PG round-trips below would all be empty
+            # selects. Returning a tombstone-shaped model keeps the
+            # `is_pg_hit` predicate at the public boundary happy (the
+            # `artwork_checked_at` stamp is set, so the seam treats it as
+            # a hit) and the public method's tombstone translation
+            # converts it back to None for the caller.
+            if release_row["not_found"]:
+                return ReleaseMetadataResponse(
+                    release_id=release_id,
+                    title="",
+                    artist="",
+                    release_url=f"https://www.discogs.com/release/{release_id}",
+                    not_found=True,
+                    artwork_checked_at=release_row["artwork_checked_at"],
+                    cached=True,
+                )
 
             # Fetch all child tables in parallel (independent queries)
             artist_rows, label_rows, track_rows, track_artist_rows = await asyncio.gather(
@@ -670,6 +691,36 @@ class DiscogsCacheService:
             CacheUnavailableError: If database is unreachable
         """
         try:
+            # LML#510: tombstone branch. A 404-shaped write doesn't have any
+            # rich content to cascade — and crucially, we must NOT clobber
+            # any prior hydrated row's identifier columns (title / year /
+            # artwork) nor wipe the child cascade. The narrow UPSERT below
+            # only touches `not_found` + `artwork_checked_at`; the
+            # ``ON CONFLICT DO UPDATE SET`` clause intentionally OMITS
+            # title / year / artwork_url / released so a 404 after a 200
+            # leaves the parent's identifier columns intact, and the early
+            # `return` skips the child DELETE+INSERT cascade so existing
+            # `release_artist` / `release_label` / `release_track` /
+            # `release_track_artist` / `release_genre` / `release_style` /
+            # `release_video` rows survive a tombstone overwrite.
+            if release.not_found:
+                async with self.pool.acquire() as conn:
+                    await conn.execute(
+                        """
+                        INSERT INTO release (
+                            id, title, release_year, artwork_url, released,
+                            artwork_checked_at, not_found
+                        )
+                        VALUES ($1, '', NULL, NULL, NULL, now(), TRUE)
+                        ON CONFLICT (id) DO UPDATE SET
+                            artwork_checked_at = EXCLUDED.artwork_checked_at,
+                            not_found = TRUE
+                        """,
+                        release.release_id,
+                    )
+                logger.debug(f"Tombstoned release {release.release_id}")
+                return
+
             async with self.pool.acquire() as conn, conn.transaction():
                 # Wrap the entire DELETE+INSERT cascade in a single transaction.
                 # Without this, asyncpg autocommits each statement and a
@@ -686,19 +737,24 @@ class DiscogsCacheService:
                 # avoid re-fetching genuinely-imageless releases
                 # (WXYC/library-metadata-lookup#423, backed by the schema
                 # column from WXYC/discogs-etl#239).
+                #
+                # `not_found = FALSE` in the SET clause (LML#510) clears any
+                # prior tombstone in one statement so a recovered 200 makes
+                # the row reachable again without an admin intervention.
                 await conn.execute(
                     """
                     INSERT INTO release (
                         id, title, release_year, artwork_url, released,
-                        artwork_checked_at
+                        artwork_checked_at, not_found
                     )
-                    VALUES ($1, $2, $3, $4, $5, now())
+                    VALUES ($1, $2, $3, $4, $5, now(), FALSE)
                     ON CONFLICT (id) DO UPDATE SET
                         title = EXCLUDED.title,
                         release_year = EXCLUDED.release_year,
                         artwork_url = EXCLUDED.artwork_url,
                         released = EXCLUDED.released,
-                        artwork_checked_at = EXCLUDED.artwork_checked_at
+                        artwork_checked_at = EXCLUDED.artwork_checked_at,
+                        not_found = FALSE
                     """,
                     release.release_id,
                     release.title,
@@ -986,13 +1042,29 @@ class DiscogsCacheService:
             # predicate can distinguish stub rows (rebuild-created, never
             # hydrated from Discogs) from rows we've actually fetched. See
             # `DiscogsService.get_artist_details` and WXYC#502.
+            #
+            # `not_found` (LML#510) is the tombstone discriminator — see
+            # the short-circuit just below.
             artist_row = await self.pool.fetchrow(
-                "SELECT id, name, profile, image_url, fetched_at FROM artist WHERE id = $1",
+                "SELECT id, name, profile, image_url, fetched_at, not_found "
+                "FROM artist WHERE id = $1",
                 artist_id,
             )
 
             if artist_row is None:
                 return None
+
+            # LML#510: tombstone short-circuit. Same rationale as the
+            # release path — no child rows exist for a tombstone, so
+            # the asyncio.gather below would just be empty selects.
+            if artist_row["not_found"]:
+                return ArtistDetails(
+                    artist_id=artist_id,
+                    name="",
+                    not_found=True,
+                    fetched_at=artist_row["fetched_at"],
+                    cached=True,
+                )
 
             # Fetch all child tables in parallel (independent queries)
             alias_rows, nv_rows, member_rows, url_rows = await asyncio.gather(
@@ -1143,21 +1215,46 @@ class DiscogsCacheService:
             CacheUnavailableError: If database is unreachable
         """
         try:
+            # LML#510: tombstone branch. Same shape as write_release —
+            # narrow UPSERT, early return before the child cascade so a
+            # 404 doesn't wipe rich catalog metadata; `ON CONFLICT DO
+            # UPDATE SET` intentionally omits name / profile / image_url
+            # so a 404 after a 200 leaves identifier columns intact.
+            if details.not_found:
+                async with self.pool.acquire() as conn:
+                    await conn.execute(
+                        """
+                        INSERT INTO artist (id, name, profile, image_url, fetched_at, not_found)
+                        VALUES ($1, '', NULL, NULL, now(), TRUE)
+                        ON CONFLICT (id) DO UPDATE SET
+                            fetched_at = now(),
+                            not_found = TRUE
+                        """,
+                        details.artist_id,
+                    )
+                logger.debug(f"Tombstoned artist {details.artist_id}")
+                return
+
             async with self.pool.acquire() as conn, conn.transaction():
                 # Wrap the UPSERT + 4×(DELETE+INSERT) cascade in a single
                 # transaction so a cancellation mid-write doesn't leave the
                 # artist row updated with empty child tables. See WXYC#375.
 
-                # Upsert artist row
+                # Upsert artist row.
+                #
+                # `not_found = FALSE` in the SET clause (LML#510) clears
+                # any prior tombstone in one statement, so a recovered 200
+                # makes the row reachable again without admin intervention.
                 await conn.execute(
                     """
-                    INSERT INTO artist (id, name, profile, image_url, fetched_at)
-                    VALUES ($1, $2, $3, $4, now())
+                    INSERT INTO artist (id, name, profile, image_url, fetched_at, not_found)
+                    VALUES ($1, $2, $3, $4, now(), FALSE)
                     ON CONFLICT (id) DO UPDATE SET
                         name = EXCLUDED.name,
                         profile = EXCLUDED.profile,
                         image_url = EXCLUDED.image_url,
-                        fetched_at = now()
+                        fetched_at = now(),
+                        not_found = FALSE
                     """,
                     details.artist_id,
                     details.name,
@@ -1223,6 +1320,49 @@ class DiscogsCacheService:
         except Exception as e:
             logger.error(f"Cache write_artist_details failed: {e}")
             raise CacheUnavailableError(f"Cache write_artist_details failed: {e}") from e
+
+    async def delete_tombstone(self, entity_type: str, entity_id: int) -> str:
+        """Delete a tombstoned row so the next API call re-fetches it.
+
+        Backs the admin recovery endpoint (`DELETE /admin/discogs/tombstone/...`).
+        The `WHERE id = $1 AND not_found = TRUE` guard means a real row can
+        never be deleted through this surface — even with a typo'd id.
+
+        Args:
+            entity_type: `"release"` or `"artist"`.
+            entity_id: Discogs id.
+
+        Returns:
+            ``"deleted"`` — a tombstoned row matched and was removed.
+            ``"exists_not_tombstone"`` — `entity_id` exists but
+                ``not_found = FALSE``; refusing to delete (real data).
+            ``"not_found"`` — no row at all.
+        """
+        if entity_type not in ("release", "artist"):
+            raise ValueError(f"entity_type must be 'release' or 'artist', got {entity_type!r}")
+        table = entity_type
+        try:
+            async with self.pool.acquire() as conn:
+                # Probe the row's existence and tombstone state in one query
+                # so the response code is unambiguous. A separate DELETE …
+                # RETURNING would conflate "no row" and "row not tombstone".
+                row = await conn.fetchrow(
+                    f"SELECT not_found FROM {table} WHERE id = $1",
+                    entity_id,
+                )
+                if row is None:
+                    return "not_found"
+                if not row["not_found"]:
+                    return "exists_not_tombstone"
+                await conn.execute(
+                    f"DELETE FROM {table} WHERE id = $1 AND not_found = TRUE",
+                    entity_id,
+                )
+            logger.info("Tombstone deleted: %s/%s", entity_type, entity_id)
+            return "deleted"
+        except Exception as e:
+            logger.error("Cache delete_tombstone failed: %s", e)
+            raise CacheUnavailableError(f"Cache delete_tombstone failed: {e}") from e
 
     async def validate_track_on_release(
         self, release_id: int, track: str, artist: str
--- a/discogs/markup_parser.py
+++ b/discogs/markup_parser.py
@@ -466,14 +466,23 @@ class CachedOnlyResolver:
     async def resolve_artist(self, id: int) -> str | None:
         try:
             details = await self._cache.get_artist_details(id)
-            return details.name if details else None
+            # LML#510: tombstones carry `name = ""` as an identifier sentinel.
+            # Without this guard the markup parser emits an `artistLink` token
+            # with `display_name=""` for any tombstoned id — not None, which
+            # is what every consumer here expects.
+            if details is None or details.not_found:
+                return None
+            return details.name
         except Exception:
             return None
 
     async def resolve_release(self, id: int) -> str | None:
         try:
             release = await self._cache.get_release(id)
-            return release.title if release else None
+            # LML#510: same tombstone guard as resolve_artist.
+            if release is None or release.not_found:
+                return None
+            return release.title
         except Exception:
             return None
 
--- a/discogs/models.py
+++ b/discogs/models.py
@@ -152,6 +152,15 @@ class ArtistDetails(BaseModel):
     # the API returned a profile. Used as the cache-hit discriminator in
     # `DiscogsService.get_artist_details` (#502).
     fetched_at: datetime | None = None
+    # Tombstone marker for Discogs 404s on `get_artist_details` (#510).
+    # `True` means LML hit the live API for this id and got a 404; subsequent
+    # reads short-circuit on this flag so callers don't re-burn the
+    # rate-limit budget on the same 404. Tombstone rows carry `name = ""` and
+    # otherwise-default fields — consumers of the public `DiscogsService`
+    # surface never see them (the boundary translates to `None`), but direct
+    # `cache_service` callers (`CachedOnlyResolver`, `get_artist_details_bulk`)
+    # must explicitly guard.
+    not_found: bool = False
     cached: bool = False
 
 
--- a/discogs/router.py
+++ b/discogs/router.py
@@ -30,7 +30,9 @@ _REFRESH_DESCRIPTION = (
     "When true, evict the in-memory L1 cache entry for this entity before "
     "fetching. Forces re-traversal of L2 (PG) and L3 (Discogs API), which is "
     "needed after a manual fix to the underlying PG cache row that would "
-    "otherwise stay invisible until L1's TTL expires."
+    "otherwise stay invisible until L1's TTL expires. NOTE: does NOT clear "
+    "LML#510 tombstones (`not_found = TRUE`) in L2 — use "
+    "`DELETE /admin/discogs/tombstone/{type}/{id}` for that recovery surface."
 )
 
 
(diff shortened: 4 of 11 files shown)


jakebromberg:
**Codegen Freshness check expected to fail until WXYC/wxyc-shared#174 merges**

The check regenerates `generated/api_models.py` from `wxyc-shared` main and diffs against the committed file. This PR's committed `generated/api_models.py` includes the new `not_found` field on `DiscogsReleaseMetadata`, which lives in WXYC/wxyc-shared#174 (not yet on main). Per the issue's land order:

1. WXYC/discogs-etl#277 — schema (merged or pending).
2. WXYC/wxyc-shared#174 — `not_found` on the OpenAPI schema (1.12.0). **This needs to merge before this PR's Codegen Freshness check turns green.**
3. This PR — pins the new wxyc-shared release.

After wxyc-shared#174 merges, the codegen check on this PR should turn green on the next re-run.

jakebromberg:
### Code review

Issue #525, opened 2026-06-09
feat(cache): POST /api/v1/cache/refresh-for-identities — source-agnostic cache backfill by identity_id


jakebromberg:
## Problem

Today there is no source-agnostic "warm LML's cache for these canonical releases" surface. Callers that want to back-fill the Discogs cache for a known set of releases — most concretely [WXYC/Backend-Service#1361](https://github.com/WXYC/Backend-Service/issues/1361) and the cron landed in [BS PR #1376](https://github.com/WXYC/Backend-Service/pull/1376) — have to hit `/api/v1/discogs/release/{discogs_release_id}` and `/api/v1/discogs/artist/{discogs_artist_id}` directly. That works today but bakes in three coupling problems:

1. **BS has to hold Discogs IDs.** `rotation.discogs_release_id` (per the [BS#1029](https://github.com/WXYC/Backend-Service/issues/1029) schema) is the only handle a BS-side cron can give to LML. Discogs IDs aren't stable across Discogs admin reorganizations — the [2026-04-28 reshuffle](https://github.com/WXYC/Backend-Service/issues/624) is documented prior art that broke pinned IDs in LML's own tests.
2. **No multi-source fan-out.** Once [LML#216](https://github.com/WXYC/library-metadata-lookup/issues/216) / [#217](https://github.com/WXYC/library-metadata-lookup/issues/217) / [#218](https://github.com/WXYC/library-metadata-lookup/issues/218) populate MusicBrainz/Wikidata/Spotify legs of the reconciliation log, a caller that wants "warm everything we know about this release" still has no single endpoint to ask. They'd have to enumerate `(source, external_id)` pairs themselves, which inverts the dependency direction.
3. **Walk-to-artists belongs in LML.** Today BS extracts `release.artists[*].artist_id` after fetching the release record (BS PR #1376's `extractPhase1ArtistIds`). LML already has the release record after refresh; making the caller re-do the walk is wasted shape. (The id=0 sentinel defense — `isValidArtistId` in `jobs/rotation-artist-backfill/lml-fetch.ts` — is handled separately by [LML#518](https://github.com/WXYC/library-metadata-lookup/issues/518)'s caller-side discipline and [LML#546](https://github.com/WXYC/library-metadata-lookup/issues/546)'s audit. See "Sentinel defense" section below — the dispatcher's walk-to-artists step is itself a new caller-shaped surface and gets its own guard in the extractor.)

## Precursor

[...]

Issue #175, opened 2026-04-27
Add Bandcamp branch to /api/v1/releases/resolve


jakebromberg:
## Background

Follow-up to PR 5a (#174 — Discogs branch). Adds Bandcamp URL support to `POST /api/v1/releases/resolve`: paste a Bandcamp album URL and get back canonical fields, identifiers, and the streaming snapshot, just like the Discogs path.

This is **PR 5b**. It's stacked on the 5a branch — merge 5a first, then this rebases cleanly onto main.

## Module additions

- `release/bandcamp_resolver.py` — fetches the album page via the existing rate-limited `BandcampClient`, parses the embedded JSON-LD `MusicAlbum` blob. Tolerates `@type` as a string or a list (schema.org allows both).
- `tests/fixtures/bandcamp/album_self_released.html` and `album_with_label.html` — committed fixtures so the parser tests don't hit live Bandcamp.

## Orchestrator changes

- Bandcamp branch dispatches to `resolve_bandcamp_album`.
- Streaming check short-circuits the Bandcamp leg when the input is a Bandcamp URL (sets `bandcamp=None`, then pins `sources.bandcamp` to the input URL with confidence 100). Saves 2+ rate-limited HTTP calls per paste.
- Identity write-back gains a `bandcamp_id` parameter, populated from the URL's slug — matches what `bandcamp_pipeline.py` already writes.

## Self-released handling

When publisher subdomain == artist subdomain, Bandcamp lists the artist itself as the publisher. The parser detects that and returns `label: null` per the music director's instruction (don't pretend the artist is the label).

## Testing

15 new tests (1490 total passing):

- `test_bandcamp_resolver.py` — parser against both fixtures (self-released, label account), HTTP fetch path with mocked `BandcampClient`, error paths.
- `test_orchestrator.py` — Bandcamp short-circuit (kwarg inspection both directions), fetch-failure → warning, slug write-back.

The `TestBandcampNotYetSupported` placeholder from 5a is replaced with the real `TestBandcampBranch`.

`pytest tests/unit/`: 1490 passed. `ruff check`, `ruff format --check`, `mypy` all clean.

PR #530, opened 2026-06-10
feat(identity): release-level identity layer (LML#526)

### LINKED ISSUE #526: feat(identity): release-level identity — schema, mint-on-resolve, reconciliation log shape
## Problem

[WXYC/library-metadata-lookup#525](https://github.com/WXYC/library-metadata-lookup/issues/525) ("refresh-for-identities") and [WXYC/Backend-Service#1380](https://github.com/WXYC/Backend-Service/issues/1380) (`rotation.lml_identity_id`) both assume LML can mint and resolve a stable identity for a **release**. It can't today:

- `POST /identity/resolve` doesn't exist. `GET /identity/resolve?name=<artist>` exists (`identity/router.py:82-105`) but it's artist-name-keyed only — no `source` / `external_id` / `kind` parameters.
- `entity.identity` is per-artist. Unique key is `library_name TEXT NOT NULL`; columns are artist-source IDs (`discogs_artist_id`, `wikidata_qid`, `musicbrainz_artist_id`, `spotify_artist_id`, `apple_music_artist_id`, `bandcamp_id`). No release columns, no `kind` discriminator.
- `entity.reconciliation_log` rows are `(identity_id, source, external_id)` where `external_id` is an artist-side ID. Walking `identity_id` returns artist IDs only.
- `POST /api/v1/releases/resolve` (`release/router.py:24-44`) is the closest release-shaped endpoint, but it's stateless prefill for tubafrenzy's paste form — it returns `ReleaseIdentifiers` and persists nothing. No identity row is minted.

So the LML#525 / BS#1380 dependency chain hits a wall: LML#525's "resolve `identity_id` → `(source, external_id)` pairs" returns artist IDs today, and BS#1380's 1:1 swap of `rotation.discogs_release_id` → `rotation.lml_identity_id` has no release identity to point at.

This issue establishes the release-identity layer that the rest of the cross-cache-identity initiative depends on.

### Schema

A persistent record per canonical release that unifies known IDs across sources (Discogs release, Discogs master, MusicBrainz release, Spotify album, Apple Music album, Bandcamp album URL).

Recommended shape: a parallel table `entity.release_identity` with release-specific columns, plus a parallel `entity.release_reconciliation_log` (or extend the existing `entity.reconciliation_log` to discriminate by entity kind via a new `entity_kind` column). Reasons for parallel-table over a `kind`-discriminator on `entity.identity`:

- `entity.identity.library_name TEXT NOT NULL UNIQUE` is an artist-name constraint that doesn't apply to releases; squeezing both kinds into one table either nulls the constraint (loses safety on the artist side) or repurposes the column with confusing semantics.
- Release columns (discogs_release_id, discogs_master_id, mb_release_id, spotify_album_id, apple_music_album_id, bandcamp_album_url) don't overlap with artist columns; one table per shape is cleaner.

The implementer is free to pick the discriminator path if a deeper read of the migration surface argues for it — call the choice out in the migration PR.

#### Schema-migration ownership

The canonical DDL for `entity.identity` and `entity.reconciliation_log` is not owned by this repo — the live schema is created and maintained by the [discogs-cache](https://github.com/WXYC/discogs-etl) ETL pipeline, which populates the PG database LML reads from. LML carries the schema only as test fixtures (`tests/integration/test_entity_resolution.py:63-90`, `tests/integration/test_bulk_resolve_libraries.py:43-69`, `tests/integration/test_charset_torture_entity_identity.py:67`).

That means a "migration" for the new release-identity tables has two landing spots and the implementer must decide upfront:

- **Option A (mirrors current setup):** add canonical DDL in discogs-cache; add matching test fixtures in LML. Two PRs across two repos, but consistent with how `entity.identity` is owned today.
- **Option B (relocate ownership):** bring the canonical DDL into this repo (e.g. an Alembic / hand-managed `migrations/` directory under `entity/`) and have discogs-cache stop creating these tables. One repo, but a larger migration-ownership decision that affects the artist-side tables too and probably wants its own ADR.

Pick A unless there's a strong reason to do B — and call the choice out in the migration PR.

### Endpoint

A new `POST /api/v1/identity/resolve` that accepts a `(kind, source, external_id)` triple and returns (or mints and returns) an `identity_id`:

```
POST /api/v1/identity/resolve
Body: { "kind": "release", "source": "discogs_release", "external_id": "12345" }
Response: { "identity_id": int, "kind": "release", "minted": bool }
```

Behavior:

- **Idempotent.** Same `(kind, source, external_id)` always returns the same `identity_id` (with `minted: false` on subsequent calls).
- First call mints. Subsequent calls look up.
- Validation gate: reject sentinel / malformed inputs *before* mint. No poisoned identities. Per-source rules for v1:
  - `discogs_release`: reject `external_id <= 0` (Discogs uses `0` for the unknown-release sentinel; negative IDs are never valid).
  - `discogs_master`: reject `external_id <= 0`, same reasoning.
  - `bandcamp`: `external_id` is the canonical album URL; reject if it does not parse as a Bandcamp album URL via `release/url_parser.py` (i.e. `parse_bandcamp_url()` returns None or non-album).
  - Any future source added here owes a sentinel rule in the same PR.
- Authn: `_lml_protected` like the rest of `/api/v1`.
- Concurrency-safe: two concurrent resolve calls with the same input converge on one row (unique index + upsert, or advisory lock — implementer's call).

Scope for v1: `kind: 'release'` only. Sources accepted: at minimum `discogs_release`, `discogs_master`, `bandcamp` (matching `release/models.py` `ReleaseResolveRequest.source` Literal). The artist case (`kind: 'artist'` keyed by `(source, external_id)`) is useful for symmetry but is not strictly required by LML#525 or BS#1380 — file as follow-up.

Keep the existing `GET /identity/resolve?name=<artist>` as the artist-name-keyed lookup. It serves a different call site and shouldn't move.

### Reconciliation rows

When the resolver mints a new release identity, it writes the source row it was minted from into the (release) reconciliation log. **An idempotent re-resolve (returning `minted: false`) writes no new reconciliation row** — the resolver only logs the act of minting, not the act of looking up. This differs from `merge_identity_by_library_name` on the artist side, which writes a reconciliation row on every join because the join itself carries new evidence; resolve here carries no new evidence on a repeat call.

Future cross-source joins — e.g. a MusicBrainz release_id matched to a known Discogs release_id — attach to the same `identity_id` and write their own reconciliation rows at the time of the join. The joining logic itself is out of scope for this issue (it lives under [LML#207](https://github.com/WXYC/library-metadata-lookup/issues/207) and its source-leg children); this issue only needs to lay the row shape down.

## Where

- `entity/` — schema migration for `entity.release_identity` (and either the parallel reconciliation table or a discriminator column on the existing one). See "Schema-migration ownership" above — the canonical DDL likely lives in discogs-cache, not here, and the LML-side change is test fixtures only.
- `entity/store.py` — new `get_release_identity_by_source(source, external_id)` and `mint_release_identity(...)` helpers. Or extend existing identity functions if the discriminator path wins.
- `identity/router.py` — new `POST /api/v1/identity/resolve` registered on `api_v1_router` (the protected `/api/v1/...` surface), not on the open `router` that hosts the existing `GET /identity/resolve`. The two endpoints live at different URLs (`/api/v1/identity/resolve` vs `/identity/resolve`) and on different routers — they only share a path suffix, not a base path.
- `identity/models.py` — request/response models for the polymorphic resolve.
- `wxyc-shared/api.yaml` — lock the new endpoint shape so Backend's generated types track changes.
- `tests/integration/test_entity_resolution.py` — fixtures and assertions for release-identity mint + idempotent re-resolve + sentinel rejection + concurrent-mint convergence.
- `docs/api-endpoints.md` — document the new endpoint and link to the release-identity model.

## Constraints

- Existing `GET /identity/resolve?name=<artist>` stays as-is.
- The new endpoint is `_lml_protected`.
- No identity is minted for invalid `(source, external_id)` pairs. Validation runs before any write.
- Cross-source reconciliation join logic (matching a MusicBrainz release to a Discogs release identity) is out of scope here. The schema must support it; the joining belongs to [LML#207](https://github.com/WXYC/library-metadata-lookup/issues/207) and its children.
- Migration is additive. Existing `entity.identity` and `entity.reconciliation_log` stay untouched on the artist side.

## Acceptance criteria

- [ ] Migration adds the release-identity schema (parallel table or discriminator — pick one, document the choice).
- [ ] `POST /api/v1/identity/resolve` accepts `{kind, source, external_id}` and returns `{identity_id, kind, minted}`.
- [ ] First call with new input mints and returns `minted: true`. Repeat call returns same `identity_id` with `minted: false`.
- [ ] Invalid sentinel input (Discogs `external_id == 0`, malformed Bandcamp URL, etc.) rejected with 422 before any DB write.
- [ ] Endpoint is `_lml_protected`.
- [ ] Reconciliation row is written on mint, linking `identity_id` to its `(source, external_id)` of origin.
- [ ] Integration tests cover: mint, idempotent re-resolve, sentinel rejection, concurrent re-resolve (no duplicate rows under race).
- [ ] Response shape locked in `wxyc-shared/api.yaml`.
- [ ] `docs/api-endpoints.md` updated.

## Notes for implementer

- TDD per project CLAUDE.md: red test first, then schema, then endpoint, then green.
- The concurrent-mint test matters — two concurrent paste flows resolving the same URL must converge on one `identity_id`. Whether that's enforced via unique index + upsert or advisory lock is the implementer's call, but the test is non-negotiable.
- The artist-case follow-up — `POST /api/v1/identity/resolve` accepting `{kind: 'artist', source, external_id}` — would let LML#525 fan out symmetrically. File alongside if scope allows; otherwise defer.

## Out of scope

- Cross-source reconciliation joining (matching MusicBrainz release X to Discogs release Y under the same identity). Lives under [LML#207](https://github.com/WXYC/library-metadata-lookup/issues/207) and its source-leg children.
- Artist-side `(source, external_id)`-keyed resolution. Existing `GET /identity/resolve?name=<artist>` covers the name-keyed artist case; the source-keyed artist case is a follow-up if needed.
- Backfilling existing release records into the new identity table. It starts empty and grows as paste-flow and other consumers call resolve.

## Related

- [WXYC/library-metadata-lookup#525](https://github.com/WXYC/library-metadata-lookup/issues/525) — refresh-for-identities multiplexer. **Blocked on this.**
- [WXYC/library-metadata-lookup#207](https://github.com/WXYC/library-metadata-lookup/issues/207) — reconciliation pipeline parent. Cross-source release joins land under its umbrella once this schema exists.
- [WXYC/library-metadata-lookup#216](https://github.com/WXYC/library-metadata-lookup/issues/216) / [#217](https://github.com/WXYC/library-metadata-lookup/issues/217) / [#218](https://github.com/WXYC/library-metadata-lookup/issues/218) — MusicBrainz / Wikidata / Spotify source legs. Will populate release-side reconciliation rows once they grow release-aware projections.
- [WXYC/Backend-Service#1380](https://github.com/WXYC/Backend-Service/issues/1380) — `rotation.lml_identity_id` schema migration. **Blocked on this.**
- [WXYC/Backend-Service#1381](https://github.com/WXYC/Backend-Service/issues/1381) — cron migration. Blocked on LML#525, which is blocked on this.

jakebromberg:
## Summary

Adds `POST /api/v1/identity/resolve` — the release-side counterpart to the existing artist-side `entity.identity` surface. Closes the blocker that [LML#525](https://github.com/WXYC/library-metadata-lookup/issues/525) (refresh-for-identities) and [BS#1380](https://github.com/WXYC/Backend-Service/issues/1380) (`rotation.lml_identity_id`) both depend on.

- Two new tables — `entity.release_identity` and `entity.release_reconciliation_log` — alongside the existing artist-side schema, untouched.
- Idempotent mint: same `(kind, source, external_id)` triple always returns the same `identity_id`; subsequent calls return `minted: false` and write **no** new reconciliation row.
- Concurrency-safe: two simultaneous mints with the same input converge on one row via per-source `UNIQUE` constraints + `INSERT ... ON CONFLICT DO NOTHING RETURNING id` + fallback `SELECT`, all in one transaction.
- Validation runs **before any DB write**: Discogs `external_id <= 0` (covers the `0` sentinel and negative IDs) and malformed / non-album Bandcamp URLs are rejected with 422 — no poisoned identity rows can land. Bandcamp URLs are canonicalised via the existing `release.url_parser.parse_url` so trailing-slash variants collapse to one row.
- Endpoint inherits `_lml_protected` from `main.py` via `api_v1_router`. v1 accepts `kind: release` only; `kind: artist` symmetric extension is the documented follow-up (will not break the wire shape).

## Schema-migration ownership

Option A from the issue: the runtime DDL is owned by [discogs-cache](https://github.com/WXYC/discogs-etl), mirroring how `entity.identity` is owned today. This PR carries the new tables as **test fixtures only** plus a one-page DDL reference in `entity/release_identity.sql` for the sibling discogs-cache PR to copy verbatim. Until that lands, the endpoint returns 503 in prod — same posture as `entity.identity` missing.

### Sibling PRs needed

1. **discogs-cache** — apply `entity/release_identity.sql` to the prod schema. The two `CREATE TABLE IF NOT EXISTS` statements are idempotent.
2. **wxyc-shared** — append the `IdentityKind` / `ReleaseIdentitySource` / `ReleaseIdentityResolveRequest` / `ReleaseIdentityResolveResponse` schemas and the `/api/v1/identity/resolve` path to `api.yaml`. (The schemas were defined inline in `generated/api_models.py` here because the local wxyc-shared sibling was 19 commits behind origin/main at codegen time and regenerating against the stale snapshot dropped types the rest of the codebase depends on.)

The LML PR can merge first: its tests run against locally-applied fixtures, prod just degrades to 503 until the discogs-cache PR lands.

## Files

| Path | Change |
|---|---|
| `identity/release_validation.py` | New — pure validation + coercion functions; the `RELEASE_SOURCE_COLUMN` dict, `InvalidReleaseExternalIdError`, `validate_and_canonicalize_external_id`, `coerce_external_id` |
| `entity/store.py` | Adds `mint_or_get_release_identity`, `get_release_identity_by_source`, `log_release_reconciliation` + `_LOG_RELEASE_RECONCILIATION_SQL` |
| `identity/router.py` | Adds `POST /resolve` handler on `api_v1_router` |
| `generated/api_models.py` | Hand-edited additions for the new request/response shapes (regen-equivalent) |
| `entity/release_identity.sql` | Canonical DDL reference |
| `tests/unit/test_release_validation.py` | 31 unit tests for validation rules + coercion |
| `tests/integration/test_release_identity.py` | 20 integration tests: store-level mint/idempotent re-mint/concurrent-mint, endpoint-level mint/idempotent/sentinel rejection/auth |
| `docs/api-endpoints.md`, `CLAUDE.md` | Topic-guide updates |

## Test plan

- [x] `pytest tests/unit/test_release_validation.py` — 31 passed
- [x] `pytest -m pg tests/integration/test_release_identity.py` — 20 passed (against local Docker PG on port 5433)
- [x] `pytest -m "not pg and not external_api"` — 2653 passed (one pre-existing teardown flake in `test_no_module_level_asyncio_lock_in_dependencies` unrelated to this change; passes in isolation)
- [x] `ruff format --check . && ruff check .` — clean
- [x] Manual curl smoke through the running app:
  - First `POST /api/v1/identity/resolve` with `{"kind": "release", "source": "discogs_release", "external_id": "12345"}` → `{"identity_id": N, "kind": "release", "minted": true}`
  - Same again → `{"identity_id": N, "kind": "release", "minted": false}`
  - `"external_id": "0"` → 422
  - Bandcamp URL round-trip → mint then idempotent
  - Bandcamp non-URL → 422

Closes #526


jakebromberg:
**CI status: 11/12 green, 1 known failure.**

`Codegen Freshness` is expected to fail on this PR — the cross-repo `wxyc-shared/api.yaml` change is a sibling PR that has to land first for the regen-and-compare gate to pass. Local `wxyc-shared` carried unrelated WIP at the time of this commit that I did not want to disturb, so the api.yaml edits stayed in my local checkout only and the `generated/api_models.py` additions in this PR are hand-edited equivalents of what the regen will produce once the sibling PR lands.

Order to land:
1. Sibling **wxyc-shared** PR — append the four schemas (`IdentityKind`, `ReleaseIdentitySource`, `ReleaseIdentityResolveRequest`, `ReleaseIdentityResolveResponse`) and the `/api/v1/identity/resolve` path to `api.yaml`. The exact body is in the `wxyc-shared` working tree on my machine and was test-rendered into `generated/api_models.py` here.
2. Sibling **discogs-cache** PR — apply `entity/release_identity.sql` to the prod `discogs` schema (two idempotent `CREATE TABLE IF NOT EXISTS` statements).
3. **This PR** — once the wxyc-shared PR lands, rerun `bash scripts/generate_api_models.sh` here; the diff against the hand-edited classes should be empty (or trivial whitespace), Codegen Freshness flips green, merge.

Until the discogs-cache PR lands the endpoint returns 503 in prod — same posture as `entity.identity` missing today.

jakebromberg:
### Code review

Found 2 issues, both fixed in 5ad8cf7.

1. WX-3.B boundary policy: the inline reconciliation-log INSERT inside `mint_or_get_release_identity` passed `source` and `external_id` raw to asyncpg, while the public `log_release_reconciliation` helper (and the artist-side `log_reconciliation`) both wrap every TEXT bind in `to_pg_text_form` per the documented WX-3.B mojibake policy (see the `upsert_identity` docstring). First-mint and conflict-fallthrough handled NUL-bytes differently — exactly the asymmetric class of bug the policy warns against.

https://github.com/WXYC/library-metadata-lookup/blob/9c329289f679cf706b1d84be67351981489d34d4/entity/store.py#L723-L734

2. Sentry observability: the new `resolve_release_identity` handler was missing the explicit `sentry_sdk.start_span(op="http.server")` wrapper and entry/exit INFO logs that every other `/api/v1/` handler in `identity/router.py` carries — pattern added in PR #434 specifically because the FastApiIntegration auto-transaction does not reliably land for `/api/v1/` routes (LML#355 audit). Without the explicit span, trace-explorer queries like `op:http.server span.description:*identity/resolve*` would miss this endpoint's traffic.

https://github.com/WXYC/library-metadata-lookup/blob/9c329289f679cf706b1d84be67351981489d34d4/identity/router.py#L352-L389

Compare the fixed handler against the sibling pattern in `bulk_resolve_libraries` at the same module.

jakebromberg:
### Code review

Iteration 2. Found 1 issue, fixed in 652f1c4.

1. WX-3.B comment over-promise: the comment added in 5ad8cf7 says "TEXT-bound arguments at every PG write must go through to_pg_text_form", but the Bandcamp `bind_value` passed to the INSERT, ON CONFLICT, and conflict-fallthrough SELECT skipped the wrap. Practically unreachable (the URL bind comes from `parse_url`'s canonical form which has no NUL bytes by construction) but the comment over-stated what the code delivered.

https://github.com/WXYC/library-metadata-lookup/blob/5ad8cf7aad8d9800bca2b3d701bfa0cdbdc3471c/entity/store.py#L711-L723

Iteration-2 also surfaced two non-issues:
- "`assert row is not None` becomes a 500 under non-READ-COMMITTED isolation" — the pool is configured RC only, the path is unreachable.
- "entry log and validation run outside the Sentry span" — checked the sibling `bulk_resolve_libraries` handler at L225-L246; it has the same structure (auth + entry log before the span open). The new handler matches the established pattern.

jakebromberg:
### Code review

Iteration 3. Found 3 small issues, fixed in bc2a7c8.

1. Docstring precision: `resolve_release_identity` claimed the span "wraps the handler body" but the entry log fires before the span opens (intentional — so a hang inside the store call still leaves a log trace, matching the LML#430 rationale) and the exit log fires after it closes. Rephrased to describe what the code actually does.

https://github.com/WXYC/library-metadata-lookup/blob/652f1c436b22dbee8336df6c6c7f86a62d9f8ece/identity/router.py#L356-L370

2. Test cache isolation: `test_requires_auth_when_enabled` cleared the `@lru_cache`'d `get_settings` before the test but not after. `monkeypatch` restores the env var on teardown, but the cache entry primed with `require_auth=true` survived. Wrap in `try/finally` with `cache_clear()` in both arms.

https://github.com/WXYC/library-metadata-lookup/blob/652f1c436b22dbee8336df6c6c7f86a62d9f8ece/tests/integration/test_release_identity.py#L415-L433

3. Dead import: `import core.auth as auth_module  # noqa: F401  (kept for namespace)` in the same test was copy-paste residue. Removed; the comment misrepresented why it was there.

Also surfaced and dismissed:
- "Commit message 5ad8cf7 is hard-wrapped at ~80 cols" — verified `git show --no-patch 5ad8cf7`, paragraphs are single lines. False positive.
- "`claude-session` trailer violates anonymity" — that's a git-notes trailer added by the harness pre-commit hook, not by the commit author. Rewriting pushed commits to scrub it is destructive; the user controls the hook config.

jakebromberg:
### Code review

Iteration 4. No issues found. Checked for bugs and CLAUDE.md compliance.

Loop converged after 3 fix rounds:
- iter 1 (5ad8cf7): WX-3.B wrap in mint reconciliation log + http.server span on the new handler
- iter 2 (652f1c4): WX-3.B wrap on the Bandcamp bind_value in both store methods
- iter 3 (bc2a7c8): docstring precision + auth test cache isolation + dead `core.auth` import removed

jakebromberg:
Status update on Codegen Freshness:

Sibling PR is open: https://github.com/WXYC/wxyc-shared/pull/175 (1.13.0 — schemas + `/api/v1/identity/resolve` path). Both CI checks pass.

Preflight: I ran the codegen locally against #175's `api.yaml` and diffed the output against the hand-edited `generated/api_models.py` on this branch — byte-identical. Once #175 merges and this PR's Codegen Freshness job re-fetches `wxyc-shared/main`, the gate goes green with no further commits required on this PR.

Blocker: PR #175 is BLOCKED by branch protection requiring 1 approving review (and the author can't self-approve). Needs an approve + rebase-merge there, then a re-run of the Codegen Freshness workflow here.

jakebromberg:
### Code review

Iteration 6. Found 2 small issues, fixed in bb0422f.

1. Drift-test gap: TestRegistryDriftInvariant's docstring named four lockstep places but only three had CI-time coverage — the `coerce_external_id` if-chain would still only fail at first-real-request time if a future source was added without a coerce branch. Added test_every_dict_entry_has_a_coerce_branch.

https://github.com/WXYC/library-metadata-lookup/blob/2a76990f67a37fab44d17320b3bce3416d4cf8fb/tests/unit/test_release_validation.py#L145-L215

2. Undocumented programmer-error raise: iter-5's S6 fix added `raise InvalidReleaseExternalIdError(...)` to `log_release_reconciliation` for unknown source values. That's a `ValueError` subclass, not a `PostgresError`/`OSError`. A future LML#207 joiner wrapping the call in `except (PostgresError, OSError)` would let it escape as a 500 rather than the intended 422. Extended the docstring with an explicit `Raises:` section.

https://github.com/WXYC/library-metadata-lookup/blob/2a76990f67a37fab44d17320b3bce3416d4cf8fb/entity/store.py#L820-L848

Also surfaced and dismissed:
- "commit body hard-wrapped" — verified with `awk '{print length}'` on the raw commit; each paragraph is one 200-460-char line. False positive.
- "`claude-session` notes trailer violates anonymity" — dismissed in iter 3; harness-generated, not author-written.
- "bug scan" returned [] — no new bugs in 2a76990.

jakebromberg:
### Code review

Iteration 7. Found 1 small issue, fixed in 230bee0.

1. Docstring imprecision: test_every_dict_entry_has_a_coerce_branch's comment said the explicit RuntimeError at the bottom of coerce_external_id "fires here" on missing-source drift. The for-loop's `assert source in canonical_inputs` actually fires first with an AssertionError; the RuntimeError only fires for the narrower case where the author added a canonical_inputs entry but not the matching if-branch. Both failure modes catch the same drift — rewrote the docstring to describe both paths so an author following the comment knows what error to expect.

https://github.com/WXYC/library-metadata-lookup/blob/bb0422f108e326d8312b9c8ca6838b527488ad63/tests/unit/test_release_validation.py#L213-L236

jakebromberg:
### Code review

Iteration 8. No issues found. Checked for bugs and CLAUDE.md compliance.

Loop converged after 5 fix rounds (iter 1, 2, 3, 5+6+7 grouped):
- iter 1 (5ad8cf7): WX-3.B wrap in mint log + http.server span on the new handler
- iter 2 (652f1c4): WX-3.B wrap on Bandcamp bind_value in both store methods
- iter 3 (bc2a7c8): docstring precision + auth test cache isolation + dead import removed
- iter 5 (2a76990): xhigh-pass — 9 findings (regex int validator, explicit raises, hoisted imports, log helper, FK index, drift tests, etc.)
- iter 6 (bb0422f): 4th drift test + Raises section on log_release_reconciliation
- iter 7 (230bee0): docstring precision on test_every_dict_entry_has_a_coerce_branch

Tests: 41 unit + 20 pg, ruff clean throughout.

PR #587, opened 2026-06-16
fix(lookup): album-titled request no longer hits 'song not on any album' fallback (LML#586)

### LINKED ISSUE #586: Album-titled request lands in 'song not on any album' fallback — the album is in the result list
## Repro

User typed in Slack: `on patrol, sun araw`.

Bot reply:

> "On Patrol" is not on any album in the library, but here are some albums by Sun Araw:
>
> **Sun Araw** — On Patrol — *Rock cd SU 138/1*

The album "On Patrol" by Sun Araw **is** in the library — it's the album right there in the result list. The context message contradicts its own results.

## Cause

`request-o-matic` parses `"on patrol, sun araw"` as `song="On Patrol"` / `artist="Sun Araw"` / `album=None`.

Pipeline trace in `lookup/orchestrator.py`:

1. `resolve_albums_for_track` (orchestrator.py:784) asks Discogs which releases by Sun Araw contain a **track** called "On Patrol". None do — "On Patrol" is the album title, not a track on it. Returns `albums=[], song_not_found=True`.
2. ARTIST_PLUS_ALBUM strategy falls through to the `artist + song` FTS branch (`search_library_with_fallback`, orchestrator.py:1239–1251), which finds the Sun Araw album "On Patrol" (the album title contains both query terms). Returns `Outcome.artist_fallback(items)`.
3. Step 3b `filter_results_by_track_validation` (orchestrator.py:~3038) tries to confirm "On Patrol" is on the album's tracklist. It isn't. Promotion via `find_library_albums_with_cached_track` also fails. `song_not_found` stays `True`.
4. `build_context_message` (orchestrator.py:2837–2841) fires the "is not on any album" path because `song_not_found=True` and `has_results=True` — even though one of the results' titles literally matches `parsed.song`.

Nothing in the pipeline asks "did the requested track title match an *album* title by this artist?"

## Fix

In step 3b, after `filter_results_by_track_validation` and `find_library_albums_with_cached_track` both come back empty: if any surviving artist-fallback row's title fuzzy-matches `parsed.song` (`rapidfuzz.fuzz.token_set_ratio >= 90`), promote those rows and clear `song_not_found`. Last-resort guard, so it never overrides a confirmed track-on-album result.

Effect on the reported case: `song_not_found=False` → `build_context_message` returns `None` → the bot shows the album with no misleading prefix.

## Acceptance

- Unit test: `song="On Patrol"`, `artist="Sun Araw"`, library has a Sun Araw album titled "On Patrol", Discogs has no track titled "On Patrol" → response has `song_not_found=False` and `context_message=None`.
- Existing track-validation and cache-promotion tests still pass.
- Floor is high enough that an artist-fallback row with a coincidentally-overlapping title (e.g. song "Patrol" against album "On Patrol Reissue") does not silently override the song-not-found path.

## Files

- `lookup/orchestrator.py` — add `_filter_results_by_song_as_album_title` helper alongside `_filter_results_by_album_match`; call from step 3b in `perform_lookup`.
- `tests/unit/test_orchestrator.py` — regression test in `TestPerformLookupFallback`.

jakebromberg:
Closes #586.

## Summary

Slack request `on patrol, sun araw` returned the Sun Araw album "On Patrol" in the result list **and** the message *"On Patrol" is not on any album in the library, but here are some albums by Sun Araw* — about the album sitting in its own result list. Adds a last-resort step-3b branch in `perform_lookup`: when validation + cached-track promotion both fail and a surviving artist-fallback row's title clears 90 against `parsed.song` (rapidfuzz `token_set_ratio`), promote those rows and clear `song_not_found`. `build_context_message` then returns `None` so the bot shows the album with no misleading prefix.

## Why a high floor

The new fallback overrides the song-not-found message based on title-equals-song. A coincidental word overlap (song "Patrol" vs album "On Patrol Reissue") should NOT override a legitimate "couldn't find that track" signal, so `_SONG_AS_ALBUM_TITLE_FLOOR = 90.0`, above the 80 used by `_ALBUM_MATCH_FLOOR`. Same `normalize_for_comparison` + `rapidfuzz.fuzz.token_set_ratio` pair as the existing helper.

## Why step 3b (not pre-pipeline)

The check needs the surviving artist-fallback rows, which only exist after FTS + per-result track validation + cached-track promotion run. A pre-pipeline check would either re-derive that work or fire on rows we'd otherwise reject. The check lives right after the cached-track promotion: it's the same shape ("validation said no, here's a recovery probe") and it shares the `elif song_not_found:` gate so it never overrides a confirmed track-on-album result.

## Test plan

- [x] New regression test in `TestPerformLookupFallback`: `song="On Patrol"`, `artist="Sun Araw"`, library has a Sun Araw album titled "On Patrol", Discogs has no track titled "On Patrol" on any release, cache lookup empty — asserts `song_not_found=False`, `context_message=None`, result list contains the album.
- [x] `tests/unit/test_orchestrator.py` (190 tests) + `tests/unit/test_orchestrator_helpers.py` + `tests/unit/test_orchestrator_gaps.py` all pass.
- [x] Full unit suite: 2618 pass; one pre-existing event-loop teardown flake in `test_dependencies.py` unrelated to this change (passes in isolation).
- [x] `ruff check` + `ruff format --check` clean.

## Files

- `lookup/orchestrator.py` — `_SONG_AS_ALBUM_TITLE_FLOOR` constant + `_filter_results_by_song_as_album_title` helper alongside `_filter_results_by_album_match`; new `else` branch in `perform_lookup` step 3b.
- `tests/unit/test_orchestrator.py` — `test_song_matching_album_title_clears_song_not_found`.

### DIFF
--- a/lookup/orchestrator.py
+++ b/lookup/orchestrator.py
@@ -1198,6 +1198,52 @@ async def _library_miss_discogs_search(
     return library_item, best
 
 
+# Minimum fuzzy score for promoting an artist-fallback row whose *title*
+# matches the requested "song" — i.e. recognising the parsed song as the
+# album the user actually wanted. Set higher than `_ALBUM_MATCH_FLOOR`
+# because the consequence here is asserting the user's intent (album,
+# not track); a borderline match should *not* override the song-not-found
+# message.
+_SONG_AS_ALBUM_TITLE_FLOOR = 90.0
+
+
+def _filter_results_by_song_as_album_title(
+    results: list[LibraryItem],
+    song: str | None,
+) -> list[LibraryItem]:
+    """Pick artist-fallback rows whose title matches the requested song.
+
+    Handles the request shape "on patrol, sun araw" — request-o-matic routes
+    it as ``song="On Patrol"`` / ``artist="Sun Araw"``, but the user typed
+    an album name. The artist+song FTS branch of
+    :func:`search_library_with_fallback` surfaces the matching album because
+    the album title contains the song words; per-result track validation
+    then reasonably comes back empty (no track titled "On Patrol" exists on
+    that album — it IS the album) and ``song_not_found`` stays set,
+    producing the misleading 'not on any album' context message about a
+    result sitting in its own list.
+
+    Floor is ``_SONG_AS_ALBUM_TITLE_FLOOR`` (>= 90 via
+    ``rapidfuzz.fuzz.token_set_ratio``) — high enough that a coincidental
+    word overlap won't override the song-not-found path.
+
+    Returns the subset of ``results`` whose normalised title clears the
+    floor against the normalised song. Empty input or whitespace-only song
+    returns ``[]`` cleanly.
+    """
+    if not song or not song.strip() or not results:
+        return []
+    from rapidfuzz import fuzz
+
+    norm_song = normalize_for_comparison(song)
+    matches: list[LibraryItem] = []
+    for item in results:
+        title_norm = normalize_for_comparison(item.title or "")
+        if fuzz.token_set_ratio(norm_song, title_norm) >= _SONG_AS_ALBUM_TITLE_FLOOR:
+            matches.append(item)
+    return matches
+
+
 async def search_library_with_fallback(
     db: LibraryDB,
     parsed: ParsedRequest,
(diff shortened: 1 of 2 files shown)


jakebromberg:
### Code review

Issue #646, opened 2026-06-20
Row-less carry-through misses the album-title-probe arm (non-library trio / odd-credit case)


jakebromberg:
Follow-up from the #628 review (finding A — in-spec-by-omission, not a #628 bug).

## Problem

The #628 row-less carry-through (`_resolve_nonlibrary_release`) calls `resolve_release_for_track(..., max_validations=5)` **without** `also_probe_album_title=True`. So a non-library release discoverable *only* via the album-title fallback wave (the #319/#237 trio-collaboration / oddly-filed-track-credit shape — e.g. typed "Orcutt Shelley Miller" where Discogs files the per-track credit unusually) is dropped by `process_release` / the kernel **and** missed by the carry-through's track-only probe.

In-library trio releases are still covered (the #604 lazy-bind fires the album-title wave); only the **non-library** trio / odd-credit case is uncovered. #628 followed its spec (`also_probe_album_title=False` default), so this is a deliberate omission to track, not a regression.

## Fix

Thread `also_probe_album_title=True` into the carry-through's `resolve_release_for_track` call (gating on album presence as the in-library path does), with a regression test for the non-library trio shape. Mind the cost: the album-title wave adds a Discogs probe — keep it inside the bounded (`max_validations=5`) budget and amortized by the #632 cache.

## Where

- `lookup/orchestrator.py` — `_resolve_nonlibrary_release` (the `resolve_release_for_track(...)` calls) + the TRACK_ON_COMPILATION carry-through call site.

## Related

- #628 (the carry-through), #625 (epic), #319 / #237 (the album-title-wave shape this misses)

jakebromberg:
Fixed in #655 — threads `also_probe_album_title=bool(album)` into `_resolve_nonlibrary_release`'s resolver call, with an integration regression for the non-library trio shape and a unit test pinning the gate.

Issue #633, opened 2026-06-20
A1-infra: bounded sequential-early-exit on resolve_release_for_track


jakebromberg:
Sub-issue of #625; infra dependency of #628 (the A1 carry-through). **Launchable independently** — adds an opt-in parameter; the existing caller (the #604 lazy bind) stays byte-for-byte unchanged. No dependency on #626.

## Problem

`resolve_release_for_track` ([lookup/release_resolution.py:188-293](https://github.com/WXYC/library-metadata-lookup/blob/9d9644dd22dbf2d2b5c86ecba9a30db6948864c0/lookup/release_resolution.py#L188-L293)) validates **every** merged candidate sequentially (`for release in candidates`, [L270-291](https://github.com/WXYC/library-metadata-lookup/blob/9d9644dd22dbf2d2b5c86ecba9a30db6948864c0/lookup/release_resolution.py#L270-L291)) with no cap, then ranks. For the #604 lazy-bind (library-row case) that's tolerable, but #628 points this primitive at **every non-library add**, where a popular track yields 20-40 candidates → 20-40 `validate_track_on_release` calls on a cold cache. That's the cold-cache cost the hardening project exists to prevent.

## Desired end state

An opt-in bounded mode: pre-rank candidates by title score, validate in rank order, **stop at the first that validates** (sequential-early-exit), with a backstop cap of **N=5** attempts. Existing callers keep validate-all behavior unless they opt in.

## Design

- Add a parameter to `resolve_release_for_track` (e.g. `max_validations: int | None = None` + `early_exit: bool = False`, or a small `bound` object). **Default = today's behavior** (validate-all, rank-after) so #604 callers are unchanged.
- Bounded mode: pre-rank `candidates` by `score_match(album, candidate.album)` — the signal `rank_resolved_releases` already applies *after* validation, moved *before* it — then validate in order, return the first that validates, stop after N=5 attempts.
- **No-album case** (`album` empty): no title to pre-rank on → order by the #629 rule (prefer `is_compilation=False`, then stable `release_id`), early-exit on first validation, cap 5.
- **Sequential, not concurrent:** minimizes Discogs quota (the binding constraint) and degrades gently under the fallthrough seam's rate-limit cool-down. (Parallelism, if ever needed, is revisited only for #629's no-album path.)

## Decision record — why sequential-early-exit

[...]

Issue #632, opened 2026-06-20
A1-infra: positive lml_cache.release_resolution_cache (durable repeat-add amortization)


jakebromberg:
Sub-issue of #625; infra dependency of #628 (the A1 carry-through). **Launchable independently** — purely additive infra, no behavior change, no dependency on #626. Models the existing `lml_cache.album_streaming_url_cache` (#573).

## Problem

A1 (#628) re-runs Discogs search + per-release track-validation for non-library releases on **every** flowsheet add, because no tier persists a positive `(artist, track/album) → release_id` resolution: `search` / `validate_track_on_release` are read-only (`pg_write=None`), the negative cache stores only misses (7-day TTL), and the in-memory L1 / `resolve_release_for_track_cached` tiers evaporate on restart. Without a durable positive cache, the search+validate cost #536 deferred is re-paid on every repeat add of the same non-library release — the cold-cache cost class that triggered the post-launch hardening project.

## Desired end state

A durable, LML-owned positive cache mapping a normalized `(artist, title, kind)` to a resolved Discogs `release_id`, so the 2nd-and-later add of the same non-library release short-circuits the probe+validate entirely.

## Design (clone `album_streaming_url_cache`)

LML-owned, **lifespan-bootstrapped (NOT alembic)** per the `lml_cache.*` ownership rule in CLAUDE.md:

```sql
CREATE TABLE IF NOT EXISTS lml_cache.release_resolution_cache (
    artist_normalized TEXT NOT NULL,
    title_normalized  TEXT NOT NULL,
    is_track          BOOLEAN NOT NULL,     -- true: (artist,track); false: (artist,album)
    release_id        INTEGER,              -- NULL = known miss
    resolved_at       TIMESTAMPTZ NOT NULL DEFAULT now(),
    PRIMARY KEY (artist_normalized, title_normalized, is_track),
    CONSTRAINT release_id_validity CHECK (release_id IS NULL OR release_id > 0)
);
```

- Keys normalized via `to_match_form()` from `wxyc_etl.text` (same as the streaming + negative caches). Key on the **typed** artist (consistent with the #626 two-channel decision).
- New module `entity/release_resolution_cache.py` mirroring `entity/streaming_url_cache.py`: `set_up_release_resolution_cache_schema(pg)`, a read (`get_cached_release_id`), a write (`set_cached_release_id`); best-effort (swallow + log on PG error).
[...]

Issue #627, opened 2026-06-20
Contract: document + parity-test the row-less result shape (release_id>0, album_id=null)


jakebromberg:
Sub-issue of #625. **Launchable now.** Re-scoped after the #625 grill confirmed the wire contract is **already satisfied** end-to-end — this is no longer a cross-repo change, just documentation + a guard test.

## Background — the shape is already legal and already rendered

Verified across all four repos at the current checkout:

- **wxyc-shared `api.yaml` (v1.14.0):** `LookupResultItem.artwork` (a `DiscogsMatchResult`, carrying `release_id`/`release_url`) is **optional and independent** of `library_item`. `LibraryCatalogItem` has no `album_id` — library identity is just `id`. A result with `library_item.id = 0` + a present `artwork.release_id > 0` is **already schema-valid**.
- **Backend-Service:** the Discogs link gates on `discogs_url` / `isSyntheticArtwork()` (`release_id === 0 && release_url === ''`), **never** on `album_id`. Free-form rows (`album_id IS NULL`) already carry `discogs_url`.
- **wxyc-ios-64:** the Discogs button renders on `metadata.discogsURL != nil`, independent of `album_id`; decode is `flatMap` (no force-unwrap).

So once #628 routes the resolved `release_id`/`release_url` through (instead of the `release_id=0` sentinel), BS derives `discogs_url` and iOS renders — **with no consumer change.**

## Problem this ticket closes

Both consumers render the shape correctly *today*, but **nothing pins it**: a future BS/iOS refactor of the `isSyntheticArtwork` / `discogsURL != nil` gate could silently regress the "`release_id > 0`, `album_id` null" case with no failing test.

## Desired end state

- `api.yaml` description text documents the `(library_item.id == 0, artwork.release_id > 0)` shape as a valid "Discogs identity, no catalog row" result — distinct from the `release_id == 0` BS#1185 sentinel.
- An LML-side parity/contract test asserts `perform_lookup` emits that shape for a row-less resolved release (and that it serializes against the generated models).
- Optional: thin follow-up issues filed into Backend-Service + wxyc-ios-64 to add their own gate assertions.

## Discriminator

`release_id > 0 && album_id == null` — unambiguous against the `release_id == 0` sentinel, backward-compatible with any consumer keying "has Discogs identity" on `release_id != 0`.

## TDD seed (write first)

[...]

PR #647, opened 2026-06-20
feat(#628): surface non-library resolvable releases via the row-less carry-through

### LINKED ISSUE #628: A1: surface non-library resolvable releases via the row-less carry-through (spine)
Sub-issue of #625 — the **A1 spine** (the carry-through). Blockers #626, #632, #633 are **merged**; this is now unblocked. Flag-gated rollout. **Re-grounded against `main` @ `a7a485a`** — the original body referenced prod `9d9644d`, but #622's search-strategy refactor plus the merged #626/#632/#633 moved every cut-site and added a third in-scope strategy. The line refs below are current on main.

## Problem

The track-validating Discogs-aware strategies find candidate Discogs releases, gate on a fuzzy library-row match **first**, and drop the release if none exists — track-validation is deferred until *after* the gate (#536), so a non-library release is dropped before the validation that would confirm it ever runs.

**#622 consolidated these into two cut-sites covering three strategies:**

1. **`process_release`** (TRACK_ON_COMPILATION) — [orchestrator.py:1721](https://github.com/WXYC/library-metadata-lookup/blob/main/lookup/orchestrator.py#L1721): `if not matches: return []` (plus a deferred-validation drop at 1735-1739).
2. **The shared kernel `_match_track_releases_to_library` → `_validate_one`** (orchestrator.py:1004-1149) — the double-gate at **1078** (`if not matches: return None`) and **1086** (`if not eligible: return None`). This kernel now backs **both `SONG_AS_TRACK`** (a thin wrapper) **and `SWAPPED_INTERPRETATION`** (`search_with_alternative_interpretation` → `_narrow_swapped_by_track` → the kernel).

So **`SWAPPED_INTERPRETATION` is a third violation site** the original pre-#622 analysis missed — it finds a Discogs release containing the track, then discards it for lack of a library row, identical shape. Fixing the kernel covers `SONG_AS_TRACK` **and** `SWAPPED` together.

The resolvable Discogs identity is lost; the #604 bind has nothing to bind to, and `enrich_artwork_results` (now at 2452) collapses the result to the BS#1185 `release_id=0` sentinel (synthesized at 3098).

(`SONG_AS_ARTIST` is **not** here — different caller, no track-validation; tracked in #631.)

## Desired end state

For a release **not** in `library.db` that resolves + validates from the inputs, `TRACK_ON_COMPILATION`, `SONG_AS_TRACK`, **and `SWAPPED_INTERPRETATION`** emit the row-less shape `_library_miss_discogs_search` already produces (`LibraryItem(id=0)` + real `DiscogsSearchResult`, orchestrator.py:1288-1293) — real `release_id`/`release_url`, not the sentinel. The #604 spine case (A Guy Called Gerald / "When There Is No Sun") binds end-to-end.

## Design (interfaces verified current on main)

- At each drop point (`process_release`@1721; kernel `_validate_one`@1078/1086), instead of dropping, resolve + validate the non-library candidate(s) and route into the row-less carry-through. Step-6 builds `LibraryCatalogItem(id=0, call_number="(external)", ...)` for id==0 items (orchestrator.py:3534-3544); Step-3b keeps id==0 out of re-validation (`real_results = [r for r in library_results if r.id != 0]`, orchestrator.py:3371).
- **Resolve** via `resolve_release_for_track(song, artist, album, discogs_service, *, also_probe_album_title=False, max_validations: int | None = None)` (release_resolution.py:215) in **#633's bounded mode** — pass `max_validations=5`. Do not add a second uncapped validate path.
- **⚠️ Cache-key fix (required — the #633 landmine):** the L1 wrapper `resolve_release_for_track_cached` (release_resolution.py:360) does **not** accept `max_validations`, and its `@async_cached` key is `(discogs_service, song, artist, album, also_probe_album_title)` — so a bounded `[]` (gave up at N=5) would **coalesce** with a default `[]` (validated all, found none). Two acceptable resolutions: **(a)** the non-library path calls the *uncached* `resolve_release_for_track(..., max_validations=5)` and relies on the #632 PG cache for durability (avoids the L1 wrapper entirely — cleanest), **or (b)** if you route bounded mode through the L1 wrapper, add `max_validations` to its signature so it joins the key. Either way, don't regress the existing #604 lazy-bind caller (orchestrator.py:2415), which calls the wrapper unbounded.
- **Positive cache (#632), keyed on the TYPED artist:** read `get_cached_release_id(pg, *, artist, title, is_track) -> ReleaseResolution(release_id, was_present)` first and short-circuit on `was_present`; on a cold resolve, `set_cached_release_id(pg, *, artist, title, is_track, release_id)`. **Key on `parsed.artist` (typed)** — NOT `library_artist_for(parsed)` (that's the #626 library channel) and NOT the canonical-swapped `probe_artist` the #604 lazy-bind happens to key on (orchestrator.py:2415 — a discipline that path does not follow; do not propagate it here). `ReleaseResolution` is three-valued: fresh hit / fresh known-miss (`was_present=True, release_id=None` → skip probe) / absent-or-stale (`was_present=False` → probe).
- **Flag:** add `lml_resolve_nonlibrary_release: bool = Field(default=False)` to `config/settings.py` (mirroring `lml_resolve_compilation_release` @ 167-184), read via `get_settings().lml_resolve_nonlibrary_release`. Env `LML_RESOLVE_NONLIBRARY_RELEASE`, default off, flipped after staging validation.

## Constraints

- **SWAPPED's `require_artist`:** the kernel passes `require_artist=artist` (a library-row re-filter). On the row-less path there's no library row to filter, and the artist is already settled upstream (the typed-artist `search_releases_by_track` probe + the `validate_track_on_release` track-credit check) — so the row-less carry-through should **bypass the `require_artist` library re-filter** rather than drop on it.
- Don't regress the #604 title-gate that strips loose fuzzy *library* mismatches — the row-less path bypasses it (no library row to mis-trust).
- In-library comp binding must not regress (staging `bound:True` cases still bind).
- Keep id==0 items out of Step-3b validation (orchestrator.py:3371).

## TDD seed (write first)

`tests/integration/test_lookup_nonlibrary_release.py` (fixture/mock Discogs): a non-library compilation track (the #604 repro — A Guy Called Gerald / "When There Is No Sun") with `LML_RESOLVE_NONLIBRARY_RELEASE` on → assert `library_item.id == 0`, `artwork.release_id > 0`, non-empty `discogs_url`; flag off → today's empty/sentinel behavior. **One case per strategy** (`TRACK_ON_COMPILATION`, `SONG_AS_TRACK`, `SWAPPED_INTERPRETATION`). Plus a unit test asserting the second identical lookup hits the #632 cache (no second resolve), and — if you take resolution path (b) above — a unit test pinning that a bounded `[]` and a default `[]` do NOT share a cache entry.

## Acceptance criteria

- [ ] `/lookup` for a non-library release resolvable on Discogs returns a non-empty `discogs_url` via `TRACK_ON_COMPILATION`, `SONG_AS_TRACK`, **and `SWAPPED_INTERPRETATION`** (row-less carry-through, not the sentinel), behind `LML_RESOLVE_NONLIBRARY_RELEASE`.
- [ ] #604 spine case binds end-to-end.
- [ ] Resolution goes through #633's bound (`max_validations=5`) and is amortized by #632's cache (keyed on the typed artist); a repeat add does not re-pay search+validate.
- [ ] No bounded-vs-default `[]` cache coalescing (the #633 landmine is closed by path (a) or (b)).
- [ ] In-library comp binding + the #604 title-gate do not regress; flag-off path = today's behavior.
- [ ] Unit + integration coverage incl. one case per strategy and the cache-key separation.

## Related

- #625 (parent), #626 / #632 / #633 (merged prerequisites), #622 (the kernel consolidation this re-grounds against), #604 + #616 (the bind / PR2), #536 (deferred validation), #401 / BS#1185 (sentinel)

jakebromberg:
## What

The spine of #625 (A1): when every library-membership wave drops its candidates on the library gate, a track may still resolve and validate against a **non-library** Discogs release (rotation, a DJ's personal collection, an unfiled reissue). Instead of returning empty, surface that release **row-less** — a `LibraryItem(id=0)` carrier plus the real `DiscogsSearchResult` (real `release_id`/`release_url`) — so Backend-Service derives `discogs_url` and iOS renders the Discogs button, with no consumer change.

This is the `release_id > 0 && album_id == null` shape, distinct from the BS#1185 `release_id == 0` sentinel. Its wire contract is pinned by the parity test already on main (`tests/unit/test_rowless_result_contract.py`, #627) and documented by the paired #627 doc PRs.

## How

A shared helper `_resolve_nonlibrary_release(...)` is dropped at the three library-gate cut-sites:

- **TRACK_ON_COMPILATION** — `process_release` in `lookup/orchestrator.py`, keyed on the typed `parsed.artist`.
- **SONG_AS_TRACK** and **SWAPPED_INTERPRETATION** — the shared kernel `_match_track_releases_to_library._validate_one` (the #622 consolidation made SWAPPED a third site; SWAPPED anchors on the surfaced release's own credit and keeps per-track validation).

The helper reads the #632 positive-resolution cache (`get_cached_release_id`), short-circuits on a fresh hit/known-miss, and otherwise resolves through the **uncached, bounded** resolver (`resolve_release_for_track(..., max_validations=5)` from #633) — never the L1 `resolve_release_for_track_cached` wrapper (whose key omits `max_validations`, the #633 landmine) — then writes the resolution back.

**Flag-gated, off by default** (`lml_resolve_nonlibrary_release: bool = False`, mirroring `lml_resolve_compilation_release`). Safe to merge — no production behavior change until the flag flips.

## Review fixes folded in

- **Data safety** — a known-good positive cache entry that fails to rehydrate during a transient `get_release` outage is **no longer demoted** to a 7-day miss when the live re-resolve (same outage) also comes up empty. It survives and self-heals (`cached_positive_unhydrated` guard; TDD-pinned by `test_transient_rehydrate_failure_does_not_demote_positive_to_miss`).
- **Test robustness** — per-strategy tests now pin routing via `search_type` (SWAPPED → `alternative`; the others → not), so a priority change can't silently re-route them.
- **Typecheck** — renamed the carry-through local off `resolved` (collided with #604's non-optional `ResolvedRelease` binding in the same `process_release` scope) to clear the `mypy . --ignore-missing-imports` gate.

## Verification

`ruff check` + `ruff format --check` clean; `mypy . --ignore-missing-imports` clean (348 files); unit suite **2902 passed**; carry-through unit + integration **18 passed**. Rebased onto current `main` (atop #573's Bandcamp landing, which shares `settings.py`/`orchestrator.py`).

## Follow-up (out of scope, tracked)

#646 — the carry-through doesn't fire the album-title-probe wave, so a **non-library** trio/odd-credit release is uncovered (in-library is fine via #604). In-spec-by-omission for this ticket.

Closes #628

## Related

- #625 (epic), #627 (contract this shape satisfies), #632 (positive cache), #633 (bounded resolver), #646 (album-title-arm follow-up), BS#1185 (the sentinel this sits beside)


jakebromberg:
## Adversarial review loop — converged (2 rounds, 5 reviewers)

**Round 1** (3 reviewers, distinct lenses):
- **BLOCKER (fixed, `57c15c5`)** — `enrich_artwork_results`'s LML#487 title gate clobbered the carry-through's validated row-less binding down to the BS#1185 `release_id=0` sentinel whenever the request `album` didn't fuzzy-match the resolved release title. Empirically reproduced (album-bearing TRACK_ON_COMPILATION → `release_id=0`). Fixed by exempting the row-less item (no library row ⟹ no LML#487 sibling-leak risk) from the gate. Regression test + a #632 cache-threading integration test added; routing assertions tightened to `== "compilation"`.
- **SUBSTANTIAL (filed #649)** — SONG_AS_TRACK anchors the #632 cache key + validation on the release-level "Various" credit for V/A comps (needs track-credit plumbing; flag-off; narrow path).
- NIT — api.yaml provenance under-credited the carry-through as a producer → fixed in WXYC/wxyc-shared#185.

**Round 2** (2 reviewers, fix-verification + fresh-eyes): confirmed the fix **correct and complete** (verified by surgical revert; the #583 library-miss path is a strict no-op), **no new BLOCKER/SUBSTANTIAL**. Two nits applied (`ccfe7d7`): empty-title rehydrate hardening, and an `extended=true` parametrization (BS forces extended on the write path).

Local CI green: full suite **3155 passed**, ruff + mypy clean. Follow-ups: #646, #649.

Issue #625, opened 2026-06-19
/lookup is anchored to library-row membership — non-library releases (rotation, personal collection) lose resolution even when LML has the metadata


jakebromberg:
## Principle

LML's job is to resolve metadata for **any** release the inputs identify — whether the record is in rotation, the permanent library, or a DJ's personal collection. Library membership confirms WXYC owns a copy and supplies call numbers / filing / curated streaming links; it should **enrich** a result, never **authorize** it. Today the `/lookup` pipeline inverts that: it is *rooted* in `library.db`, and Discogs is used only to cross-reference, validate, or enrich rows that already matched a library row. A release absent from `library.db` is dropped even when LML has already resolved and validated it.

> This was originally framed as "rotation comps are invisible to `library.db`." That's a *symptom*, not the bug. Sourcing `ROTATION_RELEASE` into `library.db` would paper over the rotation case while doubling down on catalog coupling — and still wouldn't help a personal-collection or one-off release. The fix is to decouple resolution from catalog membership.

## Concrete prod repro

Live flowsheet entry, enriched **after** the prod flag flip (22:34Z), so `LML_RESOLVE_COMPILATION_RELEASE` was ON:

| field | value |
|---|---|
| artist / track / album | **Chez Damier & Ben Vedren / "The Three Dimensions of Air" / "When There Is No Sun"** |
| `add_time` | `2026-06-19T23:05:55.931Z` |
| `rotation_id` | 43171 (rotation track, not in `library.db`) |
| `metadata_status` | `enriched_match` |
| `discogs_url` | `''` |
| `album_id` / `artwork_url` | `None` / `None` |

iOS "More Info" renders cover art + streaming **search** buttons but **no Discogs link** — the #604 symptom, on the same comp ("When There Is No Sun") as #604's own spine case (A Guy Called Gerald). So #604's fix doesn't surface a Discogs link for the case it was written for — not because of "rotation", but because that release has no library row.

[...]

Issue #604, opened 2026-06-19
Bind discogs_url for Various-Artists compilation tracks (release resolution)


jakebromberg:
## Problem

A flowsheet track that lives on a Various-Artists (V/A) compilation never gets a Discogs link in the app, even when LML can prove exactly which Discogs release the track sits on. Concrete case: **A Guy Called Gerald — "Message to Black Youth"**, on the compilation **"When There Is No Sun"** ([release/36907527](https://www.discogs.com/release/36907527), credited "Various"). The iOS "More Info" section renders its Discogs button only when `album.discogsURL` is non-nil; that value traces to `album_metadata.discogs_url` in Backend, which LML never bound.

Streaming buttons and cover art *do* appear (they resolve through compilation-agnostic paths). Only the structured Discogs link is missing — systematically, for **every** compilation track.

## Root cause

Two facts combine:

1. **No producer fires for a found album row.** Every LML path that resolves a specific V/A `release_id` by track (`TRACK_ON_COMPILATION`, `SONG_AS_TRACK`, the library-miss probe) is gated on `song_not_found` / library-miss. The enrichment worker supplies the album, so the pipeline short-circuits before the compilation strategy runs.
2. **The binding step re-derives the release through an artist floor that rejects "Various".** `fetch_artwork_for_items.fetch_one` re-searches Discogs by the track artist and scores candidates at the 80/80 artist floor; "A Guy Called Gerald" can never clear 80 against a release credited "Various" → no `DiscogsMatchResult` → no `discogs_url`.

**Why a fix is possible:** `validate_track_on_release` matches the *per-track* artist credit, not the release-level credit. "Message to Black Youth" credits "A Guy Called Gerald" on its own tracklist row, so the comp validates `True` for the track artist. Routing the binding step through validation instead of the floor is the bypass.

## Approach

Extract one deep release-resolution module (`lookup/release_resolution.py`) that owns "find and validate the Discogs release a track sits on" — the probe set + per-candidate track-credit validation + ranking (title-match to album, stable `release_id` tie-break). The binding step calls it lazily when its floor search fails; the miss-path strategies delegate to its shared primitives.

Full design, decisions table, and test matrices: [`docs/plans/compilation-release-resolution.md`](https://github.com/WXYC/library-metadata-lookup/blob/main/docs/plans/compilation-release-resolution.md) (lands with PR1).

## PR sequencing

[...]

Issue #319, opened 2026-05-13
Album-title text search fallback for self-named / trio compilations


jakebromberg:
## Problem

Carved out of #237. The trio-collaboration failure shape from #237 cannot be fixed by an artist resolver alone — there's no single canonical Discogs artist entity for "Orcutt Shelley Miller" (each trio member is credited separately), so no fuzzy artist match exists at any score floor. The release **is** in `discogs-cache` (release 34993109), it has 14+ flowsheet plays since 2026-01-12, and a direct Discogs page search locates it instantly — but the two parallel probes in `search_compilations_for_track` ([`lookup/orchestrator.py:464-469`](https://github.com/WXYC/library-metadata-lookup/blob/main/lookup/orchestrator.py#L464-L469)) both filter or scope by artist, and neither lights up.

The validation side from PR #236 already accepts trio-style credits via `rapidfuzz.token_set_ratio` ([`discogs/service.py:1028-1114`](https://github.com/WXYC/library-metadata-lookup/blob/main/discogs/service.py#L1028-L1114)). All that's missing is a **candidate-generation path that doesn't require artist filtering** — search Discogs by `release_title` alone (with `format=Compilation`), then let the now-fuzzy `validate_track_on_release` filter the candidates back down.

## Desired end state

When the existing parallel gather returns no usable releases AND the request includes an album title AND the sibling resolver ticket's pre-pass did not produce a high-confidence canonical swap (so the existing probes already saw the best-canonical artist form), `search_compilations_for_track` fires one additional Discogs API call: a title-only compilation search. Candidates are routed through the existing `process_release` post-process loop and validated by the existing fuzzy `validate_track_on_release`.

For the trio repro from #237: `POST /api/v1/lookup` with `{artist: "Orcutt Shelley Miller", album: "Orcutt Shelley Miller", song: "A Star Is Born"}` returns release 34993109 in `results`.

## Where

- **New service method**: [`discogs/service.py`](https://github.com/WXYC/library-metadata-lookup/blob/main/discogs/service.py) — `search_compilations_by_title(album: str, limit: int = 10) -> TrackReleasesResponse` (or a new `ReleaseSearchResponse` if more natural; reuse `ReleaseInfo` for items). API-only — no cache leg. See "Why API-only" below. Builds Discogs API params:
  ```python
  {"type": "release", "release_title": album, "format": "Compilation", "per_page": limit}
  ```
[...]

PR #655, opened 2026-06-21
fix(#646): fire album-title wave in the row-less carry-through resolve

### LINKED ISSUE #646: Row-less carry-through misses the album-title-probe arm (non-library trio / odd-credit case)

jakebromberg:
Closes #646.

Follow-up from the #628 review (finding A — in-spec-by-omission, not a #628 bug).

## Problem

The #628 row-less carry-through (`_resolve_nonlibrary_release`) resolved a non-library release via `resolve_release_for_track(..., max_validations=5)` **without** `also_probe_album_title=True` — so it probed only the artist-scoped track waves (Wave A/B). A release discoverable *only* through the album-title wave — the #319/#237 trio-collaboration / odd-credit shape, where Discogs files the per-track credit unusually (e.g. typed "Orcutt Shelley Miller") — was dropped at `process_release`'s library gate **and** missed by the carry-through's track-only resolve.

In-library trio releases were already covered: the #604 lazy-bind fires the album-title wave on `bool(album)` (orchestrator.py `_resolve_fallback_artwork`'s resolve call). Only the **non-library** case was uncovered — a deliberate omission of #628's spec (`also_probe_album_title=False` default), not a regression.

## Fix

Thread `also_probe_album_title=bool(album)` into the single `resolve_release_for_track` call inside `_resolve_nonlibrary_release`. All three Discogs-aware strategies (`TRACK_ON_COMPILATION`, `SONG_AS_TRACK`, `SWAPPED_INTERPRETATION`) inherit it through the shared helper — both call sites already thread `album` in and delegate the resolver call.

Unlike the in-library #604 caller, this helper keys on the **typed** artist and applies no canonical swap, so `bool(album)` is the whole gate (no `not swapped` clause needed — there's no swap to make the artist-scoped probe authoritative).

### Cost

- One extra Discogs search on a **cold miss only** — the album-title probe joins the existing concurrent `asyncio.gather`.
- Stays inside the `max_validations=5` budget: the album-title candidates merge into the same pre-ranked pool and compete for the existing 5 validation slots, rather than adding validations.
- Amortized by the #632 cache, which short-circuits the resolve entirely on a warm typed-artist key (the cache read precedes the resolver call).
- Scope: the extra album-title wave fires for **every** non-library add with an album typed, across all three strategies (`SONG_AS_TRACK` / `SWAPPED_INTERPRETATION` / `TRACK_ON_COMPILATION`). Song-only adds (no album typed) are unaffected — the `bool(album)` gate is what spares them.

## Tests

- **Integration** (`test_lookup_nonlibrary_release.py`): a non-library release where the track-artist probe returns nothing and the release is reachable only by album title → row-less surfaces with a real `release_id`/`release_url`. Confirmed RED before the fix (`0 == 1`, no row-less item), green after.
- **Unit** (`test_nonlibrary_release_resolution.py`): parametrized over present / `None` / empty album, pinning that the carry-through passes `also_probe_album_title=bool(album)` to the resolver — the isolated discriminator, immune to integration-routing churn.

## Verification

- Full suite: 3170 passed, 1 skipped, 2 xfailed (`-m "not pg and not external_api and not stack and not slow"`).
- `ruff format` + `ruff check`: clean. `mypy . --ignore-missing-imports`: clean (348 files).

## Related

- #628 (the carry-through this extends), #625 (epic), #319 / #237 (the album-title-wave shape this misses), #604 (the in-library lazy-bind at parity), #633 (the `max_validations` bound), #632 (the positive cache that amortizes it).

### DIFF
--- a/lookup/orchestrator.py
+++ b/lookup/orchestrator.py
@@ -1071,18 +1071,23 @@ async def _resolve_nonlibrary_release(
        fresh known miss (``was_present=True, release_id=None``) short-circuits to
        ``None`` *without* a live probe; an absent/stale entry falls through.
     2. **Bounded resolve on a cold miss** — the **uncached**
-       ``resolve_release_for_track(..., max_validations=5)`` (#633). Deliberately
-       NOT ``resolve_release_for_track_cached``: the L1 wrapper's key omits
-       ``max_validations``, so a bounded ``[]`` would coalesce with a default
-       ``[]`` (the #633 landmine). Cold-cache durability comes from the #632 PG
-       cache here instead.
+       ``resolve_release_for_track(..., also_probe_album_title=bool(album),
+       max_validations=5)`` (#633). The album-title wave (#646, gated on a
+       present ``album``) fires the third ``search_releases_by_album_title``
+       probe so a non-library release discoverable *only* by its album title
+       — the #319/#237 trio-collaboration / odd-credit shape the track-artist
+       probes miss — still resolves, at parity with the in-library #604
+       lazy-bind. Deliberately NOT ``resolve_release_for_track_cached``: the L1
+       wrapper's key omits ``max_validations``, so a bounded ``[]`` would
+       coalesce with a default ``[]`` (the #633 landmine). Cold-cache
+       durability comes from the #632 PG cache here instead.
     3. **#632 cache write-back** — the resolved id, or ``None`` to record a known
        miss so the next identical add short-circuits.
 
     ``pg`` is best-effort: ``None`` (or a PG failure, swallowed inside the cache
     helpers) degrades to an uncached bounded resolve. ``album`` (often absent on
-    the kernel paths) only ranks the validated candidates by title; the bounded
-    resolver returns at most one.
+    the kernel paths) both fires the album-title probe (layer 2) and ranks the
+    validated candidates by title; the bounded resolver returns at most one.
     """
     if discogs_service is None or not song or not artist:
         return None
@@ -1106,7 +1111,12 @@ async def _resolve_nonlibrary_release(
             cached_positive_unhydrated = True
 
     candidates = await resolve_release_for_track(
-        song, artist, album, discogs_service, max_validations=5
+        song,
+        artist,
+        album,
+        discogs_service,
+        also_probe_album_title=bool(album),
+        max_validations=5,
     )
     best = candidates[0] if candidates else None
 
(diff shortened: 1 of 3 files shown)


jakebromberg:

Focused review against one risk: a resolved row-less item reaching `enrich_one` with `release_id <= 0` or not top-1 would silently miss the #628 `library_row_acceptable` bypass and collapse to the BS#1185 `release_id=0` sentinel.

**Verdict: SAFE.** The `also_probe_album_title=bool(album)` change makes the trio/odd-credit case resolve to a `release_id > 0` that binds onto the row-less artwork via `_bind_resolved_release`; the item is the sole result (top-1), and `test_track_on_compilation_surfaces_rowless_via_album_title_wave` pins `artwork.release_id == SPINE_RELEASE_ID (> 0)` end-to-end. The precision floor still holds — an album-wave candidate that fails per-track `validate_track_on_release` correctly returns empty → sentinel (parity with the in-library #604 path), which is why the test mocks validation true. The `bool(album)` gate is belt-and-suspenders with the resolver's own `fire_album = also_probe_album_title and bool(album)`, so song-only queries don't pay the extra wave.

Worth a line in the description: because the change lands in the shared `_resolve_nonlibrary_release`, all of SONG_AS_TRACK / SWAPPED / TRACK_ON_COMPILATION now fire one extra `search_releases_by_album_title` call per non-library add **that has an album typed** (song-only adds unaffected). It's bounded by `max_validations=5` and amortized by the #632 PG cache, so not a hot-path amplifier — just a per-add quota note for album-present non-library adds.

PR #656, opened 2026-06-21
feat(#631): row-less SONG_AS_ARTIST for non-library listener requests

### LINKED ISSUE #631: SONG_AS_ARTIST: row-less Discogs resolution for non-library listener requests (request-o-matic)
Sub-issue of #625 — a **sibling** to the A1/A2 flowsheet-resolution spine, deliberately separated rather than folded into #629. SONG_AS_ARTIST serves a different caller (request-o-matic listener requests) with a different value and risk profile, and shares none of the spine's track-validation machinery.

> **Re-grounded against `main` @ `dbe19cf`** (original body referenced prod `9d9644d`; line refs below are current). Two corrections since the original framing: (1) the #628 relationship is sharpened from "reuses the carry-through" to **precondition** — #628's row-less machinery only *enriches* once a real `release_id` is bound, and SONG_AS_ARTIST had **no `release_id` source**, so #631's load-bearing deliverable is a new artist→release resolver; #628's enrichment then comes for free. (2) The "open design question" (which release represents an artist request) is **decided** — see *Release selection* below. Implemented in **PR #656** (in review).

## Problem

`search_song_as_artist` ([orchestrator.py:961-1023](https://github.com/WXYC/library-metadata-lookup/blob/dbe19cf/lookup/orchestrator.py#L961-L1023)) reinterprets the typed *song* token as an *artist* name, looks up that artist's releases on Discogs, and keeps only those that cross-reference a WXYC library row. A non-library artist is dropped even when Discogs cleanly resolves the name — and, crucially, the old `lookup_releases_by_artist` returned only `(artist, album)` **name strings** ([discogs/lookup.py:103](https://github.com/WXYC/library-metadata-lookup/blob/dbe19cf/discogs/lookup.py#L103)), so there was **no `release_id` anywhere on this path** to carry a resolved Discogs identity.

This strategy primarily serves **request-o-matic**: a listener requests an artist, rom calls LML `/lookup` and posts the `(library_item, artwork)` result to Slack (`request-o-matic/routers/request.py:308-341`). So when a listener requests an artist WXYC doesn't own, rom posts "not found" instead of "here's the Discogs release they mean (not in our library)" — losing useful context for the DJ reading the channel.

## Relationship to #628 (precondition, not free pre-delivery)

The #628 row-less carry-through is two halves: an **enrich/bind tail** (the `discogs_titles[ROWLESS_LIBRARY_ID]` seam → `_make_rowless_item` → `_bind_resolved_release` → the `library_row_acceptable` bypass in `enrich_one` that surfaces the full extended payload + bio/wiki) and a **resolver** (`_resolve_nonlibrary_release` → `resolve_release_for_track`). The tail is **reusable as-is**. The resolver is **not**: it is track-keyed (probes `search_releases_by_track` + validates per-track credit), and SONG_AS_ARTIST has no track — the typed token *is* the artist — so it returns nothing here. (This is exactly why #625's A1 spine *excluded* SONG_AS_ARTIST.)

Consequence: #628's bypass only fires once `artwork.release_id > 0`, and nothing on this path produced a `release_id`. So **#631 must originate the release_id itself** via an artist-only resolver; only then does #628's enrichment land for free. The genuine new work is the resolver + selection + confidence gate; emission/enrichment is inherited.

## Why this is separate from #629 (the spine)

- **Different caller / surface:** rom → Slack (human-read), not the flowsheet-add → iOS metadata cache the spine serves.
- **Different risk tolerance:** a wrong Slack post is discounted by a human; the spine writes a machine-read artwork cache, so its bar is tighter. This path's bar is looser.
- **No shared resolver:** SONG_AS_ARTIST does artist-search, not track-validation — the A1 cold-path bound (#633) and the #632 positive resolution-cache key shape (track-keyed) don't apply. It needs its own confidence model.

## Desired end state

For a listener request whose artist resolves on Discogs but isn't in `library.db`, SONG_AS_ARTIST surfaces a row-less result (`LibraryItem(id=0)` + the resolved release on the `discogs_titles` seam) so rom can post the Discogs context to Slack — gated so a coincidental name match doesn't surface a stranger's release, and flag-gated on `LML_RESOLVE_NONLIBRARY_RELEASE` (shared with #628).

## Confidence gate

Surface only when the typed token **normalizes-equal** to the credited Discogs artist name: `normalize_for_comparison(token) == normalize_for_comparison(release.artist)` (the artist-only analog of `find_best_typed_match`'s floor). "Discogs returned some releases" is **not** sufficient. V/A compilations credited to "Various" fail the gate by construction, so survivors are the artist's own releases.

## Release selection (decided)

Among the releases that clear the normalize-equal gate, pick the **most-relevant** by Discogs `confidence` (descending), with a stable `release_id` tiebreak for determinism — `_select_rowless_artist_release` ([orchestrator.py, PR #656](https://github.com/WXYC/library-metadata-lookup/pull/656)). This is the artist-only "best representative" rule that was the original open product question; Discogs's own relevance ranking is the signal, the normalize-equal gate is the floor, and the tiebreak makes the choice reproducible across requests. To enable this, `lookup_releases_by_artist` now returns the full ranked `DiscogsSearchResult` list (carrying `release_id`/`release_url`) instead of `(artist, album)` strings.

## Where

- `lookup/orchestrator.py` `search_song_as_artist` (the library-filter drop) + new `_select_rowless_artist_release`
- `discogs/lookup.py` `lookup_releases_by_artist` — widened return to carry `release_id`/`release_url`
- `lookup/strategies/song_as_artist.py` — forward the `discogs_titles` seam instead of discarding the tuple's second element
- reuses the #628 enrich/bind tail (`_make_rowless_item`, the seam, `_bind_resolved_release`) for emission
- `normalize_for_comparison` for the gate

## Constraints

- Depends on #628 (the row-less `id=0 + release_id` emission tail). #628 is **merged**.
- Don't regress library-backed SONG_AS_ARTIST (in-library requests still resolve as today; second tuple element stays `None` on those paths).
- Confirm with the request-o-matic owner that row-less request results are wanted in Slack (vs an explicit "not in library" note).

## Acceptance criteria

- [ ] A rom-path request for a non-library artist that resolves on Discogs surfaces a row-less result, gated on typed-token ⟂ Discogs-artist-name normalized equality.
- [ ] A coincidental token→artist-name match (not normalized-equal) does **not** surface a release.
- [ ] Library-backed SONG_AS_ARTIST behavior unchanged.
- [x] The "which release represents an artist request" rule is decided + documented (most-relevant by Discogs `confidence`, `release_id` tiebreak — see *Release selection*).
- [ ] Unit coverage for the gate (equality pass/fail) and the release-selection rule; integration coverage that the row-less result carries `release_id > 0` (not the BS#1185 sentinel).

## Related

- #625 (parent), #628 (the enrich/bind tail this reuses; resolver it must *not* reuse), #629 (the spine sibling)
- request-o-matic (the consuming pipeline)

jakebromberg:
Closes #631.

## Problem

`search_song_as_artist` reinterprets a request-o-matic listener's typed token as an artist name, looks up that artist's releases on Discogs, and keeps only those that cross-reference a WXYC library row. A non-library artist was dropped even when Discogs cleanly resolved the name — so rom posted "not found" to Slack instead of "here's the Discogs release they mean (not in our library)", losing useful context for the DJ reading the channel.

## Change

When no library row matches and the typed token **normalize-equals** a resolved Discogs artist name, emit the row-less `LibraryItem(id=0)` + `ResolvedRelease` via the #628 `discogs_titles` carry-through (real `release_id`/`release_url`, not the BS#1185 sentinel), gated behind the shared `LML_RESOLVE_NONLIBRARY_RELEASE` flag (default **off**, ships dark).

### Confidence gate + selection rule (the #631 open design question, decided)

`to_match_form(token) == to_match_form(discogs_artist_name)` is the artist-only analog of `find_best_typed_match`'s floor: a coincidental token-to-name hit is dropped, and V/A compilations credited to "Various" fail it by construction — so the survivors are the artist's **own** releases. Among them the most-relevant (Discogs `confidence`, stable `release_id` tiebreak) represents the request. True community-count ranking would need a per-release `get_release` fetch (the cost #633 bounds) and is deferred; the gate already excludes comps and the pick is deterministic.

### Mechanics

- Widen `lookup_releases_by_artist` to return full `DiscogsSearchResult`s, so one fetch serves both the library cross-reference and the row-less identity.
- Thread the `discogs_titles` seam through `SongAsArtist.attempt` and a new optional `Outcome.found(discogs_titles=)`, so `fetch_artwork_for_items` binds the carried release onto the `id=0` item. The binding side already gates any `id=0` item on `LML_RESOLVE_NONLIBRARY_RELEASE`, so reusing that flag (vs. a new one) keeps the change minimal.
- Library-backed SONG_AS_ARTIST is unchanged.

## Tests

- Unit (`test_orchestrator_gaps.py`): gate pass/fail, flag-off, library-backed-unchanged, selection (comp-exclusion + confidence ordering).
- Contract (`test_discogs_lookup.py`): widened `lookup_releases_by_artist` return shape.
- Integration (`test_lookup_song_as_artist_rowless.py`): end-to-end `perform_lookup` — `search_type == "song_as_artist"`, `artwork.release_id > 0`, non-empty `discogs_url`, behind the flag; flag-off asserts today's behavior.

Full suite: 3185 passed (pg/external_api deselected). Lint + format clean.

## Acceptance criteria (#631)

- [x] Non-library artist that resolves on Discogs surfaces a row-less result, gated on typed-token ⟂ Discogs-artist-name normalized equality.
- [x] A coincidental (not normalize-equal) match does **not** surface a release.
- [x] Library-backed SONG_AS_ARTIST behavior unchanged.
- [x] Release-selection rule decided + documented before implementation (own-release-not-comps; documented in `_select_rowless_artist_release`).
- [x] Unit coverage for the gate and the selection rule.

## Related

- #625 (parent), #628 (the row-less carry-through this reuses for emission).
- request-o-matic gets a sibling fix so a row-less `id=0` result doesn't post a malformed `<|WXYC>` Slack link (empty `library_url`); filed alongside.

### DIFF
--- a/core/search.py
+++ b/core/search.py
@@ -454,7 +454,12 @@ class Outcome:
         return cls(items=[])
 
     @classmethod
-    def found(cls, items: list[LibraryItem]) -> "Outcome":
+    def found(
+        cls,
+        items: list[LibraryItem],
+        *,
+        discogs_titles: "dict[int, ResolvedRelease] | None" = None,
+    ) -> "Outcome":
         """Direct match where the strategy can vouch for the song.
 
         Clears ``state.song_not_found`` because the strategy explicitly
@@ -463,13 +468,19 @@ class Outcome:
         etc.). Use this for the 90% case where success means "the song
         was matched."
 
+        ``discogs_titles`` carries resolved Discogs releases for artwork lookup
+        (defaults to ``None`` — a no-op). SONG_AS_ARTIST's LML#631 row-less path
+        passes ``{0: ResolvedRelease}`` here so ``fetch_artwork_for_items`` binds
+        the non-library release onto the ``id=0`` item instead of falling
+        through to the BS#1185 sentinel.
+
         For ARTIST_PLUS_ALBUM's direct path — where the artist's album was
         found by name but the song's presence on the album wasn't verified —
         use :meth:`album_match` instead, which leaves ``song_not_found``
         alone so a prior album-resolution signal can still drive
         TRACK_ON_COMPILATION downstream.
         """
-        return cls(items=items, song_not_found_after=False)
+        return cls(items=items, song_not_found_after=False, discogs_titles=discogs_titles)
 
     @classmethod
     def album_match(cls, items: list[LibraryItem]) -> "Outcome":
--- a/discogs/lookup.py
+++ b/discogs/lookup.py
@@ -7,7 +7,7 @@ When omitted, a cacheless service is created as a fallback.
 
 import logging
 
-from discogs.models import DiscogsSearchRequest
+from discogs.models import DiscogsSearchRequest, DiscogsSearchResult
 from discogs.service import DiscogsService
 
 logger = logging.getLogger(__name__)
@@ -80,7 +80,7 @@ async def lookup_releases_by_artist(
     artist: str,
     limit: int = 10,
     service: DiscogsService | None = None,
-) -> list[tuple[str, str]]:
+) -> list[DiscogsSearchResult]:
     """Look up releases by an artist using Discogs.
 
     Args:
@@ -90,7 +90,10 @@ async def lookup_releases_by_artist(
             creates a cacheless fallback service.
 
     Returns:
-        List of (artist, album) tuples for releases by or featuring the artist.
+        The ranked ``DiscogsSearchResult`` list for releases by or featuring the
+        artist. Callers that only need album titles read ``r.album``; the LML#631
+        row-less path also reads ``r.release_id`` / ``r.release_url`` to carry the
+        resolved identity for a non-library artist.
     """
     if service is None:
         service = _get_service()
@@ -100,4 +103,4 @@ async def lookup_releases_by_artist(
     request = DiscogsSearchRequest(artist=artist)
     response = await service.search(request, limit=limit)
 
-    return [(r.artist or "", r.album or "") for r in response.results]
+    return list(response.results)
--- a/lookup/orchestrator.py
+++ b/lookup/orchestrator.py
@@ -1078,12 +1078,15 @@ def _select_rowless_artist_release(
     if not own:
         return None
     best = min(own, key=lambda r: (-r.confidence, r.release_id))
-    item = _make_rowless_item(artist=best.artist or token, title=best.album or "")
+    # One album-title fallback feeds both the synthetic row's title and the
+    # carried release's album_title, so they can't drift.
+    album_title = best.album or ""
+    item = _make_rowless_item(artist=best.artist or token, title=album_title)
     resolved = ResolvedRelease(
         release_id=best.release_id,
         release_url=best.release_url,
         is_compilation=False,
-        album_title=best.album or "",
+        album_title=album_title,
     )
     return item, resolved
 
--- a/lookup/orchestrator.py
+++ b/lookup/orchestrator.py
@@ -1046,12 +1046,22 @@ def _select_rowless_artist_release(
 ) -> tuple[LibraryItem, ResolvedRelease] | None:
     """Pick the release that represents a non-library artist request (LML#631).
 
-    Confidence gate: keep only releases whose credited artist normalizes-equal
-    to the typed token (the artist-only analog of ``find_best_typed_match``'s
-    floor). A coincidental token→name hit is dropped. V/A compilations credited
-    to "Various" fail the gate by construction, so the survivors are the artist's
-    own releases. Among them, pick the most-relevant (Discogs ``confidence``
-    descending) with a stable ``release_id`` tiebreak for determinism.
+    Credit gate: keep only releases whose credited artist normalizes-equal to
+    the typed token (the artist-only analog of ``find_best_typed_match``'s
+    floor). A coincidental token→name hit is dropped, as is any release whose
+    credit differs from the token — V/A compilations credited to "Various" and
+    collaborations credited "<artist> & <other>" among them — so the survivors
+    are releases credited to the artist alone.
+
+    Among the survivors, sort by Discogs ``confidence`` descending with a stable
+    ``release_id`` tiebreak. Note the upstream query is artist-only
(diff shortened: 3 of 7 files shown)


jakebromberg:

Focused review against one risk: a surfaced row-less item reaching `enrich_one` with `release_id <= 0` or not top-1 would silently miss the #628 `library_row_acceptable` bypass and collapse to the BS#1185 `release_id=0` sentinel — the not-found outcome this ticket exists to kill.

**Verdict: SAFE.** This is the precondition ticket (#628's track-keyed `_resolve_nonlibrary_release` cannot resolve an artist-only query), and the PR handles it correctly: `_select_rowless_artist_release` genuinely **originates** a real `release_id` from the artist-only Discogs search (`lookup_releases_by_artist` widened to return `DiscogsSearchResult`s), threads `{ROWLESS_LIBRARY_ID: resolved}` onto the seam, and `_bind_resolved_release` emits `artwork.release_id > 0`. The item is structurally top-1 — single-element `[item]`, `Outcome.found` halts the pipeline, Step-3b excludes id==0. `test_song_as_artist_surfaces_rowless_when_flag_on` pins `artwork.release_id > 0` end-to-end through the real `enrich_one` path; the normalize-equal gate, V/A exclusion, flag-off, and library-backed paths are all covered.

**One defense-in-depth suggestion (non-blocking):** `_select_rowless_artist_release` uses `best.release_id` in the tiebreak and the `ResolvedRelease` without an explicit `> 0` guard — it relies on the upstream Discogs invariant that release ids are positive. That holds today, but a malformed/`<= 0` id surviving upstream would silently fail the `artwork.release_id > 0` check in the bypass and collapse to the sentinel (the silent no-op). A one-liner closes the only theoretical path: filter `own` on `r.release_id > 0` (or `if best.release_id <= 0: return None`).

Minor: coverage pins `artwork.release_url` rather than the literal `library_item.discogs_url` field — adequate (that's where the Discogs URL surfaces via `release_url`), just noting.

PR #657, opened 2026-06-21
fix(#649): suppress SONG_AS_TRACK row-less carry-through under a "Various" anchor

### LINKED ISSUE #649: SONG_AS_TRACK carry-through anchors the #632 cache key + validation on the release-level "Various" credit for V/A comps
Follow-up from the #628 review (corroborated independently by two reviewers). In-spec-by-omission for #628 (flag-gated off), not a regression.

## Problem

For a **song-only** query routed to SONG_AS_TRACK, the #628 row-less carry-through has no typed artist, so its anchor falls back to the **release-level** credit:

```python
# lookup/orchestrator.py _match_track_releases_to_library (~1330)
anchor_artist = artist or next(
    (r.artist for r in raw_releases if r.artist and r.artist.strip()), ""
)
```

`raw_releases[i].artist` is `ReleaseInfo.artist` — the **release** credit, which is `"Various"` for a V/A compilation (the exact case the feature targets). That `anchor_artist` is used for both (a) the #632 cache key and (b) the resolve/validate call. Consequences:

1. **Cache-key collision.** The `release_resolution_cache` row is keyed `("various", <track_title>, True)`. Two different tracks that share a title across different V/A comps (e.g. two songs titled "Intro") collide under one key — the second add can surface the *first* comp's release.
2. **Validation imprecision.** `resolve_release_for_track(song, "Various", ...)` validates against the release-level "Various" credit (the cache_service "Various" fallback) rather than the track's actual credited artist, losing per-track precision.

## Why this wasn't fixed in #628

The better signal — the track's actual credited artist — is **not available at this code point**. `search_releases_by_track` sets `TrackReleasesResponse.artist` to the *query* artist (None for a song-only lookup), not the track credit (`discogs/service.py:566-572`), and `ReleaseInfo.artist` is release-level. Sourcing the per-track credit requires `_process_search_result` to extract track-artist credits — a larger change. #628 is flag-gated off (`lml_resolve_nonlibrary_release`, default False), so there is no production impact today; the SWAPPED and TRACK_ON_COMPILATION cut-sites both carry a real typed artist and are unaffected.

## Why the current tests hide it

`tests/integration/test_lookup_nonlibrary_release.py::test_song_as_track_surfaces_rowless_when_flag_on` mocks `_spine_release(artist=SPINE_ARTIST)` — an *unrealistic* V/A release whose release-artist is the real artist, not "Various" — so the "Various" anchor fallback is never exercised. The unit test passes `artist=ARTIST` explicitly. The realistic V/A-comp-via-song-only case is untested.

## Desired end state

SONG_AS_TRACK's carry-through anchors the resolve + #632 cache key on the track's **actual credited artist**, not the release-level "Various" — eliminating the cross-track cache collision and restoring per-track validation precision. Minimum viable fix if full credit-plumbing is deferred: **do not cache** (and do not validate) under a pure-"Various"/compilation-marker anchor, which removes the collision risk even without the track credit.

## Where

- `lookup/orchestrator.py` — `_match_track_releases_to_library` anchor selection (~1330-1347)
- `discogs/service.py` — `search_releases_by_track` / `_process_search_result` (the track-credit source, ~566-572)

## Acceptance criteria

- [ ] SONG_AS_TRACK carry-through for a V/A comp keys the #632 cache on the track credit (or does not cache under "Various").
- [ ] A SONG_AS_TRACK test with `_spine_release(artist="Various")` pins the real anchor behavior (today's test uses `artist=SPINE_ARTIST`, hiding it).
- [ ] No two distinct same-titled tracks collide under the `("various", <title>, True)` cache key.

## Related

- #628 (the carry-through this refines), #625 (epic), #632 (the positive-resolution cache), #646 (sibling follow-up — the album-title-probe arm)

jakebromberg:
## Problem

For a song-only query routed to SONG_AS_TRACK, the #628 row-less carry-through has no typed artist, so its anchor falls back to the release-level credit (`ReleaseInfo.artist`), which is `"Various"` for a V/A compilation — the exact case the feature targets. That `anchor_artist` fed both the #632 cache key and the resolve/validate call:

1. **Cache-key collision.** The `release_resolution_cache` row keyed `("various", <track_title>, True)` collapses two distinct tracks that share a title across different V/A comps (e.g. two songs titled "Intro") onto one row — the second add could surface the *first* comp's release.
2. **Validation imprecision.** `validate_track_on_release(..., "Various")` falls back to the release-level "Various" credit rather than the track's actual performer, losing per-track precision.

Flag-gated off (`lml_resolve_nonlibrary_release`, default False), so no production impact today; in-spec-by-omission for #628, not a regression.

## Fix (MVP)

When the SONG_AS_TRACK anchor resolves to a compilation marker (`is_compilation_artist`), suppress the carry-through entirely — no resolve, no validate, no #632 cache write — rather than surface an imprecise, collision-prone row-less item. A "Various" credit is not a usable per-track artist; what a live resolve under it would surface is validated only by track-title presence on a Various-credited release (blind to the performer), so for the common-title case it is actively misleading. SWAPPED and TRACK_ON_COMPILATION carry a typed artist and never reach this point.

This is the minimum-viable fix the issue calls out ("do not cache, and do not validate, under a pure-Various anchor"). The precise end state — plumbing the per-track credit through `_process_search_result` as the anchor — is the larger follow-up (sibling #646).

## Tests

- `test_song_as_track_various_anchor_suppresses_rowless` — the realistic default-`Various` comp surfaces nothing row-less (the prior surfacing test masked this by overriding `artist=SPINE_ARTIST`).
- `test_song_as_track_various_anchor_never_writes_632_cache` — asserts `pg.execute_calls == 0`, so nothing is keyed under `("various", …)` and no cross-track collision is possible.
- Annotated the sibling `..._surfaces_rowless_when_flag_on` so its `SPINE_ARTIST` override reads as load-bearing (a real release credit still surfaces).

Both new tests were confirmed RED against the unfixed kernel before the guard landed.

## Acceptance criteria

- [x] SONG_AS_TRACK carry-through for a V/A comp does not cache under "Various".
- [x] A SONG_AS_TRACK test with `_spine_release(artist="Various")` pins the real anchor behavior.
- [x] No two distinct same-titled tracks collide under the `("various", <title>, True)` cache key.

## Checks

`ruff format --check` + `ruff check` + `mypy` clean; full default suite 3180 passed / 1 skipped / 2 xfailed locally.

Closes #649

### DIFF
--- a/lookup/orchestrator.py
+++ b/lookup/orchestrator.py
@@ -1202,7 +1202,12 @@ async def _match_track_releases_to_library(
     ``search_releases_by_track`` probe + the per-track ``validate_track_on_release``
     credit check. The #632 cache is keyed on the typed anchor artist (the kernel's
     ``artist`` for SWAPPED; the surfaced release's own credit for SONG_AS_TRACK,
-    which has no typed artist).
+    which has no typed artist). **LML#649:** when that SONG_AS_TRACK fallback
+    credit is a compilation marker ("Various" for a V/A comp), it is not a usable
+    per-track artist — the carry-through is suppressed entirely rather than
+    resolved/validated/cached under it (which would validate only against the
+    release-level "Various" credit and collide same-titled tracks across comps on
+    the ``("various", <title>, True)`` cache key).
 
     Returns:
         Tuple of (library_items, matched_via_by_id, discogs_titles).
@@ -1332,6 +1337,24 @@ async def _match_track_releases_to_library(
         anchor_artist = artist or next(
             (r.artist for r in raw_releases if r.artist and r.artist.strip()), ""
         )
+        # LML#649: a compilation-marker anchor ("Various") is not a usable
+        # per-track artist. SONG_AS_TRACK on a V/A comp falls back here to the
+        # release-level credit, which is "Various" for the exact case the
+        # carry-through targets. Resolving under it would (a) validate only
+        # against the release-level "Various" credit — blind to the track's
+        # actual performer (validate_track_on_release's release-artist fallback)
+        # — and (b) key the #632 cache on ``("various", <title>, True)``,
+        # collapsing same-titled tracks across different comps onto one row (the
+        # second add could surface the first comp's release). Suppress the
+        # carry-through rather than surface an imprecise, collision-prone item;
+        # the precise fix plumbs the per-track credit as the anchor. SWAPPED and
+        # TRACK_ON_COMPILATION carry a typed artist and never reach this.
+        if is_compilation_artist(anchor_artist.strip()):
+            logger.info(
+                f"{label}: suppressing row-less carry-through — anchor artist "
+                f"'{anchor_artist}' is a compilation marker, not a per-track credit"
+            )
+            return matched_items, matched_via_by_id, {}
         resolved = await _resolve_nonlibrary_release(
             discogs_service,
             pg,
--- a/tests/integration/test_lookup_nonlibrary_release.py
+++ b/tests/integration/test_lookup_nonlibrary_release.py
@@ -305,7 +305,11 @@ async def test_song_as_track_surfaces_rowless_when_flag_on(library_db, enable_no
     # SONG_AS_ARTIST: no releases by the song-as-artist name.
     svc.lookup_releases_by_artist = AsyncMock(return_value=[])
     # Song-only probe surfaces the spine release (its own artist credit anchors
-    # the row-less resolve, since there is no typed artist here).
+    # the row-less resolve, since there is no typed artist here). The
+    # ``artist=SPINE_ARTIST`` override is load-bearing post-LML#649: a *real*
+    # release credit anchors and surfaces, whereas the realistic default
+    # ``"Various"`` credit is suppressed (see
+    # ``test_song_as_track_various_anchor_suppresses_rowless``).
     svc.search_releases_by_track = AsyncMock(
         return_value=_track_releases(_spine_release(artist=SPINE_ARTIST))
     )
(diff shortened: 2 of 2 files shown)


jakebromberg:
## Review: does this PR's row-less path interact with the #628 enrichment bypass?

Focused review against one risk: a surfaced row-less item reaching `enrich_one` with `release_id <= 0` or not top-1 would silently miss the #628 `library_row_acceptable` bypass and collapse to the BS#1185 `release_id=0` sentinel.

**Verdict: SAFE — and it closes a wrong-identity hazard, not just an absence.** The `is_compilation_artist(anchor)` guard returns the clean-miss shape (empty `discogs_titles` seam) before `_resolve_nonlibrary_release`, so no row-less item is emitted under a "Various" anchor. That matters beyond the cache write: without the guard, the `("<song>", "various", <album>, True)` cache key collides across comps sharing a track title, and a second add reads back the **first** comp's `release_id > 0` — which would fire the #628 bypass and surface the **wrong** album's Discogs identity (wrong-but-present, worse than not-found). Suppressing emission degrades it to a clean miss instead.

The two new tests are genuinely load-bearing: run against unguarded `main` they both fail (`..._suppresses_rowless` emits the row-less item; `..._never_writes_632_cache` fails `assert 1 == 0` — main does write the "various" key); against this PR they pass. The diff is orthogonal to the bypass — no `library_row_acceptable` / `enrich_one` / eligibility-gate lines touched. (The eventual precise fix, plumbing the per-track credit as the anchor instead of suppressing, remains a clean follow-up; this PR correctly scopes to suppression.)

Issue #583, opened 2026-06-16
Close the library-miss gap: confidence-floored Discogs search-and-cache for unknown (artist, album)


jakebromberg:
## Problem

`/api/v1/lookup` returns `results: []` whenever the SQLite library search misses, even when the requested `(artist, album)` exists on Discogs. Backend-Service writes `metadata_status: enriched_no_match` and the v2 iOS app loses its Discogs URL, release year, and Discogs-sourced artwork. The client-side v1 `DiscogsArtworkService` fallback in `wxyc-ios-64` finds the same release with a single `api.discogs.com/database/search` query, proving the data is reachable — LML just doesn't reach for it.

The orchestrator's Discogs-touching steps are all gated on `library_results` being non-empty:

| Step | Line | Gate |
|---|---|---|
| Track validation | https://github.com/WXYC/library-metadata-lookup/blob/2928498a7be78f48d018028af9573cc1148be851/lookup/orchestrator.py#L2989 | `if library_results and ...` |
| Streaming status | https://github.com/WXYC/library-metadata-lookup/blob/2928498a7be78f48d018028af9573cc1148be851/lookup/orchestrator.py#L3029 | `if library_results ...` |
| Artwork fetch | https://github.com/WXYC/library-metadata-lookup/blob/2928498a7be78f48d018028af9573cc1148be851/lookup/orchestrator.py#L3036 | `if library_results:` |
| Metadata enrichment | https://github.com/WXYC/library-metadata-lookup/blob/2928498a7be78f48d018028af9573cc1148be851/lookup/orchestrator.py#L3051 | `if items_with_artwork:` |
| Identity resolution | https://github.com/WXYC/library-metadata-lookup/blob/2928498a7be78f48d018028af9573cc1148be851/lookup/orchestrator.py#L3091 | `if entity_store and library_results:` |

The `include_external_caches` opt-in at https://github.com/WXYC/library-metadata-lookup/blob/2928498a7be78f48d018028af9573cc1148be851/lookup/orchestrator.py#L3132 is mojibake-recovery only — returns a text-only `LibraryCatalogItem(id=0, …, call_number="(external)")` with no artwork, no `release_url`, no enrichment. Backend-Service never sets the flag.

This contradicts LML's mission: metadata for arbitrary `(artist, album, song)` queries with on-demand caching. The library is the bootstrap, not the gate.

## Concrete failure

[...]

PR #658, opened 2026-06-21
A2/A4: row-less cached-track release_id + soft no-album confidence (#629)

### LINKED ISSUE #629: A2/A4: widen the library-miss probe and stop the cached-track safety net dropping resolved releases
Sub-issue of #625 (fronts **A2** + **A4**). Blocked by **#628** (reuses its row-less carry-through and #633's bound).

## A2 — widen the library-miss probe

The only path returning Discogs metadata with an empty library requires artist **and** album ([orchestrator.py:3207](https://github.com/WXYC/library-metadata-lookup/blob/9d9644dd22dbf2d2b5c86ecba9a30db6948864c0/lookup/orchestrator.py#L3207)). A **song-only** or **artist+song (no album)** query for a resolvable non-library release returns empty.

Fix: widen the precondition / route the strategies' resolved releases through #628's row-less carry-through so the song-only / artist+song shapes reach it.

**No-album disambiguation (locked in the #625 grill):** with no typed album, prefer the artist's own release — order validated candidates by `is_compilation = False` first, then stable `release_id` — and surface the best via #633's sequential-early-exit. Carry a slightly lower `confidence` on no-album results so a consumer can treat them as soft. (This is the no-album ordering #633 defers to; revisit concurrency here if the top candidate fails often.)

## A4 — cached-track safety net carries the release_id

`find_library_albums_with_cached_track` ([orchestrator.py:2030](https://github.com/WXYC/library-metadata-lookup/blob/9d9644dd22dbf2d2b5c86ecba9a30db6948864c0/lookup/orchestrator.py#L2030), called [L3244](https://github.com/WXYC/library-metadata-lookup/blob/9d9644dd22dbf2d2b5c86ecba9a30db6948864c0/lookup/orchestrator.py#L3244)) pulls track-confirmed releases from the PG cache, then drops any whose fuzzy library row fails `artist_matches_item` ([L2073-2074](https://github.com/WXYC/library-metadata-lookup/blob/9d9644dd22dbf2d2b5c86ecba9a30db6948864c0/lookup/orchestrator.py#L2073-L2074)) and returns bare `LibraryItem`s — the resolved `release_id` is never threaded out.

Fix: thread the resolved `release_id` into the row-less result; don't drop a cache-confirmed release for want of an artist-matching library row.

## TDD seed (write first)

- `tests/integration/test_lookup_nonlibrary_no_album.py`: an artist+song (no album) query for a resolvable non-library track → row-less result with `discogs_url`; when the track is on the artist's own album **and** a comp, assert the **own album** (non-comp) is surfaced.
- `tests/unit/test_cached_track_safety_net.py`: `find_library_albums_with_cached_track` returns the resolved `release_id` for a cache-confirmed release with no artist-matching library row.

## Acceptance criteria

- [ ] song-only / artist+song (no album) non-library query returns a resolved `discogs_url` (A2), best-ranked by prefer-own-release.
- [ ] `find_library_albums_with_cached_track` threads `release_id` and stops dropping on the artist gate (A4).
- [ ] Reuses #628's carry-through + #633's bound (no new uncached path).
- [ ] Unit + integration coverage.

## Related

- #625 (parent), #628 (carry-through), #633 (no-album ordering), #583 (library-miss Step 3a), #400 (typed-match floor)

jakebromberg:
Closes #629. Sub-issue of #625 (the non-library-resolution epic), fronting **A4** and the residual of **A2**.

## A4 — cached-track safety net threads `release_id`

`find_library_albums_with_cached_track` confirmed a track on a Discogs release via the PG cache but **dropped** it whenever no WXYC library row artist-matched (`if not artist_matches_item(...): continue`), returning bare `LibraryItem`s — the resolved `release_id` never threaded out. A song-only / artist+song add for such a release lost resolution entirely.

It now surfaces the best cache-confirmed release **row-less**, reusing #628's `{0: ResolvedRelease}` carry-through seam so the `release_id` (hence `discogs_url`) still binds in Step 4. Specifics:

- Return type widens to `tuple[list[LibraryItem], dict[int, ResolvedRelease]]` (the same 2-tuple `search_compilations_for_track` already uses); the `perform_lookup` caller unpacks it and merges the resolved map into `discogs_titles`.
- Ordering follows the **#629 no-album rule** — prefer the artist's own release (`is_compilation=False`) over a compilation, then stable `release_id`.
- Gated on `lml_resolve_nonlibrary_release`, matching the other carry-through sites: flag-off preserves the exact pre-#629 drop (`fetch_artwork_for_items` only binds a row-less item when the flag is on).
- Cache rows are already track-confirmed by the trigram query, so this is **pure cache reuse — no new uncached Discogs path** (acceptance criterion).

## A2 — soft confidence on no-album results

A row-less bind from a no-album (song-only / artist+song) query now carries `ROWLESS_NO_ALBUM_CONFIDENCE` (0.8) instead of 1.0. The (artist, track) pair validated, but with no typed album the chosen release is the best-ranked guess, not a user-confirmed album — so a consumer (request-o-matic) can treat it as soft. Album-typed binds keep the full 1.0. `parsed.album` is threaded through `fetch_artwork_for_items` → `_bind_resolved_release` to gate it.

**Scope note:** A2's "route the no-album shapes through the carry-through" and the prefer-own-release ordering already shipped via #628 and #633 (`prerank_candidates_for_validation`). The new integration test pins them as regression guards rather than re-implementing them.

## Tests (TDD)

- `tests/unit/test_cached_track_safety_net.py` (new TDD seed) — row-less surfacing with `release_id`, prefer-own-over-comp ordering, flag-off drop.
- `tests/integration/test_lookup_nonlibrary_no_album.py` (new TDD seed) — end-to-end: artist+song no-album → row-less `discogs_url`, own-album-over-comp, soft confidence.
- `tests/unit/test_orchestrator_helpers.py` — new `TestRowlessNoAlbumConfidence` (soft vs full), and the existing 11 cached-track tests updated to the tuple contract.

Full suite green (3185 passed), `ruff check` + `ruff format --check`, and `mypy . --ignore-missing-imports` all clean locally.

## Acceptance criteria

- [x] song-only / artist+song (no album) non-library query returns a resolved `discogs_url` (A2), best-ranked by prefer-own-release.
- [x] `find_library_albums_with_cached_track` threads `release_id` and stops dropping on the artist gate (A4).
- [x] Reuses #628's carry-through + #633's bound (no new uncached path).
- [x] Unit + integration coverage.

## Related

- #625 (parent epic), #628 (carry-through spine), #633 (bounded resolve / no-album ordering), #583 (library-miss Step 3a), #400 (typed-match floor)

### DIFF
--- a/lookup/orchestrator.py
+++ b/lookup/orchestrator.py
@@ -88,6 +88,7 @@ from lookup.models import LookupRequest, LookupResponse, LookupResultItem
 from lookup.release_resolution import (
     ResolvedRelease,
     merge_wave_b_compilations,
+    prerank_candidates_for_validation,
     rank_resolved_releases,
     resolve_release_for_track,
     resolve_release_for_track_cached,
@@ -1028,14 +1029,16 @@ this floor is replaced with ``library_identity.confidence`` per row.
 
 
 ROWLESS_NO_ALBUM_CONFIDENCE: float = 0.8
-"""Soft confidence for a row-less non-library release surfaced by a *no-album*
-query (A2, LML#629).
+"""Soft confidence for a row-less non-library release that was not album-matched
+(A2, LML#629).
 
-A song-only / artist+song query validates the (artist, track) pair, but with no
-typed album the bound release is the best-ranked candidate (prefer-own-release,
-then stable release_id) — not a user-confirmed album. Carrying a confidence
-below the album-typed 1.0 lets a consumer (request-o-matic) treat it as soft.
-Album-typed binds keep the full 1.0: the user named the album.
+Applied when a row-less bind is *not* a user-confirmed album — either because the
+request typed no album (song-only / artist+song), or because the binding route
+never matched the release against the typed album. The latter is the cached-track
+safety net (A4), which picks by track only; it stamps this on the seam so the
+bind stays soft even when an album was typed. An album-ranked carry-through
+(A1/#628 with a typed album) keeps the full 1.0 — the typed album shaped the
+pick. Lets a consumer (request-o-matic) treat the soft results as tentative.
 """
 
 # Synthetic library-id for a row-less result — a Discogs release with no WXYC
@@ -2484,26 +2487,28 @@ async def find_library_albums_with_cached_track(
     # surface the best one row-less so its release_id (hence discogs_url) still
     # binds via #628's {0: ResolvedRelease} seam. Gated on the same flag as the
     # other carry-through sites: when off, fetch_artwork_for_items won't bind a
-    # row-less item, so we preserve the pre-#629 drop. No-album ordering (the
-    # #629 rule prerank_candidates_for_validation applies on the bounded resolve
-    # path): prefer the artist's own release (is_compilation=False) over a
-    # compilation, then stable release_id. Cache rows are already track-confirmed
-    # by the trigram query, so no re-validation is needed.
+    # row-less item, so we preserve the pre-#629 drop. Cache rows are already
+    # track-confirmed by the trigram query, so no re-validation is needed.
+    #
+    # Ordering reuses the shared #629 no-album rule (prefer is_compilation=False,
+    # then stable release_id) instead of restating it — keeping this path and the
+    # bounded-resolve path on one definition. There is no typed album to rank
+    # against here (the cache keyed on track only), so confidence is soft: the
+    # pick was never album-matched, and the soft value rides the seam so the bind
+    # surfaces it even when the request did type an album.
     if not get_settings().lml_resolve_nonlibrary_release:
         return [], {}
-    best = min(
-        (r for r in cached_releases if r.release_id),
-        key=lambda r: (r.is_compilation, r.release_id),
-        default=None,
-    )
-    if best is None:
+    ranked = prerank_candidates_for_validation([r for r in cached_releases if r.release_id], None)
+    if not ranked:
         return [], {}
+    best = ranked[0]
     rowless = _make_rowless_item(artist=artist or "", title=best.album)
     resolved = ResolvedRelease(
         release_id=best.release_id,
         release_url=best.release_url or "",
         is_compilation=bool(best.is_compilation),
         album_title=best.album or "",
+        confidence=ROWLESS_NO_ALBUM_CONFIDENCE,
     )
     return [rowless], {ROWLESS_LIBRARY_ID: resolved}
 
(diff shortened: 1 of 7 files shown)


jakebromberg:

Focused review against one risk: a surfaced row-less item reaching `enrich_one` with `release_id <= 0` or not top-1 would silently miss the #628 `library_row_acceptable` bypass and collapse to the BS#1185 `release_id=0` sentinel — the exact not-found outcome this ticket exists to kill.

**Verdict: SAFE.** Both bypass preconditions hold for every row-less item this PR emits. `release_id` is filtered truthy at each emit site (A4: `best = min((r for r in cached_releases if r.release_id), ...)` then `if best is None: return [], {}`; A2 via the guarded `_resolve_nonlibrary_release`). And the row-less item is always the sole result, hence top-1 (A4 sets `library_results = promoted = [rowless]`; A2 yields `[rowless]`, with Step-3b's `real_results = [r for r in library_results if r.id != 0]` keeping the id==0 item out of re-validation so nothing displaces it from index 0). `test_artist_song_no_album_surfaces_rowless_with_soft_confidence` pins `artwork.release_id == 12345 (> 0)` + the row-less shape end-to-end through the real `enrich_one` path; flag-off emits no item.

Minor, non-blocking: the second integration test (`test_no_album_prefers_own_album_over_compilation`) doesn't re-assert confidence/`release_url` — the first test covers all three on the same fixture, so coverage isn't a gap.

jakebromberg:
Confirmed the **SAFE** verdict against the code — the bypass at [`orchestrator.py:3004`](https://github.com/WXYC/library-metadata-lookup/blob/df3f574/lookup/orchestrator.py#L3004) (`item.id == ROWLESS_LIBRARY_ID and artwork is not None and artwork.release_id > 0`) forces `library_row_acceptable` true, so the row-less item keeps its `release_id`. Both emit sites filter `release_id` truthy.

One precision on the review's reasoning (doesn't change the verdict): the L3004 bypass is **position-independent** — it preserves `release_id` regardless of index. `is_top1 (idx==0)` only gates `is_album_derived_eligible` (album-derived year/label enrichment), not the sentinel-collapse. So "not top-1" wouldn't drop to `release_id=0`; it'd just skip some album-derived enrichment. Moot here since the row-less item is always sole/top-1.

The one thing not covered: the cited integration test exercises the **A2 / TRACK_ON_COMPILATION** carry-through (`cache_service=None`) — nothing drove `perform_lookup` down the **A4 cached-track branch**, so the A4 caller wiring (the `discogs_titles = {**discogs_titles, **promoted_titles}` merge) was untested end-to-end. A broken merge there would still pass the unit tests but collapse to the sentinel. Added `tests/integration/test_lookup_cached_track_rowless.py` (df3f574) to drive that branch and assert `artwork.release_id > 0` survives the merge → bind → enrich chain, plus the soft confidence and flag-off inert case.

Issue #672, opened 2026-06-22
fix(streaming): unify the two streaming_availability.db lineages — CI release lacks Apple/track_results, sync reads release not volume


jakebromberg:
## Problem

`streaming_availability.db` is maintained as **two independent copies that drift apart**, and production reads the thinner one:

- **`streaming-data-v1` release asset** — updated weekly by `refresh-streaming.yml` (CI). CI has only Spotify creds, so the run does **Spotify + Deezer only**: Apple is gated on `APPLE_MUSIC_TEAM_ID/KEY_ID/PRIVATE_KEY` and skipped — logged, not failed — (`scripts/streaming_availability/__main__.py:842-851`), and the separate `scripts/track_streaming` pipeline is never run. This copy is what `WXYC/discogs-etl` `sync-library.yml` reads to enrich `library.db`.
- **Railway volume copy** (`POST/GET /admin/{upload,download}-streaming-db`, added in #238) — the **rich local working DB**: full pipeline *with* Apple creds + `track_streaming` + deeper coverage. Updated only by manual `/admin/upload-streaming-db`.

Because the two are never reconciled and **sync reads the release**, production `library.db` silently carried **zero Apple Music links** while 288 Apple URLs + 72,341 `track_results` sat unused in the volume backup. Discovered 2026-06-22 while executing the #668 Bandcamp drain: the release (CI) and volume (local, last backed up ~2026-04-20) had forked across services — the release had fresher Spotify/Deezer *timestamps* but fewer URLs and no Apple; the volume was richer everywhere but with stale timestamps. It was reconciled by hand for that run (a lossless merged superset was republished to the release + both volumes; sync then surfaced Apple 0→294 and bandcamp +211 into prod `library.db`), but the **structural cause remains** and the copies will re-fork.

`docs/deployment.md` already documents the *intent* that the volume be canonical (that is why the download endpoint exists, #238), but `sync-library.yml` was never migrated to read it.

## Desired end state

One authoritative `streaming_availability.db` lineage feeding production, such that Apple + track-level data reach `library.db` automatically and the release/volume cannot silently diverge. The clean target is a **single file with a single weekly writer**; the two options differ only in *which* file that is.

[...]

PR #676, opened 2026-06-22
feat(#672): make the Railway volume the canonical streaming_availability.db

jakebromberg:
Part of #672 (LML side, PR-1 of the cross-repo cutover). Implements **option 2** (volume canonical) from the issue.

## Problem

`streaming_availability.db` was maintained as two copies that drifted: the `streaming-data-v1` GitHub Release asset (written weekly by `refresh-streaming.yml`, CI-only creds → Spotify/Deezer-only) and the Railway **volume** copy (the rich working DB with Apple + `track_streaming`, uploaded manually). The daily library-sync read the *release*, so production `library.db` silently carried **zero Apple Music links** while hundreds sat unused on the volume.

## Changes

- **Coverage-regression guard** on `POST /admin/upload-streaming-db` (`routers/admin.py`). The upload — a full-file replace — is now compared against the file **on disk at replace time** across five metrics (`COUNT(apple_url|spotify_url|deezer_url)` over `albums`, `COUNT(*)` of `albums`, `COUNT(*)` of `track_results`) and rejected with **409** if any drops below `prior × (1 − 0.05)` **or** goes non-zero → zero. `?force=true` overrides an intentional shrink (logged loudly). `_streaming_coverage` reads a missing file/table/column as `0`, so a disappearing column is caught as an N → 0 regression. Comparing against live on-disk state also serializes the two-writer race (a stale writer that lost Apple is rejected and re-applies next cycle).
- **`refresh-streaming.yml`** round-trips the volume: download via `GET /admin/download-streaming-db` (hard-fail — the `|| echo` empty-db fallback is **dropped**) → Spotify/Deezer incremental → upload via `POST /admin/upload-streaming-db`. `library.db` (the input catalog) still comes from the release. The release asset is **transitionally dual-written** until discogs-etl cuts over.
- **`docs/deployment.md`**: documents the single enforced lineage, the coverage guard, and the Railway-uptime failure mode.

## Tests

16 new tests in `tests/unit/test_admin_streaming_guard.py` (pure-function tolerance/boundary maths + endpoint integration incl. the 288→0 incident, force override, missing-`track_results` table). Full suite green: **2999 unit + 269 integration**, ruff + actionlint clean.

## ⚠️ Operational prerequisite

`refresh-streaming.yml` needs **`ADMIN_TOKEN`** and **`PRODUCTION_URL`** repo secrets on `WXYC/library-metadata-lookup` (it currently has only Railway/Spotify/Discogs). Add them before the next weekly run, or the volume round-trip will fail (loudly, not silently).

## Rollout

1. **This PR** — guard + volume round-trip + docs (dual-writes the release as a safety net). Merge first.
2. **WXYC/discogs-etl** companion PR — reads the volume with hard-fail + apple floor.
3. Verify one sync cycle yields non-zero `apple_music_url` in prod `library.db`.
4. Follow-up PR — drop the release dual-write, retire the `streaming_availability.db` release asset (kept for `library.db`). **Gated on step 3; not in this PR.**


jakebromberg:
Companion (consumption-side) PR: WXYC/discogs-etl#290. Merge this PR first; #290 lands after.

Issue #683, opened 2026-06-23
Wire automated alerts (Sentry + PostHog) on the LML#681 row-less-flag counters so LML_RESOLVE_NONLIBRARY_RELEASE degradation pages instead of going silent


jakebromberg:
## Problem

`LML_RESOLVE_NONLIBRARY_RELEASE` (LML#628 row-less carry-through) was enabled in **production 2026-06-23**. LML#681 made the flip *observable* but **not alarmed**: the new counters/tags ride the `cache_stats` → PostHog/Sentry seam as dashboard telemetry, and there is **no alert-as-code anywhere in the LML repo** (no CloudWatch/PagerDuty/PostHog/Sentry alert config). So today a regression in this feature is caught only if a human happens to be looking at a dashboard.

This matters because the feature's worst failure mode is **silent by design**: the row-less resolve path swallows its Discogs probe errors (returns `[]`/`None`), so a Discogs outage shows up as *fewer* row-less surfaces, **not** an error spike in Sentry — LML#681 explicitly scoped silent-degradation alerting *out*. The only machine-readable signals are the LML#681 counters, and nothing watches them. The cold-path Discogs cost is also the exact sensitivity behind the 2026-05-13 regression (#337): a call-rate spike here can degrade `/lookup` latency for the whole live stack.

## Desired end state

A Discogs call-rate spike or a cache miss-rate climb on the row-less path **pages/notifies automatically** — on **both** Sentry and PostHog — instead of being noticed late or never. The kill-switch (`railway variables --set "LML_RESOLVE_NONLIBRARY_RELEASE=false" --environment production`) is then a deliberate response to an alert, not a guess.

## Where (the signals to alert on — permalinks @ `0b88aa8`)

All flow through the existing `cache_stats` seam — **no new code needed in LML to emit them**:

- Per-request payload keys: [`_LML_CACHE_STATS_EXTRA_KEYS`](https://github.com/WXYC/library-metadata-lookup/blob/0b88aa8abe5ee29e3c4d12b3d25c7b4ad6d083fd/lookup/router.py#L85) — seeded so the shape is stable every request.
[...]

PR #756, opened 2026-07-10
feat(#755): Discogs saturation circuit-breaker sheds load to cache-only

### LINKED ISSUE #755: Discogs saturation should shed load (circuit-breaker), not queue+retry into caller timeouts
## Problem
On 2026-07-10 the daily BS `flowsheet-metadata-backfill` flood drove LML `/lookup` into `LML request timed out` then `502 Bad Gateway` for the whole live stack, and `/health` showed `discogs_api: rate-limited` for ~8h. Root cause is NOT a missing rate limiter — LML already has a correct process-global outbound Discogs limiter: `AsyncLimiter(discogs_rate_limit=50, 60)` + a 5-permit semaphore in `discogs/ratelimit.py`, applied per-attempt in `discogs/service.py` (`_request`, https://github.com/WXYC/library-metadata-lookup/blob/c0d8611/discogs/service.py#L386-L457), with 429 retry + jittered exponential backoff up to ~62s (`discogs_max_retries=5`, https://github.com/WXYC/library-metadata-lookup/blob/c0d8611/config/settings.py#L117-L129).

The gap is that under sustained over-demand the limiter converts excess load into **unbounded per-request latency** instead of shedding it: every lookup fans out ~5 live Discogs calls (the #683 `api_calls` metric hit avg 5.21/req during the flood), those calls queue on the 50/min `rate_limiter.acquire()`, and 429s add up to ~62s of retry backoff per call — so a lookup's wall time grows past BS's 35s client timeout → timeouts, then worker saturation → 502s. LML already *knows* it is saturated (`/health` reports `discogs_api: rate-limited`, and Discogs `X-Discogs-Ratelimit-Remaining` is read at `discogs/service.py:434`) but the lookup path does not act on it.

## Desired end state
When Discogs is saturated (rate-limit-remaining exhausted / 429 rate or `rate_limiter` queue-depth over a threshold), a circuit-breaker trips and lookups **fast-fail to cache-only** — return library + cached-Discogs results and skip the live Discogs fan-out — so the service degrades gracefully (fast, partial metadata) instead of every request queuing into a timeout. Breaker opens on saturation, half-opens after a cool-down, closes when `X-Discogs-Ratelimit-Remaining` recovers. Emits a `cache.*`-style counter so the #683 observability surface can alert on breaker-open time.

## Where
- `discogs/service.py` `_request` — breaker check before `rate_limiter.acquire()`; feed the breaker from 429s + `X-Discogs-Ratelimit-Remaining`.
- `discogs/ratelimit.py` — breaker state alongside the per-loop limiter/semaphore.
- `lookup/` orchestration seam where the live Discogs cascade is invoked (a cache-only path already exists for cache hits — this shorts to it).
- `config/settings.py` — new threshold / cool-down env vars.
- `/health` `discogs_api` already surfaces the state (health handler in `main.py`) — reuse the same signal.

## Constraints / notes
- Cache-only degrade must preserve existing match quality for cache hits; only the live-probe tail is shed.
- Complements (does not replace) the caller-side limiters (BS `BACKFILL_LML_*`). The point is that LML self-protects regardless of which caller floods, because the Discogs 60/min ceiling is shared across the backfill **plus** organic iOS/dj-site traffic and no caller-side limiter can see that aggregate.
- Reuses the closed Epic A perf direction (#337 / #338: cap wall-clock budgets, cache negatives) and the #683 degradation-alert seam.
- **Open question for impl:** if LML runs multiple uvicorn workers, the per-event-loop limiter/breaker is per-worker, not fleet-global — verify aggregate outbound stays under 60/min (size per-worker or coordinate).

## Acceptance criteria
- Under simulated Discogs saturation (forced 429s), `/lookup` p99 stays under BS's 35s timeout by returning cache-only results instead of queuing; no 502s.
- Breaker-open time is emitted as a counter and is visible to the #683 alert surface.
- Organic cache-hit lookups are unaffected when Discogs is healthy.
- Unit/integration test for the open → half-open → close transitions (TDD per repo policy).

## Context
2026-07-10 incident: BS `flowsheet-metadata-backfill` cron (`0 6 * * *` UTC) ran 8h at ~82% cache-miss, ~5.21 Discogs calls/req, tripping the #683 "Discogs call-rate" alert and producing LML timeouts → 502s. Mitigated by stopping the cron; `/health` recovered within seconds. This ticket is the durable server-side fix; the caller-side reduction is tracked in the Backend-Service backfill drain-order ticket.

## Related
- WXYC/library-metadata-lookup#337, WXYC/library-metadata-lookup#338 (closed Epic A perf: budgets + negative caching — same direction)
- WXYC/library-metadata-lookup#683 (row-less flag degradation alerts — observability seam this reuses)
- WXYC/Backend-Service#1591 (backfill play-priority drain — caller-side complement)

jakebromberg:
## Problem

On 2026-07-10 the daily Backend-Service `flowsheet-metadata-backfill` cron flooded LML `/lookup`, driving `LML request timed out` → `502 Bad Gateway` across the live stack; `/health` showed `discogs_api: rate-limited` for ~8h.

LML already has a correct process-global outbound Discogs limiter (`AsyncLimiter(50, 60)` + a 5-permit semaphore in `discogs/ratelimit.py`, applied per-attempt in `discogs/service.py:_request_with_retry`). The gap is that under **sustained over-demand** the limiter converts excess load into *unbounded per-request latency* rather than shedding it: each lookup fans out ~5 live Discogs calls, they queue on the 50/min `rate_limiter.acquire()`, and 429 backoff stacks to ~62s/call — so a lookup's wall time exceeds Backend-Service's 35s client timeout → timeouts → worker saturation → 502s. LML *knew* it was saturated (reads `X-Discogs-Ratelimit-Remaining`, `/health` = `rate-limited`) but the lookup path never acted on it.

## What this does

Adds the missing feedback loop: a **saturation circuit-breaker** that makes lookups fast-fail to **cache-only** when Discogs is saturated — return library + already-cached-Discogs results and skip the live fan-out — so the service degrades gracefully (fast, partial metadata) instead of every request queuing into a timeout.

- `discogs/breaker.py` — `DiscogsCircuitBreaker`, a three-state machine (CLOSED → OPEN → HALF_OPEN → CLOSED). Fed from 429s (`record_failure`) and `X-Discogs-Ratelimit-Remaining` (`record_success(remaining)`).
- `discogs/ratelimit.py` — `get_discogs_breaker()` holds breaker state in a per-event-loop dict, the same pattern/scope as the limiter and semaphore.
- `discogs/service.py:_request_with_retry` — checks the breaker **before** `rate_limiter.acquire()`. When OPEN it returns `None` immediately (no network, no limiter wait, no backoff). Every call site already guards with `if response is not None`, so the seam already exists — the lookup degrades to whatever the library + cached-Discogs legs produced. Cache hits never reach here (they short-circuit in `fallthrough()`), so **only the live-probe tail is shed and healthy cache-hit behavior is unchanged**.
- `config/settings.py` — three new env vars (see below).
- Emits `discogs_breaker_open_shed` on the #683 `cache.*` counter surface (`get_cache_stats_recorder().record(...)`) on every shed, so breaker-open time is alertable on the same PostHog/Sentry seam as the row-less flag degradation alerts.

### State machine
- **CLOSED** → requests flow. Trips OPEN on `DISCOGS_BREAKER_FAILURE_THRESHOLD` consecutive 429s **or** a response reporting remaining ≤ `DISCOGS_BREAKER_REMAINING_FLOOR`. A single healthy response resets the failure run.
- **OPEN** → sheds live probes for `DISCOGS_BREAKER_COOLDOWN_SECONDS`, then admits one half-open trial.
- **HALF_OPEN** → one trial in flight (concurrent callers still shed); a healthy trial CLOSES, a 429 / still-exhausted remaining re-OPENS.

## New env vars (defaults)

| Var | Default | Meaning |
|---|---|---|
| `DISCOGS_BREAKER_FAILURE_THRESHOLD` | `8` | Consecutive 429s that trip OPEN (sized above the brief flurry one warm-cache lookup absorbs via retry). |
| `DISCOGS_BREAKER_REMAINING_FLOOR` | `3` | Proactive trip on `X-Discogs-Ratelimit-Remaining` ≤ this; `0` disables. |
| `DISCOGS_BREAKER_COOLDOWN_SECONDS` | `20.0` | OPEN duration before a half-open trial (~⅓ of the Discogs 60s window). |

All read via pydantic-settings; unparseable/zero/negative fall back to the default with a WARN. Read at breaker-build, so a change takes effect on the next restart. `DISCOGS_*` prefix matches the existing `DISCOGS_RATE_LIMIT` / `DISCOGS_MAX_CONCURRENT` knobs.

## Multi-worker finding

**Single worker.** `entrypoint.sh` runs bare `uvicorn main:app --host 0.0.0.0 --port ${PORT}` with no `--workers` (default 1), confirmed against `railway.toml` (Dockerfile builder, no worker override). So the per-event-loop breaker is **process-global today** and the default thresholds are sized directly against the single shared 60/min Discogs ceiling — no cross-worker coordinator needed. The per-worker caveat (re-size before `UVICORN_WORKERS > 1` ships — tracked in LML#747) is documented in `discogs/breaker.py`, `docs/env-vars.md`, and `docs/architecture.md`, matching the existing `LML_LOOKUP_MAX_CONCURRENT` / `LML_BULK_GLOBAL_MAX_CONCURRENT` convention.

## Tests (TDD, pure unit — no `pg` / `external_api` markers)

- `tests/unit/test_discogs_breaker.py` — the state machine: starts CLOSED; closed→open on consecutive 429s and on the remaining-floor; healthy success resets the failure run; stays open before the cool-down; half-opens after it and admits exactly one trial; half-open success closes, half-open 429 / still-saturated remaining reopens.
- `tests/unit/test_discogs_breaker_request.py` — `_request_with_retry` integration with a mocked Discogs client: OPEN breaker short-circuits to `None` without touching the limiter/semaphore or client; emits the counter; CLOSED + healthy is a pure pass-through (cache-miss lookups unaffected); sustained 429s recorded through the request path trip the breaker so the next call sheds; and an **AC1 bounded-wall-time** test asserting an OPEN breaker returns in <0.5s even against an upstream that would otherwise sleep a 30s `Retry-After`.

## Review follow-up (second commit)

A `/code-review max` pass surfaced correctness gaps, all fixed:

1. **Shed must not poison durable caches.** The shed now **raises `DiscogsBreakerOpenError`** rather than returning `None`. A shed is "couldn't ask", never "confirmed empty/false", so it must not persist a 7-day `lookup_negative` row, a `set_cached_release_id(release_id=None)` null-pin, an `@async_cached` memoized `[]`, or a definitive `validate_track_on_release` → `False` that drops a valid candidate. `search_releases_by_track` returns `None` on a shed **and** on retry-exhaustion/httpx-error (fixing pre-existing `None`-laundering); `resolve_release_for_track` raises at entry when OPEN so its L1 wrapper never memoizes and the row-less binder never null-pins; `get_release` re-raises the shed (genuine 404/None/5xx still → `None`).
2. **Make it trip.** The proactive floor now rides the 429 path, and the counting unit is **failed requests, not attempts** — `_request_with_retry` records exactly one terminal outcome per call. Default threshold lowered to **4** (reachable under per-request counting).
3. **Shed in-flight.** The retry loop re-checks the breaker each attempt and sheds mid-flight instead of riding the full ~62s backoff.
4. **Register the counter.** `discogs_breaker_open_shed` added to `_LML_CACHE_STATS_EXTRA_KEYS` so it seeds to 0 and is a queryable baseline (AC2).
5. **5xx is neutral.** A 5xx neither resets the failure run nor closes a half-open trial.
6. **HALF_OPEN epoch guard.** A monotonic epoch prevents a stale response (CLOSED-era or superseded trial) from deciding the current trial; a 429-then-200 trial closes correctly.
7. **Nits.** OPEN/CLOSED logged once on transition (per-shed is `debug`); `ge=` bounds on the three settings Fields (out-of-range now raises a startup validation error — env-vars doc corrected to match, no silent fallback); `force_open` trimmed to a test seam; off-loop-callers-get-no-shedding note; stronger request-path test assertions.

## Round-2 review follow-up (third commit)

Round 2 caught that the raise-based shed was half-wired (raises but no catch boundary → a shed 500s on `/lookup`) plus a stuck-breaker bug. All four fixed:

- **R2-1 (stuck HALF_OPEN latch).** The mid-flight re-check used the state-mutating `allow_request()`, which promoted the *retrying* request to the half-open trial under a fresh epoch — but that request holds its stale entry epoch, so its terminal `record_*` epoch-mismatched, never CLOSED, and every other request was then shed in HALF_OPEN → the breaker latched until restart. Added a **read-only, epoch-aware `should_shed_inflight(epoch)`** predicate (no epoch bump, no promotion) and use it for the re-check. Regression test proves the breaker **recovers** after an in-flight shed during a cool-down-elapsed window.
- **R2-2 (shed 500s on `/lookup` + `/lookup/bulk`).** Established **one catch boundary** in the `core/search.py` per-strategy try — a `DiscogsBreakerOpenError` from any strategy converts to `Outcome.empty()` (cache-only) and the cascade `continue`s. Added a **`perform_lookup` backstop** (`_build_degraded_response`) covering the enrichment tail + identity resolution. Integration tests assert `/lookup` (SWAPPED + TRACK_ON_COMPILATION) and `/lookup/bulk` return **200 cache-only** under a forced-open breaker, never 500.
- **R2-3 (raw 500 on diagnostic routes).** `GET /discogs/release/{id}` and the `/discogs/entity/release/{id}` release branch now catch the shed → **HTTP 503** (transient/retryable).
- **R2-4 (shed laundered into song-not-found).** `lookup/validation.py:validate_one` now catches the shed *specifically* (narrower than its broad `except`) and **keeps the library row** unvalidated rather than dropping it to `None` (which had marked `song_not_found` — a wrong 200). Genuine validation errors still drop the row.

Unifying principle: `DiscogsBreakerOpenError` means "couldn't ask" and degrades to cache-only / keep-existing at **every** boundary — never a 500, never a confirmed-negative.

## Deferred review items

Left for the user to decide (not implemented in this PR):

- **Cooldown default (20s) vs the 60s rate-limit window** — potential oscillation is a tuning judgment; leaving the conservative 20s (⅓ window) default and revisiting with prod telemetry.
- **Wiring the breaker's OPEN state into `/health` `discogs_api`** — the ticket suggested reusing that signal; moderate scope (the health probe is a separate `check_api` path from the lookup breaker), deferred to keep this PR focused.
- **Budget-aware retry backoff (`caller_budget_ms`)** — pre-existing and broader than this breaker; out of scope here.

## Local checks
- `ruff check .` — clean
- `ruff format --check .` — clean
- Unit suite (`-m 'not pg and not external_api'`) — **3278 green** (3249 baseline + 29 new across the three commits)
- Integration suite — 276 passed (the one failure, `test_pg_source_bad_credentials`, is pre-existing and environment-specific: it needs a live Postgres and fails identically on `origin/main`; it passes in CI's PG container)

## Related flood levers
This is the durable **server-side** self-protection. It complements, and does not replace, the caller-side reductions:
- WXYC/Backend-Service#1591 — backfill play-priority drain (caller-side)
- WXYC/discogs-etl#308 — cache coverage (fewer cache-miss live probes)

Closes #755


jakebromberg:
Merged. The three deferred review items are routed as: **#757** (surface breaker state on `/health discogs_api` — the ticket's original 'reuse the signal' half) and **#758** (budget-aware retry backoff — pre-existing, exposed here) are filed on [Project #32](https://github.com/orgs/WXYC/projects/32). Cooldown left at 20s (revisit with telemetry).

Issue #766, opened 2026-07-12
entity.identity mint hardening: store-level fill-if-null primitive, clobber provenance, and case-sibling awareness


jakebromberg:
## Problem

The #759 bare-name resolver (PR C1, #765) and `release/orchestrator.py:_writeback_identity` both mint `entity.identity.discogs_artist_id` via `EntityStore.upsert_identity`, whose ON CONFLICT arm is `COALESCE(EXCLUDED.discogs_artist_id, entity.identity.discogs_artist_id)` — **new-wins-when-non-null**. The store's docstring promises only NULL-safety ('populated fields are never overwritten with NULL'), but both writers treat it as first-write-wins. Consequences, confirmed during #765's review:

1. **TOCTOU clobber**: the resolver's no-overwrite property rests entirely on its tier-1 read gate, which is up to ~30s stale at mint time (serial 50/min API pacing across a 25-name batch). A concurrent writer that sets `discogs_artist_id=X` for the same `library_name` in that window is silently overwritten with the resolver's bare-name winner Y. The realistic conflicting writer is `_writeback_identity` (release-derived, i.e. higher-evidence than a bare-name search win — the clobber runs in the WRONG direction), and the race is symmetric and pre-existing in both directions today. Staging sharing prod `entity.identity` multiplies windows.
2. **Silent, provenance-free**: neither writer logs a `reconciliation_log` row on overwrite, so a clobbered id leaves no trail of the previous value.
3. **`unreconciled`-status leak**: rows minted by the resolver INSERT with default `reconciliation_status`, so later reconciler campaigns (`scripts/entity_resolution/__main__.py`) reprocess them through the first-match-wins cascade and can overwrite the API-verified id without any TOCTOU at all.
4. **Case-sibling blindness**: `bulk_resolve_library_names`' LOWER leg returns the lowest-id row (`DISTINCT ON ... ORDER BY id ASC`); a Discogs-id-less `wishy` row shadows an id-bearing `WISHY` sibling, so the resolver re-mints instead of short-circuiting, and the store accretes case-variants with potentially conflicting ids.

## End state

[...]

PR #775, opened 2026-07-13
feat(#759): bulk artist-resolve drain script (PR D)

### LINKED ISSUE #759: Bulk bare-name artist resolution: POST /api/v1/artists/resolve/bulk — API-verified minting with entity.identity write-back
## Problem

Backend-Service needs to resolve clean touring-artist names — artists WXYC has never cataloged — to Discogs artist identities for the concerts pipeline (WXYC/Backend-Service#1614). Measured on prod 2026-07-11: of 395 unresolved upcoming concerts, 252 are clean single-artist names (`Wishy`, `REZN`, `L'Rain`, `glaive`, `The Tubs`, `SiM`, `Popsicle`, …) absent from the library-derived `artists` table and the alias substrate; Backend's strict+alias resolver is fully drained (0 strict, 1 alias), so further recall must come from outside the library-derived corpus.

LML's design intent covers this — match anything, persist what it learns (`release/orchestrator.py` `_writeback_identity` creates `entity.identity` rows on first sight) — but no current HTTP surface composes the existing machinery for a bare-name input: `GET /identity/resolve` and `POST /identity/bulk` are reads of `entity.identity` with no escalation on miss (verified against prod 2026-07-11: all 7 sample touring names return 404; control `Juana Molina` returns 200 with `discogs_artist_id: 187553`). The matching machinery exists but is embedded in other flows (the `DiscogsReconciler` cascade in `scripts/entity_resolution/discogs.py`, the lookup pre-pass, the row-less SONG_AS_ARTIST arm).

### Resolution model: verify before mint

The original framing treated the discogs-cache as a cheaper equivalent of the Discogs API, escalating only on miss. That is wrong for bare touring names: the prod cache is the pair-wise-filtered ~50K build — a **biased sample** of Discogs. The reconciler could trust cache hits because its inputs came *from* the library (the corpus was filtered around exactly those artists); bare touring names void that warranty. A touring "Popsicle" that exact-matches the one cached (different) "Popsicle" would mint a confidently wrong `entity.identity` row — and the COALESCE never-clobber upsert makes a wrong mint **durable and un-self-correcting**. Overloaded (same-named) artist handling is a first-class requirement, and the residual is exactly the short-generic-name population where collisions concentrate.

Therefore: **every write-back-eligible resolution must clear a full-universe uniqueness check against the Discogs API.** Per name:

1. **`entity.identity` read** (`resolve_library_name` three-leg) — true short-circuit, unverified (a pre-existing row is the store's resolved-at-most-once promise; its name-keyed collision hazard is inherent to the store, not this endpoint).
2. **Cache candidate generation** — the reconciler's equality legs (exact / member / alias / name_variation) plus trigram, **rewritten to return candidate sets** (`array_agg`/`GROUP BY`), because the script's `SELECT DISTINCT` + dict-comprehension collapse silently picks an arbitrary candidate on ambiguity — it cannot honor "ambiguous names must not mint." Cache results are evidence (corroboration / conflict detection / telemetry), never verdicts.
3. **Discogs API exact-form check** — `/database/search?type=artist&per_page=100` (single page, no page-2 crawl), results filtered to those whose `to_identity_match_form(title)` equals the input's form. The form already strips trailing parentheticals, so "Popsicle (2)" collides with "Popsicle" — the disambiguator strip is **load-bearing for overload detection**, not cosmetics (`strip_discogs_disambig` is not needed; use the shared normalizer on both sides).

**Verdict table**: exactly 1 exact-form API candidate, cache candidates (if any) agreeing → `resolved`. ≥2 exact-form candidates → `ambiguous` (no auxiliary disambiguation in v1 — no guessing by recency/genre). 1 exact-form candidate but a cache candidate points at a different artist id → `ambiguous` (conflict means doubt, doubt means NULL). 0 exact-form candidates, even when cache alias/member/name-variation legs found something → `not_found` in v1 (alias-shaped matches can't be globally uniqueness-checked; counted in telemetry to size a possible v2 alias arm). 0 candidates anywhere → `not_found`.

**Accepted risk, documented**: a brand-new touring band whose name collides with a single obscure Discogs artist resolves confidently to the wrong artist (count=1; Discogs cannot testify about artists it doesn't contain). No cheap defense; the mitigation is the human spot-check gate in the drain (below) estimating the wrong-mint rate before the live run.

### Contract (specced in `wxyc-shared/api.yaml`; shared PR merges before the LML regen PR)

`POST /api/v1/artists/resolve/bulk`, mounted on `artists/router.py` (uniformly `LML_API_KEY`-authed; model-introspected 413 gate per the `search-aliases/bulk` pattern). Request: `{names: [..], dry_run: false}` with `minItems 1`, `maxItems 25` (a fully-escalating batch ≈ 25 API calls ≈ 30s at the shared 50/min budget; callers page), per-item `maxLength` per the sibling. Single mode — escalation always on; an `escalate=false` PG-only mode is a non-breaking future addition if a consumer materializes.

Per-name verdict, index-aligned with the request, echoing verbatim input as `name`:

- `discogs_artist_id` + `canonical_name` (the **raw** Discogs title, "(N)" suffix included — the caller renders its own raw name and provenance wants the true Discogs string) on success.
- `method: identity_store | api_search` — what *decided* the resolution. Two values only; cache legs never decide (see above).
- `cache_corroboration: string[]` (`cache_exact | cache_member | cache_alias | cache_name_variation | cache_trigram`) on **both** resolved and unresolved verdicts — on resolved, which legs agreed; on unresolved, which legs had candidates v1 declined to mint. This is the per-leg yield + v2-alias-arm sizing telemetry, in-band.
- `unresolved_reason: not_found | ambiguous | escalation_unavailable`. `escalation_unavailable` is the **retryable** verdict: tiers 1–2 missed and the API tier was shed (LML#755 breaker open / outage cool-down / 429 / 5xx-after-retries). On a mid-batch breaker trip, remaining names short-circuit to `escalation_unavailable` immediately (no queueing behind the cool-down). No batch-level 503 — PG tiers keep answering during Discogs outages. Consumers must not apply a no-match TTL to `escalation_unavailable` (it means "couldn't ask," not "asked and missed").
- `candidate_count`: exact-form candidates observed when the API tier ran (1 resolved / ≥2 ambiguous / 0 not_found); **null = not measured** (identity_store short-circuit, escalation_unavailable) — null never means zero.
- `dry_run: true` runs everything identically (all tiers, real API verification, full verdicts) but skips `upsert_identity`. No response-level dry-run echo; no partial-write mode.

### Write-back

Unambiguous resolutions upsert `entity.identity` keyed on the **verbatim input name** (the string future callers will ask about; Discogs-canonical "(N)" keys are unreachable via the three-leg read and pure pollution), populating **`discogs_artist_id` only** (COALESCE never-clobber; other id columns belong to their own enrichment flows). Ambiguous / not_found / escalation_unavailable names mint nothing.

### Processing

Batched PG pre-pass for the whole request (tier-1 reads + cache candidate queries, 1–3 round-trips), then a **serial** API-verification leg — the shared limiter paces globally at 50/min, so intra-request concurrency can't beat it and only starves interleaved live-lookup traffic of semaphore slots (and the #370/#372 no-gather-fan-out lesson stands). Inputs deduped on the identity-match form; each unique form resolved once, the shared verdict fanned back to every input position, and only the **first occurrence's verbatim** mints (two rows for one Discogs id is how the store accretes near-duplicate keys). API calls go through `DiscogsService._request_with_retry` — anything else silently opts out of the limiter/semaphore/breaker.

### Where

- `artists/router.py` — route (auth, 413 gate, disconnect sentinel, Sentry span; thin like siblings).
- `artists/resolver.py` (new) — tier orchestration, verdict assembly, dedupe, write-back, dry_run.
- `discogs/cache_service.py` — new candidate-set queries (the rewritten cascade legs, riding the existing `wxyc_identity_match_artist` functional indexes).
- `discogs/service.py` — new `search_artists` method (`type=artist`; first non-release search in LML).
- `entity/store.py` — existing `resolve_library_name` + `upsert_identity`, unchanged.
- `scripts/entity_resolution/discogs.py` — **untouched**. Its first-match-wins semantics are correct for its library-name inputs; the endpoint's candidate-set SQL is deliberately different. Cross-reference comments in both files so the divergence reads as intentional.

### Observability

Mirror the siblings: explicit `http.server` Sentry span (`POST /api/v1/artists/resolve/bulk`) with `lml.artist_resolve.names / .deduped / .resolved / .not_found / .ambiguous / .escalation_unavailable / .api_calls / .minted / .dry_run` + 499/503/500 status pinning; one aggregate PostHog `artist_resolve_completed` event per request via `_make_posthog_client_dep("artist_resolve")`. No per-name events; no new alerts in v1 (offline consumer; revisit only if this ever serves a hot path). `docs/api-endpoints.md` gets the route + verdict-semantics entry.

### Verified environment facts (2026-07-11, Railway variables)

- Staging LML's `DATABASE_URL_DISCOGS` is the **public proxy of the prod Postgres** (`tramway.proxy.rlwy.net:57291` → same DB prod reaches at `postgres.railway.internal:5432`). A "staging drain" writes production `entity.identity`. **There is no sandbox** — `dry_run` is load-bearing, not optional prudence.
- Staging and prod share the **same `DISCOGS_API_TOKEN`**, with independent uncoordinated 50/min client limiters. A drain through staging while prod serves live traffic can push ~100/min at Discogs's ~60/min ceiling → 429s land on prod → the #755 breaker trips → live lookups shed to cache-only. Therefore the drain runs against the **prod endpoint** (the only place all Discogs traffic coordinates through one limiter/breaker), off-peak, outside the 06:00 UTC flowsheet-backfill window.

## Delivery plan

1. **PR A — wxyc-shared**: api.yaml contract (merges first; LML regen-drift CI requires it on shared main).
2. **PR B — LML primitives**: `DiscogsService.search_artists` + `DiscogsCacheService` candidate-set queries, unit + `pg` tests. Folds into PR C only if trivially small.
3. **PR C — LML endpoint**: regenerated models, `artists/resolver.py`, route, write-back, dry_run, dedupe, serial + breaker short-circuit.
4. **PR D — drain script** in `scripts/` + `docs/scripts.md` entry.

Testing: TDD throughout. Unit: the full verdict table parameterized (incl. mid-batch breaker short-circuit, dedupe/first-verbatim-mint, dry_run, non-clobber). Integration (`pg`, sibling `app_client` pattern — this is the full-chain tier; no new `tests/e2e/` entry, the e2e conftest has no PG): tier-1 short-circuit asserting zero API calls; the AC-as-test two-route flow (resolve then `GET /identity/resolve` returns the minted row); dry_run inverse (subsequent read 404s); seeded overload family → `ambiguous`/`candidate_count=2`/no mint; cache-conflict → `ambiguous`; non-clobber (pre-existing `spotify_artist_id` survives); auth 401 + 413; one normalization-parity canary (diacritic + article + "(N)" name through both the PG leg and the Python filter — parity itself is locked upstream in wxyc-etl, the canary just makes a break surface here). One `external_api`-marked live smoke pinning the `type=artist` response-shape assumptions against a stable overload family.

**Drain** (the WXYC/Backend-Service#1614 measurement gate): BS owns the name-set export (the clean-name predicate must stay the same code as `extractHeadliner`'s gate — it lives BS-side); handoff is a names file. The LML script pages 25 at a time against prod, `--dry-run` default / `--live` explicit, JSONL append + resume (a crash at batch 8 must not re-pay 175 API calls), `escalation_unavailable` re-paged after a cool-down with bounded retries (2–3) then reported as residual. Report posted to WXYC/Backend-Service#1614: per-method yield, corroboration + candidate-count distributions, ambiguity rate, alias-leg-only + trigram-near-miss counts, residual — **plus a ~20-name random spot-check table with `discogs.com/artist/<id>` links**. A human eyeballs the spot-check for wrong-mints before authorizing `--live`; live-drain authorization is explicitly human. Sequence: dry drain → report + spot-check → live drain → BS writer unblocks.

## Acceptance criteria

- [ ] A name absent from `entity.identity` but exact-form-unique in Discogs resolves to `discogs_artist_id` + raw canonical name + `method`, with `cache_corroboration` populated when cache legs agreed.
- [ ] Unambiguous resolution persists to `entity.identity` keyed on verbatim input, `discogs_artist_id` only (visible to a subsequent `GET /identity/resolve`); ambiguous/not_found/escalation_unavailable names persist nothing; `dry_run` persists nothing while returning full verdicts.
- [ ] Overloaded names (multiple exact-form Discogs artists, incl. "(N)" families) return `ambiguous` with `candidate_count` — never a guess; cache-vs-API conflicts return `ambiguous`.
- [ ] `escalation_unavailable` is distinguishable from `not_found` (breaker open / outage / 429 / 5xx-after-retries), and a mid-batch breaker trip short-circuits remaining names fast.
- [ ] Contract specced in `wxyc-shared/api.yaml`; auth + 413 cap mirror the sibling bulk endpoints; 25-name cap.
- [ ] Test pyramid above (unit verdict table, `pg` integration incl. the two-route AC test and parity canary, `external_api` smoke).
- [ ] Dry-run drain over the WXYC/Backend-Service#1614 name set (~252 names) run against prod off-peak, report + spot-check table posted there before its writer lands.

## Related

- WXYC/Backend-Service#1614 — consumer (concerts headliner recall); blocked by this issue (native dependency wired).
- WXYC/Backend-Service#1588 — touring-events epic.
- #760 — follow-up: endpoint-path clarity/consistency audit (the `/identity/resolve` release-vs-artist collision this design had to route around).
- #631 / #625 — row-less non-library resolution precedent. #479 — sibling bulk-endpoint conventions. #207 — reconciliation-pipeline legs. #755 — the Discogs saturation breaker the escalation tier rides.

jakebromberg:
PR D of #759 — the drain script, the last deliverable. PR A (WXYC/wxyc-shared#218/#219), PR B (#761/#762), and PR C (resolver core #765 + endpoint wiring #764) are all merged, and the four review-loop follow-ups (#770/#772/#773/#774) landed on top. This adds the tool that runs the endpoint over the real Backend-Service name set and posts the measurement report that gates the live mint.

## What

`scripts/artist_resolve_drain/` — a CLI that drains a bare-name set through prod `POST /api/v1/artists/resolve/bulk`, minting `entity.identity` rows for the exact-form-unique touring names and producing a yield report + a wrong-mint spot-check table. Backend-Service owns the name-set export (the clean-name predicate is the same code as its `extractHeadliner` gate); the handoff is a names file — JSON array or newline-delimited.

- **`drain.py`** — the network- and clock-free core: `parse_names_file`, `compute_pending` (the resume + retry brain), `resolve_batch` (HTTP envelope with length + index-alignment guards), `make_post_batch` (transient-HTTP retry, 4xx left to fail loudly), and `run_drain` (the loop). `run_drain` takes an injected `post_batch`/`sleep`/`clock`, so resume, retry, and paging are unit-tested without touching Discogs or waiting out a cool-down.
- **`report.py`** — `build_report` (per-method yield, cache-corroboration + candidate-count distributions, ambiguity rate, alias-leg-only + trigram-near-miss counts for v2-alias-arm sizing, `escalation_unavailable` residual), `sample_spot_check` (seeded, reproducible-from-the-JSONL, `api_search`-mints-only), `format_report_markdown`.
- **`__main__.py`** — the CLI + runbook.

## Safety design (per the #759 drain section)

- **Prod endpoint only.** It is the single place all Discogs traffic coordinates through one 50/min limiter + the #755 breaker; a staging drain (shared token, uncoordinated limiter) could 429 live lookups. Run off-peak, outside the 06:00 UTC backfill window.
- **Dry run is the default; `--live` is explicit** — a mint is durable (COALESCE never-clobber), so a wrong mint is un-self-correcting. The runbook is dry drain → human eyeballs the spot-check table for wrong mints → authorize `--live`.
- **Crash-safe, mode-scoped resume.** Every verdict is appended to a JSONL log; a restart re-pages nothing already settled, and because resume/reporting are scoped by `dry_run`, a prior dry drain never makes a `--live` run skip the mint.
- **Bounded escalation retry.** `escalation_unavailable` (breaker open / outage / 429 / 5xx-after-retries) is the one retryable verdict — re-paged after a cool-down up to `--max-retries` (default 2), then reported as residual.

## Testing

TDD throughout. 46 unit tests (`tests/unit/test_artist_resolve_drain.py`) cover parsing, `compute_pending` resume/retry/mode-filter, the JSONL round-trip, `resolve_batch` (auth header, `dry_run` body, length + index-misalignment guards, HTTP error) via `pytest-httpx`, the full `run_drain` loop with an injected fake post + fake sleep (resume-skip, escalation re-page with cool-down then residual, dry-vs-live, shutdown), and the report + spot-check aggregation. Local gate green: `ruff check` / `ruff format --check` / `mypy` (fully typed — not in the relaxed-scripts override) / full unit suite. No `pg` tests — the drain is a pure HTTP client with no DB access of its own.

## Docs

`docs/scripts.md` gains a section with the dry → spot-check → live runbook; the CLAUDE.md `docs/scripts.md` router line is updated.

Closes #759. Unblocks WXYC/Backend-Service#1614 (the concerts headliner writer, which waits on this drain's report + spot-check before it lands).

## Related

- #759 (parent) — PR A wxyc-shared#218/#219, PR B #761/#762, PR C #765 (resolver core) + #764 (endpoint), follow-ups #766/#767/#768/#769 → #770/#772/#773/#774
- WXYC/Backend-Service#1614 — consumer, unblocked by this drain
- #755 — the Discogs saturation breaker the `escalation_unavailable` retry rides


jakebromberg:
Splitting this into two stacked, independently-reviewable PRs to stay under the ~1000-line target:

- **#776 (PR D1)** — the drain engine (`drain.py` + runnable CLI + engine tests), branched from `main`.
- **#777 (PR D2)** — the yield report + wrong-mint spot-check (`report.py` + report CLI wiring + report tests + runbook docs), stacked on D1. **Closes #759.**

The reviewed content here (including the two `report.py` metric fixes and the full-typing pass from the converged review loop) carries forward verbatim into the split; D1 additionally gains direct `make_post_batch` retry-semantics tests flagged by the pre-PR review. Closing this in favor of #776 → #777.

Issue #801, opened 2026-07-16
Artist+track lookup returns 5 unrelated albums when the track-bearing album is in the library (Milkman / Aphex Twin → Richard D. James Album)


jakebromberg:
## Symptom

A Slack request `Milkman Aphex Twin` — where "Milkman" is a track on *Richard D. James Album* — returned five unrelated Aphex Twin releases and never surfaced the album that actually contains the track, even though that album is in the WXYC library at **Ap 2/8** (library id 87). This is the third production instance of the class already reported and closed as fixed in #267 (`bucky skank` / Lee Perry → *Live at Maritime Hall*) and #560 (`Tenant` / Siouxsie and the Banshees → *Kaleidoscope*). The class is not fully closed.

The five returned rows are exactly the first five Aphex Twin releases by catalog number (Ap 2/1–2/5). *Richard D. James Album* is Ap 2/8 — the 8th of 18 Aphex Twin releases in the catalog — so it never enters the returned set.

## Library state (confirmed against library.db)

```
$ sqlite3 library.db "SELECT id, title, release_call_number FROM library WHERE artist LIKE '%Aphex%' ORDER BY release_call_number"
80|Selected Ambient works vol 2|1
81|I Care Because You Do|2
82|Raising The Titanic|3
83|Donkey Rhubarb 12"|4
84|Classics|5
85|Polyson Window|6
86|Hangable Auto Bulb 12"|7
87|Richard D. James Album|8      <-- contains "Milkman"; never returned
88|Come To Daddy|9
... (18 rows total)
```

## Telemetry (Request-O-Matic PostHog, project 293013)

The matching `lookup_completed` event fired at `2026-07-15 23:08 UTC` (≈1 min after the 7:07 PM EDT Slack card): `had_artist=true, had_song=true, had_album=false, search_type=compilation, results_count=3, total_duration_ms ≈ 10–14k`. Two takeaways:

- **The upstream parse was correct** — request-o-matic emitted `song="Milkman"`, `artist="Aphex Twin"`, `album=None` (the only rows in the surrounding window with `had_album=false`). This is not a request-o-matic parsing bug.
- **The 10–14s wall time places the request inside the 2026-07-15 flowsheet-metadata-backfill flood.** The surrounding `lookup_completed` stream is the uniform ~5s backfill churn with round-number timeout-backoff durations (10s/20s/30s). The discogs-cache PG pool is degraded under that flood (see #706).

## Why it falls through

[...]

Issue #706, opened 2026-06-28
Cold /api/v1/lookup tail latency regressed (~Jun 25): p95 25s / p99 30s — inline streaming post-process + event-loop starvation


jakebromberg:
## Problem

Cold-cache `POST /api/v1/lookup` tail latency has regressed sharply since ~2026-06-25 and is now far above both the [Epic A](https://github.com/WXYC/library-metadata-lookup/issues/338) target (cold p99 < 8 s) and the ~18.9 s p99 baseline that #651's "Separate finding" flagged for follow-up.

This is the dedicated cold-tail issue that #651 explicitly asked for ("the tail... should be its own issue if not already covered"). It is distinct from #651 (the Bandcamp **median** regression, resolved by keeping that leg off) and from #269 (a prior cold-tail issue whose dominant cause was trigram queries — no longer the dominant factor).

### Measured impact — Sentry, prod `/api/v1/lookup` (`is_transaction:true`)

| metric | value |
|---|---|
| count (3 d) | 14,597 |
| p50 | 4.7 ms (warm cache hit) |
| p95 | 25.4 s |
| p99 | ~30.6 s |
| max | 310 s |

`/api/v1/lookup` is ~98% of all LML transaction time. The distribution is sharply bimodal: warm hits are instant; cold misses fall into a 15–30 s tail.

### Onset — external corroboration (wxyc-canary)

The `wxyc-canary` `lml-auth` probe (a fixed Juana Molina / DOGA lookup every 5 min) was green for 3 weeks, then:

| day | canary lml-auth avg latency | max | failing 5-min datapoints |
|---|---|---|---|
| Jun 19–24 | ~730 ms | 1.6–7.1 s | 0 |
| Jun 25 | 2,455 ms | 13.5 s | 26 |
| Jun 26 | 5,727 ms | 13.3 s | 53 |

Clean step-change on Jun 25, escalating day over day. (The canary check itself has been narrowed to abstain on this timeout rather than page — `wxyc-canary` branch `fix/lml-auth-timeout-non-paging` — so this issue is the real-signal follow-up.)

### What's slow on the cold path — Sentry span breakdown (`transaction:/api/v1/lookup span.duration:>5000`, 3 d)

| span | count | p95 | max |
|---|---|---|---|
| `GET api.music.apple.com/.../search` | 1,210 | 10.2 s | 25.4 s |
| `apple_music.search` albums / songs | 1,220 | 10.3 s | 20.4 s |
[...]

Issue #333, opened 2026-05-14
validate_track_on_release should filter `release_track_artist` to `extra = 0`


jakebromberg:
## Problem

`discogs/cache_service.py:validate_track_on_release` currently treats every row in `release_track_artist` as a candidate per-track credit. After WXYC/discogs-etl#221 + WXYC/discogs-xml-converter#55 (merged 2026-05-14), that table has a new `extra integer DEFAULT 0` column that distinguishes main-artist credits (`extra = 0`) from extra-artist credits — writer, producer, performer, etc. (`extra = 1`).

LML#328 added a release-level-artist fallback to work around the cache over-reporting per-track credits; that fallback is now load-bearing on every per-track substring miss, including legitimate main-artist matches that just happen to also have an extra-credit row for the same person. Filtering the per-track read to `WHERE extra = 0` collapses the fallback path to defense-in-depth and tightens precision on actual VA compilation tracks, where per-track main credits ARE the source of truth.

## Desired end state

`validate_track_on_release` in the cache path (and the cache's `search_releases_by_track` SQL, if it joins `release_track_artist` for the per-track-artist filter) reads only main-artist credits by default. The release-level fallback added in LML#328 stays in place but is reached far less often.

## Where

- `discogs/cache_service.py:validate_track_on_release` — the `release_track_artist` fetch on ~line 965-968 of post-#328 main:
  ```python
  self.pool.fetch(
      "SELECT track_sequence, artist_name FROM release_track_artist WHERE release_id = $1",
      release_id,
  )
  ```
  Add `AND extra = 0` to the WHERE.

- `discogs/cache_service.py:search_releases_by_track` — the LEFT JOIN on `release_track_artist rta` in the trigram-similarity query (~line 105-107 of pre-#328 main; check current shape). The JOIN condition itself or the WHERE clause needs the `rta.extra = 0` constraint so VA-compilation matches don't surface a release just because a producer's name happens to look like the requested artist.

## Constraints

- **Gated on the cache having `extra` populated.** Railway prod cache won't have `extra > 0` rows until the next monthly rebuild — 2026-06-04 06:00 UTC. WXYC/discogs-etl#222 must be fixed before that tick, or the rebuild will fail at the `release_track_artist` COPY. **Don't merge this PR until the June 4 rebuild has run cleanly** and `psql $DATABASE_URL_DISCOGS -c "SELECT COUNT(*) FROM release_track_artist WHERE extra = 1;"` returns a nonzero count.
[...]

Issue #267, opened 2026-05-09
Artist-fallback returns unrelated albums when local cache holds a song-bearing match


jakebromberg:
## Symptom

`/api/v1/lookup` for `bucky skank by lee scratch perry` returns 5 unrelated Lee Perry albums plus `song_not_found=true, search_type=compilation` — even though the local Discogs PG cache holds the answer (`Lee 'Scratch' Perry — Live at Maritime Hall`, track 8 = `Bucky Skank`, WXYC library id 12682).

## Reproducer

```
$ lookup -s "bucky skank by lee scratch perry"
...
  Strategy:  compilation
  Match:     song not found -- showing other albums by artist (5)

  [1] Arkology
  [2] Dub Fire
  [3] Chicken Scratch
  [4] Ooh! Wah! 12"
  [5] History Mystery Prophecy
```

The library *does* contain `Live at Maritime Hall` (id 12682). Cross-checked against the homebrew Discogs cache (`:5432`):

```sql
SELECT r.title, rt.position
FROM release r
JOIN release_track rt ON rt.release_id = r.id
JOIN release_artist ra ON ra.release_id = r.id
JOIN artist a ON a.id = ra.artist_id
WHERE rt.title ILIKE 'bucky%skank%' AND a.name ILIKE '%perry%';
-- → 'Lee \"Scratch\" Perry Live At Maritime Hall' / position 8 (and 18 other releases)
```

## Why it falls through

Tracing the pipeline (`lookup/orchestrator.py`):

1. `resolve_albums_for_track` → `lookup_releases_by_track` queries Discogs (cache then API), then re-validates each release via `validate_track_on_release`. In production this validation rejects results the cache *just* told us contained the track (cache miss for that release's tracklist, API rate-limited, etc.) — `albums_for_search` ends up empty.
2. `search_library_with_fallback` falls through artist+song (no library row has \"Bucky Skank\" in its title) to artist-only FTS5, which returns the first 5 Lee Perry albums by primary-key order. Maritime Hall (id 12682) is the 21st row by id and never makes the cut.
3. `search_compilations_for_track` runs the same `search_releases_by_track` again (with `artist_as_keyword`), validates per release, and ends up with 0 matches.
4. `filter_results_by_track_validation` runs over the 5 fallback rows, can't confirm any, returns `None`. The orchestrator keeps the unfiltered 5 and ships them.

[...]

PR #807, opened 2026-07-16
fix(#802): prefilter search_releases_by_track by artist inside the CTE

### LINKED ISSUE #802: search_releases_by_track: matching_tracks CTE truncates by track similarity before the artist filter, pruning the artist's own release for common track titles
## Problem

`DiscogsCacheService.search_releases_by_track` builds a `matching_tracks` CTE that selects releases whose track title trigram-matches the query, orders by similarity, and truncates with `LIMIT $2` (where `$2 = limit*2`). The **artist filter is applied only in the outer query, after the CTE has already truncated**:

```sql
WITH matching_tracks AS (
    SELECT DISTINCT rt.release_id, rt.sequence, rt.title as track_title,
           similarity(lower(f_unaccent(rt.title)), lower(f_unaccent($1))) as sim
    FROM release_track rt
    WHERE lower(f_unaccent(rt.title)) % lower(f_unaccent($1))
    ORDER BY sim DESC
    LIMIT $2                              -- $2 = limit*2; NO artist filter here
)
SELECT r.id, r.title, ra.artist_name, ...
FROM matching_tracks mt
JOIN release r ON r.id = mt.release_id
JOIN release_artist ra ON ra.release_id = r.id AND ra.extra = 0
LEFT JOIN release_track_artist rta ON ... AND rta.extra = 0
WHERE ($3::text IS NULL
       OR lower(f_unaccent(ra.artist_name)) % lower(f_unaccent($3))
       OR lower(f_unaccent(rta.artist_name)) % lower(f_unaccent($3)))   -- artist filter AFTER the LIMIT
ORDER BY mt.sim DESC
```

`rows = await self.pool.fetch(query, track, limit * 2, artist)` — the artist (`$3`) is never referenced inside the CTE.

For a **common track title**, many releases carry an exact or near-exact match and all tie at `sim = 1.0`; PostgreSQL's `ORDER BY sim DESC LIMIT 40` then keeps an arbitrary 40 of them. If the artist's own release is not in that arbitrary top-40, it is pruned **before** the `ra.artist_name % $3 OR rta.artist_name % $3` filter runs, and the query returns zero rows for that artist even though the release is in the cache.

This is the shared upstream failure behind the artist+track recall miss in #801 (`Milkman` / Aphex Twin → *Richard D. James Album*), and it sits in the same method and layer as #237.

## Where

- [`discogs/cache_service.py` — `search_releases_by_track`](https://github.com/WXYC/library-metadata-lookup/blob/1ce136bd8b1fccc1c278648a7e94e5abcffcc99e/discogs/cache_service.py#L357-L446), specifically the [`matching_tracks` CTE](https://github.com/WXYC/library-metadata-lookup/blob/1ce136bd8b1fccc1c278648a7e94e5abcffcc99e/discogs/cache_service.py#L388-L416) and the `pool.fetch(query, track, limit * 2, artist)` call at [L416](https://github.com/WXYC/library-metadata-lookup/blob/1ce136bd8b1fccc1c278648a7e94e5abcffcc99e/discogs/cache_service.py#L416).

## Desired end state

When `artist` is provided and the cache holds a release by that artist containing a track matching `track`, the release is returned regardless of how many other-artist releases share the track title.

## Suggested approach

Push the artist predicate into the candidate-selection stage so the `LIMIT` is applied to the already-artist-filtered set, not the raw track-title match set. The artist columns live on `release_artist` / `release_track_artist` (not `release_track`), so this means joining those into the CTE (or a second CTE) before the `LIMIT` when `$3 IS NOT NULL`. Preserve:

- the `ra.extra = 0` / `rta.extra = 0` main-artist-credit semantics and the documented #333 recall trade-off;
- the `sim DESC` ordering and the `limit*2` → dedup-to-`limit` shape;
- the **`artist IS NULL` path unchanged** — VA / SONG_AS_TRACK runs with `artist=None`, so the CTE must still truncate purely by similarity when no artist is supplied.

Watch: the artist predicate has two legs (`ra` release-level and `rta` track-level featured-credit). Both must survive the move so a featured-credit release still matches.

## Constraints

- No schema change; read-only query.
- Keep query rationale in Python comments, not `--` SQL comments (per the existing note in the method — keeps it out of `pg_stat_statements` / PG logs).
- Cache-only method; do not add an API leg here.

## Acceptance criteria

- [ ] Unit/integration test against a seeded discogs-cache: N+1 releases share a track title where only one is by the queried artist and N > `limit*2`; `search_releases_by_track(track, artist, limit)` returns the artist's release. Fails on current code (prune), passes after.
- [ ] `artist=None` path returns the same top-`limit` by similarity as today (no regression).
- [ ] Featured-credit (`rta.extra = 0`) match path still works.
- [ ] The #801 Milkman / Aphex Twin lookup returns *Richard D. James Album* against a healthy cache.

## Related

- Parent: WXYC/library-metadata-lookup#801
- WXYC/library-metadata-lookup#237 — sibling recall gap in the same method (canonical-artist trios)
- WXYC/library-metadata-lookup#543 — fallback-cascade bounding (same method's API-call cost)
- WXYC/library-metadata-lookup#333 — the `extra = 0` recall trade-off referenced in the code

jakebromberg:
## Problem

`DiscogsCacheService.search_releases_by_track`'s `matching_tracks` CTE ranked releases by track-title trigram similarity and truncated with `ORDER BY sim DESC LIMIT limit*2`, but the artist bind (`$3`) was referenced only in the **outer** `WHERE` — so the `LIMIT` ran before the artist filter. For a common track title, dozens of releases tie at `sim = 1.0` and an arbitrary `limit*2` survive truncation; if the queried artist's own release wasn't in that set it was pruned before the artist predicate ever ran, and the query returned **zero rows for a release that is in the cache**. This is the upstream cause of the #801 "Milkman" / Aphex Twin → *Richard D. James Album* recall miss, and it starves both the Step-2 compilation probe and the #267 cache-only rescue.

## Fix

Split the single query into two module-level SQL strings selected on whether an artist was supplied:

- **`_SEARCH_BY_TRACK_SQL` (artist absent)** — the VA / SONG_AS_TRACK hot path (under #706 perf scrutiny). Structurally unchanged, so its parse tree, `pg_stat_statements` entry, and generic plan stay frozen. Two strings, not one `$3 IS NULL OR …` guard, precisely so the planner doesn't build one generic plan spanning both branches.
- **`_SEARCH_BY_TRACK_ARTIST_SQL` (artist supplied)** — the artist predicate is pushed **into the CTE `WHERE`, before the `LIMIT`**, via two correlated `EXISTS` legs (release-level `release_artist` + track-level `release_track_artist`, the latter correlated to the matched track's `sequence`). `EXISTS` (not `JOIN`) so the CTE can't multiply `release_track` rows — the `SELECT DISTINCT` / `LIMIT` still counts distinct candidate tracks. The outer `SELECT`/JOIN/`WHERE` is kept as a pure subtractive backstop, preserving the existing display-artist / `is_compilation` selection.

`rta.extra = 0` main-artist-credit semantics and the #333 recall trade-off are unchanged; query rationale lives in Python comments (out of `pg_stat_statements` / PG logs).

## Tests

Six `pg`-marked integration reproductions (release-level prefilter, `artist=None` no-regression, track-level featured credit, `extra=1` precision guard, the #801 Milkman/Aphex pin, and a trigram-floor boundary pin) + three unit tests (frozen hot-path dispatch, a whitespace-normalized snapshot change-detector for the None-branch SQL, and a structural pin that the artist `EXISTS` leg lands inside the CTE before `LIMIT`).

**Bug-fix-protocol note:** the prune is a SQL-execution artifact that can't reproduce against a mock pool, so the behavioral repros are the `pg` tests and the structural unit test is the protocol analog. RED was re-confirmed by restoring the pre-fix `cache_service.py`: the 3 behavioral repros fail returning `[]` (the prune), and the Milkman/Aphex pin returns `[]` on old code vs. the target release id with the fix.

## Verification

Local CI green: `ruff check .`, `ruff format --check .`, `mypy . --ignore-missing-imports` (no issues, 399 files), full non-pg suite (3931 passed), and `pytest -m pg` (all 6 new tests + the existing cache_service SQL-substring tests pass).

**Pre-merge gate — satisfied** (full numbers in [this comment](https://github.com/WXYC/library-metadata-lookup/pull/807#issuecomment-4988269885)): `EXPLAIN (ANALYZE, BUFFERS)` on the prod discogs-cache confirms all three trigram GIN indexes exist (`idx_release_track_title_trgm`, `idx_release_artist_name_trgm`, `idx_release_track_artist_name_trgm`) and the artist legs ride them as hashed SubPlans. Realistic `Milkman` + `Aphex Twin` = 264 ms; pathological `Intro` (3,629 matches) = 1.3 s — both dominated by the `title % $1` bitmap-heap-recheck that today's query pays identically, the fix adding only ~25–95 ms of index-driven hashed SubPlan work. No index gap, no regression vs. today's plan for the common-title case.

## Scope

Necessary-but-not-sufficient for the parent class #801 (the end-to-end assertion is a healthy-cache orchestrator replay, tracked on #801). Distinct from #237's trigram-vs-`token_set` mechanism.

Closes #802

---

## Performance analysis

**`artist=None` hot path (VA / SONG_AS_TRACK): zero change.** `_SEARCH_BY_TRACK_SQL` is structurally identical to the pre-split query — same parse tree, plan, and `pg_stat_statements` entry. This is the path the backfill flood mostly exercises, so the #706-sensitive surface is untouched.

**artist-supplied path: neutral-to-slightly-positive, gated on one index.** The artist predicate moves from post-`LIMIT` (40 rows, outer `WHERE`) to pre-`LIMIT` (the full title-match candidate set, inside the CTE via two `EXISTS` legs). `EXPLAIN (ANALYZE, BUFFERS)` on a synthetic 20k-release set (2,000 sharing a common "Milkman" title + one Aphex target; warm cache — plan shape transfers to prod, absolute ms does not):

| Query variant | Result | Exec time | Why |
|---|---|---|---|
| OLD (artist in outer `WHERE`) | **0 rows** (the bug) | 2.99 ms | prunes the artist's release before the filter |
| NEW, **no** `artist_name` trgm index | 1 row (correct) | 18.9 ms (~6×) | seq-scans `release_artist` to evaluate the predicate |
| NEW, **with** `artist_name` trgm index | 1 row (correct) | **2.14 ms** (≈ old, a hair faster) | index probe resolves the `EXISTS` cheaply |

The whole outcome hinges on the `gin_trgm_ops` index on `release_artist.artist_name` (and `release_track_artist.artist_name`). Both are defined in the discogs-etl schema that builds the prod cache — `discogs-etl/schema/create_indexes.sql:17` (`idx_release_artist_name_trgm`) and `create_track_indexes.sql:51` (`idx_release_track_artist_name_trgm`) — so prod is the "with index" row: **perf-neutral-to-positive on the artist branch, zero on the hot path**, while fixing the correctness bug.

**Outstanding verification (was not recorded pre-merge).** Those indexes use `CREATE INDEX CONCURRENTLY IF NOT EXISTS`, which can leave an `INVALID` index if a build/rebuild was interrupted (see the cache-rebuild history behind #706). If either is missing or `indisvalid = false`, the artist branch degrades to a full `release_artist` seq-scan per common-title lookup — far worse than the synthetic 6× under buffer-thrash, competing for the already-starved pool. Confirm on the prod cache:

```sql
SELECT indexrelid::regclass, indisvalid FROM pg_index
WHERE indexrelid::regclass::text
  IN ('idx_release_artist_name_trgm','idx_release_track_artist_name_trgm');
```

Valid → no action (perf-neutral). Missing/invalid → reindex in **discogs-etl** (schema-owned); do not `CREATE INDEX` from LML. Even a no-index regression lands on **organic** artist+track traffic (the #267 rescue, TRACK_ON_COMPILATION artist probe), not the backfill flood (which is artist+album/VA → hits the unchanged hot path).

### DIFF
--- a/discogs/cache_service.py
+++ b/discogs/cache_service.py
@@ -224,6 +224,101 @@ _ARTIST_TRIGRAM_CANDIDATE_THRESHOLD = 0.85
 _PG_TRGM_FLOOR = 0.3
 _SET_PG_TRGM_FLOOR_SQL = f"SET LOCAL pg_trgm.similarity_threshold = {_PG_TRGM_FLOOR}"
 
+# ``search_releases_by_track`` picks one of these two strings on whether an
+# artist was supplied. Two strings, not one ``$3 IS NULL OR ...`` guard, so the
+# artist=None path (the VA / SONG_AS_TRACK hot path, under #706 perf scrutiny)
+# keeps the exact parse tree it runs today — same pg_stat_statements entry, same
+# generic plan. A single guarded string would force the planner to build one
+# generic plan spanning both branches (it can't constant-fold the unknown ``$3``
+# bind at plan time), so the common None path could inherit a worse plan than
+# today's. See LML#802.
+#
+# ``rta.extra = 0`` constrains the credit legs to main-artist credits (mirrors
+# validate_track_on_release after #333). Neither string has a Python-side
+# release-level fuzz fallback: a release whose only matching credit is
+# ``extra = 1`` (featuring guest, composer, remixer) drops from the candidate
+# ranking; the fallthrough seam's API leg surfaces it at the cost of Discogs
+# quota. See #333 for the recall trade-off; #472-#475 track the parallel filters
+# in sibling read sites. Keep this rationale here (not in ``--`` comments inside
+# the SQL) so it stays out of pg_stat_statements and PG logs.
+_SEARCH_BY_TRACK_SQL = """\
+WITH matching_tracks AS (
+    SELECT DISTINCT rt.release_id, rt.sequence,
+           rt.title as track_title,
+           similarity(lower(f_unaccent(rt.title)), lower(f_unaccent($1))) as sim
+    FROM release_track rt
+    WHERE lower(f_unaccent(rt.title)) % lower(f_unaccent($1))
+    ORDER BY sim DESC
+    LIMIT $2
+)
+SELECT r.id as release_id, r.title, ra.artist_name,
+       mt.track_title,
+       CASE WHEN lower(ra.artist_name) LIKE '%various%' THEN true ELSE false END as is_compilation
+FROM matching_tracks mt
+JOIN release r ON r.id = mt.release_id
+JOIN release_artist ra ON ra.release_id = r.id AND ra.extra = 0
+LEFT JOIN release_track_artist rta
+    ON rta.release_id = mt.release_id
+    AND rta.track_sequence = mt.sequence
+    AND rta.extra = 0
+WHERE (
+    $3::text IS NULL
+    OR lower(f_unaccent(ra.artist_name)) % lower(f_unaccent($3))
+    OR lower(f_unaccent(rta.artist_name)) % lower(f_unaccent($3))
+)
+ORDER BY mt.sim DESC\
+"""
+
+# Artist-supplied variant: the artist predicate is pushed INTO the CTE ``WHERE``
+# (before the ``LIMIT``), via two correlated ``EXISTS`` legs, so the sim-ordered
+# ``LIMIT`` truncates the already-artist-filtered set instead of pruning the
+# artist's own release out from under it (LML#802 / #801). ``EXISTS`` not
+# ``JOIN`` so the CTE can't multiply ``release_track`` rows -- the ``SELECT
+# DISTINCT`` / ``LIMIT`` still counts distinct candidate tracks. The ``rta`` leg
+# is correlated to ``rt.sequence`` so a release credited to the artist only as a
+# track-level featured credit still matches on its matching track. The outer
+# ``SELECT`` / JOIN / ``WHERE`` are kept identical to ``_SEARCH_BY_TRACK_SQL``:
+# redundant with the CTE here (a pure subtractive backstop that can only remove
+# rows the CTE kept, never admit a wrong one) but it preserves the existing
+# display-artist / is_compilation selection for multi-credit releases.
+_SEARCH_BY_TRACK_ARTIST_SQL = """\
+WITH matching_tracks AS (
+    SELECT DISTINCT rt.release_id, rt.sequence,
+           rt.title as track_title,
+           similarity(lower(f_unaccent(rt.title)), lower(f_unaccent($1))) as sim
+    FROM release_track rt
+    WHERE lower(f_unaccent(rt.title)) % lower(f_unaccent($1))
+      AND (
+          EXISTS (SELECT 1 FROM release_artist ra
+                  WHERE ra.release_id = rt.release_id AND ra.extra = 0
+                    AND lower(f_unaccent(ra.artist_name)) % lower(f_unaccent($3)))
+          OR
+          EXISTS (SELECT 1 FROM release_track_artist rta
+                  WHERE rta.release_id = rt.release_id
+                    AND rta.track_sequence = rt.sequence AND rta.extra = 0
+                    AND lower(f_unaccent(rta.artist_name)) % lower(f_unaccent($3)))
+      )
+    ORDER BY sim DESC
+    LIMIT $2
+)
+SELECT r.id as release_id, r.title, ra.artist_name,
+       mt.track_title,
+       CASE WHEN lower(ra.artist_name) LIKE '%various%' THEN true ELSE false END as is_compilation
+FROM matching_tracks mt
+JOIN release r ON r.id = mt.release_id
+JOIN release_artist ra ON ra.release_id = r.id AND ra.extra = 0
+LEFT JOIN release_track_artist rta
+    ON rta.release_id = mt.release_id
+    AND rta.track_sequence = mt.sequence
+    AND rta.extra = 0
+WHERE (
+    $3::text IS NULL
+    OR lower(f_unaccent(ra.artist_name)) % lower(f_unaccent($3))
+    OR lower(f_unaccent(rta.artist_name)) % lower(f_unaccent($3))
+)
+ORDER BY mt.sim DESC\
+"""
+
 
 def _negative_cache_key_hash(artist: str | None, track: str, artist_as_keyword: bool) -> bytes:
     """Hash the (artist, track, artist_as_keyword) tuple into a stable 32-byte key.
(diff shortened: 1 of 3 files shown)


jakebromberg:
## Pre-merge EXPLAIN gate — satisfied ✅

Ran `EXPLAIN (ANALYZE, BUFFERS)` for `_SEARCH_BY_TRACK_ARTIST_SQL` against the **production** discogs-cache (read-only).

**Trigram indexes — all three exist and match the predicates exactly** (`USING gin (lower(f_unaccent(...)) gin_trgm_ops)`):

| column | index |
|---|---|
| `release_track.title` | `idx_release_track_title_trgm` |
| `release_artist.artist_name` | `idx_release_artist_name_trgm` |
| `release_track_artist.artist_name` | `idx_release_track_artist_name_trgm` |

No index gap → nothing to file to discogs-etl.

**Plan shape:** the planner leads with the `title % $1` bitmap index scan, then applies the two artist legs as **hashed SubPlans** (each computed once from its own trigram index — 42 `release_artist` rows / 164 `release_track_artist` rows for "Aphex Twin") probed against the title-match set during the bitmap heap scan. No per-row artist re-evaluation, no unindexed scan.

**Latency:**

| case | title matches | execution time |
|---|---|---|
| realistic — `Milkman` + `Aphex Twin` | 143 | **264 ms** (returns the 3 Aphex releases) |
| pathological — `Intro` + `Aphex Twin` | 3,629 | **1.3 s** (incl. a one-time 71 ms JIT compile the cost estimate triggers) |

Both are dominated by the `title % $1` bitmap-index-scan + heap-recheck that **today's query pays identically** (today also scans the full match set to sort it before `LIMIT`). The fix adds only the two index-driven hashed SubPlans — ~25 ms (realistic) / ~95 ms (worst case), computed once. **No regression vs. today's plan for the common-title case**, and this is the organic artist+track branch, not the VA / SONG_AS_TRACK hot path (which runs the unchanged `_SEARCH_BY_TRACK_SQL`).

CI is green on this branch, including the PG Tests job (the six new `pg` reproductions) and Type Check. The merge gate this PR's description called out is now closed.

Issue #850, opened 2026-07-19
feat(lookup): pin verified library-release→Discogs-release overrides (Alex L. catalog links, Phase 1)


jakebromberg:
## Problem

When LML resolves a **library album** (a query matching a real `library.db` row) via `POST /api/v1/lookup`, it chooses the Discogs release — and therefore the tracklist — by a **per-request fuzzy match** against the Discogs cache. There is no durable "this library release → this Discogs release" association anywhere in the system, so the pick can be wrong (wrong pressing, wrong master, a sibling release), which surfaces as wrong/empty tracklists. This is the same failure class as LML#506 and LML#560, and is squarely in scope for the Post-launch service hardening initiative (Project #32).

We now have a **hand-verified** correction for a large slice of the catalog: WXYC DJ Alex L. went through the card catalog one release at a time and recorded the correct Discogs link for each — **62,855 releases, 61,720 (98%) linked**, of which **12,218 are Discogs *release* IDs** (the rest are master IDs, deferred — see Follow-ups). This data is categorically more trustworthy than our automated reconciliation (~24.3% Discogs coverage today). But it has **nowhere to live**: `entity.release_identity` is an external-id↔external-id crosswalk with no `library_id` key and is never read on the library path; `lml_cache.release_resolution_cache` serves only the flag-gated non-library row-less path. So making these corrections take effect requires a small new override mechanism.

## Desired end state

For the 12,218 verified release links, a library-album lookup returns Alex's Discogs release (and its correct tracklist) instead of the fuzzy pick — behind a feature flag, shippable dark, fully reversible.

## Where (verified file:line touchpoints)

- **New table** `lml_cache.library_release_override(library_id INTEGER PK, discogs_release_id INTEGER NOT NULL CHECK (>0), source TEXT DEFAULT 'alex-l-2026', note TEXT, created_at, updated_at)`. LML-private application cache → **`lml_cache.*`, created in the FastAPI lifespan** (precedent: `set_up_release_resolution_cache_schema` / `lml_cache.album_streaming_url_cache`, wired at `main.py:85-99`). **Not** `entity.*`/alembic — that boundary is only for cross-service identity contracts.
[...]

PR #851, opened 2026-07-19
feat(lookup): pin verified library-release→Discogs overrides — read path (LML#850)

jakebromberg:
First of two PRs for [#850](https://github.com/WXYC/library-metadata-lookup/issues/850). This delivers the **read path**, shippable dark behind `LML_LIBRARY_RELEASE_OVERRIDE` (default `False`). The **seeder** that populates the table is the follow-up PR, stacked on this one.

## Problem

A library-album `/lookup` picks its Discogs release — and therefore the tracklist — by a **per-request fuzzy match** (`find_best_typed_match`, the LML#478 80/80 floor). It can pick the wrong pressing/master/sibling release, surfacing wrong or empty tracklists (the LML#506/#560 class the Post-launch service hardening initiative, Project #32, exists to fix). There is no durable place to pin the correct release.

## What this does

Adds an LML-owned override consulted **before** the fuzzy pick, so a hand-verified `library_id → discogs_release_id` mapping wins.

- **New table** `lml_cache.library_release_override` (`library_id` PK → `discogs_release_id` with `CHECK > 0`, `source` stamp), lifespan-bootstrapped like the other `lml_cache.*` caches. `entity/library_release_override.py` holds the schema bootstrap + a one-query prefetch helper (`+ .sql` reference).
- **`lookup/artwork.py`** `fetch_one` binds the pin **before** the LML#604 trust-bind and the fuzzy `search`, short-circuiting the Discogs call — so a hit *reduces* per-request work (aligned with #32).
- **`lookup/orchestrator.py`**: flag-gated `_prefetch_release_overrides` (one query, skips `id≤0`); `state.overridden_ids` threaded into artwork fetch + enrichment.
- **Enrichment carve-out** gated on `overridden_ids` **set membership** (NOT `item.id != 0`) — overridden rows skip the LML#477/#487 sibling-leak title gate; every other row keeps it. A test asserts a non-overridden row still collapses to the sentinel (protection intact).
- Flag `lml_library_release_override` in `config/settings.py` + `docs/env-vars.md` row.

## Testing

Test-first throughout. 5 unit files (schema, read helper, fetch override-beats-fuzzy/beats-carried, enrichment carve-out, orchestrator flag-gate) + 1 `@pytest.mark.pg` integration file (real DDL, CHECK constraint, PK, read helper). Full unit suite green (3797 passed); pg integration green against a local discogs PG. `ruff check`/`format` clean.

## Rollout

Ships dark (flag default `False`, zero added work when off). After the seeder PR lands: seed → flip flag on **staging** → ≥20-sample `extended=True` verification → prod. Fully reversible: `DELETE ... WHERE source = alex-l-2026` + flag off.

Part of #850.


jakebromberg:
Follow-up (stacked): #852 adds the seeder that populates the table. Merge this first.

Issue #803, opened 2026-07-16
[Epic] July 2026 saturation era — hardening follow-ups from the Jul 11–15 investigation


jakebromberg:
Aggregation epic for the hardening follow-ups from the 2026-07-11 → 07-15 saturation investigation, so the cluster has one durable postmortem home.

## Root-cause narrative (all times UTC)

- **Since ~Jun 27–30** — sustained `/api/v1/lookup` pressure predates everything below: the streaming warm-probe TimeoutError groups ([LIBRARY-METADATA-LOOKUP-1D](https://wxyc.sentry.io/issues/LIBRARY-METADATA-LOOKUP-1D) et al., 58K+ events) and the asyncpg RESET-on-release pool-contention fingerprint ([LIBRARY-METADATA-LOOKUP-X](https://wxyc.sentry.io/issues/LIBRARY-METADATA-LOOKUP-X)) both date from this window.
- **Jul 10 20:14** — #755 ships the Discogs saturation breaker to prod (`770e414` is the breaker's first commit; it did not exist before).
- **Jul 11 06:02** — the first `0 6 * * *` double-cron after breaker birth (Backend-Service `flowsheet-metadata-backfill` + `rotation-lml-identity-backfill` fire simultaneously, with four more LML-touching crons stacked 04:15–05:35) trips the breaker. Per-shed error logging then emits ~8K events/day for 4 days ([LIBRARY-METADATA-LOOKUP-1Q](https://wxyc.sentry.io/issues/LIBRARY-METADATA-LOOKUP-1Q), 32,520 events). The breaker itself degraded correctly (cache-only, no 500s) throughout.
- **Jul 14 04:03–04:10** — the actual 5xx burst that triggered the backfill pause: #783 (`raw_message=None` structured-only lookups → ValidationError → blanket 500), 18 events. Fixed same day (`69a40b2`). Secondary 5xx sources in the same window: the library.db hot-swap race (aiosqlite "no active connection" when `/admin/upload-library-db` swaps mid-bulk) and transient Railway PG connect failures.
- **Jul 14 ~22:00** — backfills paused → breaker storm collapses (32,453 of 32,520 events precede this cutoff). Note: this is *before* the #706 clamp deploy at 23:53 — the pause ended the storm, not the deploy.
- **Jul 14 06:00 run still alive Jul 15 05:25 (23h)** — Backend-Service#995 pacing bounds request rate but not runtime; a slow run collides with the next day's whole cron stack.
[...]

Issue #699, opened 2026-06-25
feat(lookup): expose songwriter/composer credits (release_track_artist writer roles) for BMI reporting


jakebromberg:
## Problem

WXYC must keep submitting **BMI performance lists** after the 2026-08-31 tubafrenzy turndown. Those lists carry the **songwriter/composer** per played track. Today that composer is captured only in tubafrenzy's classic `/wxycdb` flowsheet UI (the `BMI_COMPOSER` MySQL column, read by `RecentBMIServlet`) — a capture path that dies with tubafrenzy. Backend's Postgres `flowsheet` has **no composer column**, and the BS→tubafrenzy mirror writes `''` for `BMI_COMPOSER` on every dj-site/iOS entry, so composer capture is already eroding.

The data needed to replace this already exists in the discogs-cache and already flows through LML — it's just discarded:

- `release_track_artist` carries per-track credits with `extra` (0 = main performer, 1 = extra) + `role` (source strings like `Written-By`, `Composed By`, `Music By`, `Lyrics By`); added in [discogs-etl migration 0005](https://github.com/WXYC/discogs-etl/blob/main/alembic/versions/0005_release_track_artist_role.py).
- `discogs/cache_service.py::get_release` already queries `release_track_artist`. Per WXYC/library-metadata-lookup#588, the per-track path used to append all rows (incl. `extra=1` writers) to `TrackItem.artists`, then was fixed to performers-only because "`TrackItem` carries no `extra` or `role` field, so callers cannot filter." The release-level handling in the same function already separates `extra=0` → `artist_credits` and `extra=1` → `extra_artist_credits`.

So the writer credit is in hand at resolution time and currently dropped. This ticket surfaces the **writer-role subset** as a dedicated output field so a downstream consumer can populate a flowsheet composer for BMI.

## Desired end state

- LML's track/lookup response exposes a composer/writers field for the resolved track, derived from `release_track_artist` rows where `extra=1` and `role` is in a defined writer set.
- The field is added to the SSOT `wxyc-shared/api.yaml` first (schema-first rule), then codegen regenerated for LML/BS/iOS/Android.
- Provenance distinguishes a **track-level** writer credit from a **release-level fallback** (the existing `extra_artist_credits` separation gives the release-level set for free).
- Empty/omitted when nothing resolves; never fabricated. The downstream consumer falls back to artist-as-proxy (matching tubafrenzy's existing "auto-fill from Artist").

[...]

PR #896, opened 2026-07-22
perf(#894): drop 4 dead hydration child reads from /lookup (14→10 PG round-trips)

### LINKED ISSUE #894: Drop 4 dead hydration child reads from every /lookup (release_video, artist_alias, artist_name_variation, artist_member)
## Parent

WXYC/library-metadata-lookup#803 (Epic — July 2026 saturation era).

## Reworked scope (2026-07-22)

The original framing ("drop 4 dead child reads from the shared hydration") was **wrong** and its PR (#896) was closed. Those children — `release_video`, `artist_alias`, `artist_name_variation`, `artist_member` — are **not dead**: they populate the public `/api/v1/discogs/release/{id}` and `/discogs/artist/{id}` responses (`discogs/service.py` fills them on a cold API miss). Dropping them from the shared read-through-cached `get_release` / `get_artist_details` makes **warm cache hits return them empty while cold misses return them full** — a silent contract inconsistency on documented response models. This is the §6.3 fence in `plans/lookup-latency-plan.md` ("field-drops can't touch the shared cached method").

## What to build

Introduce a **lean `/lookup`-only hydration read path** — a variant of `get_release` / `get_artist_details` (or a reduced-child-set flag on them) that **omits the 4 children `/lookup` never uses** — **without mutating the shared cached methods**. The shared `get_release` / `get_artist_details` (and therefore `/discogs/*`, the warmer, and genre-agg) keep returning all children unchanged on both warm and cold paths. `/lookup` switches to the lean path → 14→10 round-trips on the lookup path only.

This is the first, safe structural step. **#895 (L4c)** then collapses this same lean path's per-child reads into `json_agg` / `LATERAL` (10→~2). Land #894 before #895.

## Acceptance criteria

- [ ] `/lookup` uses a lean hydration path that does NOT read the 4 unused children; hydration round-trips on the lookup path drop (target 14→10).
- [ ] The shared cached `get_release` / `get_artist_details` are UNCHANGED — `/api/v1/discogs/release|artist` still return `videos` / `aliases` / `name_variations` / `members` on BOTH warm-cache hits and cold misses (no warm-vs-cold divergence).
- [ ] `release_track_artist` (#699 writer credits) is still read on the lean path.
- [ ] `/lookup` response shape is unchanged for a fixture release and artist.
- [ ] TDD: failing unit test first (spy: the lean path skips the 4 children; the shared path still reads them).

## Blocked by

None. (Precedes #895, which extends this lean path.)

## Related

WXYC/library-metadata-lookup#895, #699, #803. Supersedes the closed PR #896.

jakebromberg:
## Summary

Lever **L4a** of the July 2026 saturation-era hardening epic (#803). The release/artist cache hydration in `discogs/cache_service.py` read four child tables on **every** `/api/v1/lookup` whose values are never surfaced in the `/lookup` response nor consumed by any caller:

- `release_video` (in `get_release`)
- `artist_alias`, `artist_name_variation`, `artist_member` (in `get_artist_details`)

Removing these reads cuts the per-lookup hydration from **14 → 10 PG round-trips** at zero behavioral risk.

## Why it's safe

- The corresponding response fields (`ReleaseMetadataResponse.videos` and `ArtistDetails.aliases` / `name_variations` / `members`) **stay in the models, emitted empty** — the response shape is unchanged.
- `release_track_artist` is **deliberately preserved** — it carries the #699 per-track writer credits. So are `release_genre` / `release_style` / `artist_url`.
- The artist-search-alias composer keeps its data: it reads these child rows via the **untouched bulk path** (`get_artist_details_bulk`), not the singular `get_artist_details`.
- The write paths (`write_release` / `write_artist_details`) still persist every child row, so nothing is lost in PG — the data is simply no longer re-read on hydration.

## Tests (TDD)

- New `TestLookupHydrationDropsDeadChildReads` asserts (a) none of the four dead tables are queried during a lookup, (b) `release_track_artist` and the other live children still fire, (c) the response shape is unchanged for a fixture release/artist (Stereolab, Juana Molina). Written first, confirmed red, then green.
- Updated the two stale cache-read tests (`TestGetRelease` videos, `TestGetArtistDetails.test_full_details`) to assert the new empty-field contract.
- Full unit suite green (4000 passed); mypy + ruff clean.

Closes #894

### DIFF
--- a/discogs/cache_service.py
+++ b/discogs/cache_service.py
@@ -1174,7 +1174,13 @@ class DiscogsCacheService:
                 ),
             )
 
-            # Genre/style/video tables may not exist if the pipeline hasn't been re-run
+            # Genre/style tables may not exist if the pipeline hasn't been re-run.
+            # LML#894 (lever L4a): the `release_video` child read was dropped here.
+            # Its rows populate `ReleaseMetadataResponse.videos`, which is never
+            # surfaced in the `/lookup` response nor consumed by any caller, so
+            # reading it on every hydration was pure dead cost. The column stays in
+            # the response model (emitted empty) and `write_release` still persists
+            # videos from the live-API path, so the data is preserved in PG.
             genre_style_results = await asyncio.gather(
                 self.pool.fetch(
                     "SELECT genre FROM release_genre WHERE release_id = $1",
@@ -1182,15 +1188,6 @@ class DiscogsCacheService:
                 ),
                 self.pool.fetch(
                     "SELECT style FROM release_style WHERE release_id = $1",
-                    release_id,
-                ),
-                self.pool.fetch(
-                    """
-                    SELECT sequence, src, title, duration, embed
-                    FROM release_video
-                    WHERE release_id = $1
-                    ORDER BY sequence
-                    """,
                     release_id,
                 ),
                 return_exceptions=True,
@@ -1203,11 +1200,6 @@ class DiscogsCacheService:
             style_rows = (
                 genre_style_results[1]
                 if not isinstance(genre_style_results[1], BaseException)
-                else []
-            )
-            video_rows = (
-                genre_style_results[2]
-                if not isinstance(genre_style_results[2], BaseException)
                 else []
             )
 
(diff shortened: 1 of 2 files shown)


jakebromberg:
Closing: this took the wrong approach. It dropped the 4 child reads from the SHARED read-through-cached `get_release`/`get_artist_details`, which changes the public `/api/v1/discogs/release|artist` contract — those children (`videos`/`aliases`/`name_variations`/`members`) are returned full on a cold API miss (`discogs/service.py`) but would now be empty on a warm cache hit. That is the §6.3 fence in the latency plan. #894 has been reworked to introduce a lean `/lookup`-only read path instead (shared method untouched); it will be re-implemented under that corrected scope.

Issue #782, opened 2026-07-14
Apple Music enrichment: on-Apple releases persist as null apple_music_url — first-lookup timeout / eventual-consistency, not matcher over-rejection


jakebromberg:
## Problem

`apple_music_url` persists as **null for releases that are demonstrably on Apple Music**. A read-only audit of the 60 most-recent WXYC flowsheet entries (2026-07-13) found 21 track rows with a null Apple link; independently checking each against Apple's own catalog (public iTunes Search API) shows **at least 14 of 21 (~67%) have a valid Apple Music album or track** — so the nulls are recall failures, not honest "not on Apple Music" signals.

**The earlier framing of this ticket blamed the iTunes match/verification leg (over-tightening after #389). A prod repro now shows that was wrong**, and this rewrite corrects it. The Apple URL on a *first* lookup comes from a single **4s synchronous probe** inside a `/lookup` that currently runs **~9s**; a null is a probe **timeout that Backend-Service then persists permanently**, not the matcher rejecting a valid result. Re-running the 6 "clean" pairs against prod, **5/6 resolve reliably** (the persisted nulls were transient/stale) and Daft Punk *Discovery* returns its Apple URL — so it was never a "verification over-rejection" smoking gun. The one that stays null (Friko) fails on **album-title divergence**, not verification (details below).

## The Apple leg is two paths, not one

The `/lookup` path (these audit rows are `enriched_match` with a real `discogs_url`, i.e. the `/lookup` enrichment path) assembles `apple_music_url` from **two independent legs**:

1. **Synchronous track-probe** — `lookup/enrichment/item.py:186-201`. Searches Apple by **track** (`search_term = ctx.song or item.title`, `item.py:89`), wrapped in a **4s** `asyncio.wait_for` (`apple_music_lookup_timeout_s()`, default 4000ms, env `LML_APPLE_MUSIC_LOOKUP_TIMEOUT_MS`). No feature flag — always runs. **This is the only source of `apple_music_url` on a first-ever lookup.**
2. **Album-keyed post-process** — `lookup/streaming_url_postprocess.py` (#706, 2026-06-28). Runs only when leg 1 returned null. It is **eventually-consistent by design**: on a cold-cache first lookup it returns null *synchronously* and fills `lml_cache.album_streaming_url_cache` via a fire-and-forget background warm (album-level `find_album_match`). It **cannot** supply a URL on the first lookup — only the *next* lookup of the same album hits.

[...]

Issue #337, opened 2026-05-15
Perf regression in /api/v1/lookup: cold-cache compilation queries now 9-17s (was <5s pre-2026-05-13)


jakebromberg:
## Perf regression: cold-cache `/api/v1/lookup` went from <5 s to 9–17 s on 2026-05-13

Direct prod measurement, 2026-05-15 ~02:00 UTC, auth'd `POST /api/v1/lookup` against `library-metadata-lookup-production.up.railway.app`:

| Artist (cold cache) | wall | `search_type` | `api_calls` | `api_time_ms` |
|---|---|---|---|---|
| 22 Beaches / Breathing | 16.0 s | compilation | 5 | 22,927 |
| El Keamo / Esperando | 9.5 s | compilation | 5 | 13,706 |
| Rita Villa / Czardas | 14.4 s | compilation | 7 | 18,582 |
| Sonido Dueñez / La Piragua | 9.5 s | compilation | 5 | 13,478 |
| Amy Gadiaga / Imma Pick You Up | 16.6 s | compilation | 8 | 20,829 |
| The Fly Girlz / Born 2 Be Fly | 14.3 s | compilation | 7 | 19,160 |
| Felt / Evergreen Dazed | **4.8 s** | direct | 2 | (`pg_hits=4`, mostly cached) |

The pattern is unambiguous: any query that exhausts `ARTIST_PLUS_ALBUM` and falls into `TRACK_ON_COMPILATION` is now serializing 5–8 Discogs API calls behind the 60 req/min rate limit, blowing past Backend-Service's 5 s runtime timeout on `apps/backend/services/lml/lml.client.ts`. Backend-Service is filing a paired BS-side mitigation (raise that timeout to 30 s + write synthesized search URLs on catch), but the underlying issue is here.

## Inflection point: four PRs on 2026-05-13

Each addition is locally well-motivated; the cumulative cost on a cold-cache compilation lookup is the problem.

| Commit | What it added |
|---|---|
| [`67ec51f`](https://github.com/WXYC/library-metadata-lookup/commit/67ec51f) | Canonical-artist resolver pre-pass (#318). Trigram query against `discogs-cache` PG **on every lookup, unconditionally**, for shadow logging — even when `LML_RESOLVE_ARTIST_CANONICAL=false` (still prod default). The swap that would skip downstream strategies is gated; the cost isn't. |
[...]

PR #898, opened 2026-07-22
Cache-first Apple Music URL probe on the /lookup happy path (L1)

### LINKED ISSUE #893: Cache-first Apple Music probe on the /lookup hot path (skip live Apple on repeat lookups)
## Parent

WXYC/library-metadata-lookup#803 (Epic — July 2026 saturation era).

## Reworked scope (2026-07-22): separate track-scoped cache

The original write-back-to-album-cache approach (PR #898) was **wrong** and has been closed. It wrote `find_track_url`'s per-**track** Apple deep-link into the album-keyed shared `lml_cache.album_streaming_url_cache` — the same table the streaming post-process fills with **album** URLs via `find_album_match` and reads read-through. That caused wrong-track URLs on repeat lookups (the key has no song axis), **permanent** cross-seam cache poisoning (hits never expire), and warm-vs-cold divergence on the public `apple_music_url` field.

The iOS Apple Music button must deep-link to the **played track**, so an album-granular URL is not acceptable. The fix is a **track-scoped cache**, not a bend of the album cache.

## What to build

1. A NEW LML-owned cache table `lml_cache.track_streaming_url_cache`, PK `(service, artist_normalized, album_normalized, song_normalized)` — a real song axis. Lifespan-bootstrapped via `CREATE TABLE IF NOT EXISTS` in `main.py`, same pattern + normalization as `album_streaming_url_cache`. Do NOT touch the album cache.
2. Cache-first happy-path probe in `lookup/enrichment/item.py`: when the `/lookup` request carries a played track, peek `track_streaming_url_cache` on `(apple_music_track, artist, album, song)`. **HIT** → skip the ~4.8s live `find_track_url` Apple probe and return the cached exact-track deep-link. **MISS** → run `find_track_url` live (preserving the first-lookup URL — no #782 null) AND write the resolved track deep-link to the TRACK cache. Track-absent lookups fall back to today's album/post-process behavior.
3. Thread the kill-switch flags: `lml_persist_streaming_urls` (master) + `lml_persist_streaming_url_apple_music` (per-service) must gate BOTH the peek and the write, mirroring the post-process — flipping them off during an incident must fully disable L1's cache read/write.

## Constraints / regression safety

- NEVER write track deep-links into `album_streaming_url_cache` (the PR #898 poisoning bug). Track URLs live only in the new track table.
- The peek MUST be keyed on the song, so a DIFFERENT song of the same album is a MISS (no wrong-track).
- First-lookup `apple_music_url` must stay present (no #782 / BS#1192 persisted null).
- The synthesis-path artwork probe stays synchronous and untouched.

## Acceptance criteria

- [ ] New `lml_cache.track_streaming_url_cache` table exists, lifespan-bootstrapped, keyed on `(service, artist, album, song)`.
- [ ] Repeat lookup of the SAME `(artist, album, played-track)` skips the live Apple probe and returns that track's deep-link.
- [ ] A DIFFERENT song of the same album is a cache MISS (never served the first song's URL).
- [ ] First-lookup `apple_music_url` is present (no persisted null).
- [ ] The album-keyed `album_streaming_url_cache` is never written by L1.
- [ ] Kill-switch flags disable both the peek and the write when off.
- [ ] TDD: failing tests first, including **different-song-same-album** AND **flag-off**; unit + a `pg` integration test for the cache write.

## Blocked by

None.

## Related

WXYC/library-metadata-lookup#782, WXYC/Backend-Service#1192, #706, #803. Supersedes closed PR #898.

jakebromberg:
Closes #893

## Summary

Realizes **lever L1** of the July 2026 saturation-era epic (#803), the deferred follow-up on #706. The happy-path (`library_row_acceptable`) Apple Music probe in `lookup/enrichment/item.py` now peeks `lml_cache.album_streaming_url_cache` for the REQUEST album's Apple Music URL **before** running the live `find_track_url` probe.

- **Cache hit** → surface the cached URL and skip the live Apple call entirely (~2.5ms SELECT instead of the ~4.8s live probe — a shared 1 req/s AsyncLimiter inside the `apple_music_lookup_timeout_s` ceiling that ran live outside every deadline).
- **Cache miss** → run the live probe synchronously (so the first-lookup `apple_music_url` is still present) **and** write the resolved URL back so subsequent lookups hit.

The flood is dominated by repeat lookups of the same albums, so caching the URL removes Apple from the hot path for the repeats that make up the load: roughly **−4.8s on repeat lookups**.

## Regression safety (the must-holds from the issue)

- **Cache-FIRST, not background.** The inline probe is the only source of `apple_music_url` on a first lookup; it still runs synchronously on a miss, so no persisted null is introduced (guards #782 / WXYC/Backend-Service#1192).
- **Only the URL** is cache-first. The synthesis-path artwork probe (`find_track_metadata`) is deliberately **untouched** — still synchronous — because nulling artwork is the #337 outage class.
- **Only a resolved URL is written back.** A probe miss records no cache row, leaving the album-keyed post-process to own null-miss recording; L1 never writes a null.

## Scope / gating

The peek and write-back engage only when the request carries **both an artist and an album** (the cache is album-keyed) and a discogs-cache PG handle is configured. Artist+song-only lookups fall straight through to the live probe as before. This is a distinct, additive seam from `lookup/streaming_url_postprocess.py` (which stays cache-read-only + background-warm on the response path for the non-happy-path services); the two share the same `apple_music_album` cache rows. Cache peek/write failures are logged and treated as best-effort so a PG hiccup never sinks a lookup.

## Tests (TDD: red first)

- `tests/unit/test_enrichment.py::TestAppleMusicUrlCacheFirst` — cache-hit-skips-probe, miss-probes-and-writes-back, miss-returning-none-writes-no-row, no-PG-degrades-to-live, and synthesis-path-untouched. Confirmed failing against the pre-change `item.py` before the implementation landed.
- `tests/integration/test_apple_url_cache_first.py` (`@pytest.mark.pg`) — drives the real `lml_cache.album_streaming_url_cache` write-back end to end through `enrich_artwork_results`, then asserts a second lookup of the same album is served from PostgreSQL with no second live Apple call.
- The hard-timeout off-path harness (`test_api_lookup_hard_timeout.py`) stubs the new cache seam to a clean miss so its mock `PgSource` can't masquerade as a hit and suppress the probe it asserts still runs in-request.

## Acceptance criteria

- [x] A lookup whose album already has a cached URL issues no live Apple probe.
- [x] A lookup with no cached URL runs the live probe synchronously and writes the resolved URL to `album_streaming_url_cache`.
- [x] First-lookup `apple_music_url` still present in the response — no persisted null.
- [x] Synthesis-path artwork probe unchanged (still synchronous).
- [x] Failing unit test first + a `pg`-marked integration test covering the cache write.

## Local verification

- `tests/unit/` — 4005 passed
- `tests/unit/test_enrichment.py` — 106 passed
- `tests/integration/test_apple_url_cache_first.py -m pg` — 2 passed (throwaway local Postgres 16 on :5433)
- `tests/integration/test_streaming_url_persistent_lookup.py -m pg` — 21 passed; hard-timeout + fixture-guard `-m pg` — 3 passed
- `ruff check` / `ruff format --check` / `mypy --ignore-missing-imports` clean on changed files


jakebromberg:
Closing: this approach poisons the shared album-keyed cache. It wrote find_track_url's per-track deep-link into lml_cache.album_streaming_url_cache (the album-keyed table the streaming post-process fills via find_album_match), causing wrong-track URLs on repeat lookups, permanent cross-seam poisoning, and warm-vs-cold divergence on the public apple_music_url field. #893 has been reworked to a SEPARATE track-scoped cache (new lml_cache.track_streaming_url_cache keyed on service+artist+album+song, kill-switch-gated, album cache untouched); L1 will be reimplemented under that scope.

Issue #915, opened 2026-07-23
Replace shared LML_API_KEY with a per-consumer lml_cache.api_keys table


jakebromberg:
## Tracking

This issue tracks the full effort to replace the single shared `LML_API_KEY` bearer token with a per-consumer `lml_cache.api_keys` table, motivated by the 2026-07-23 `LML_API_KEY` rotation incident (a multi-hour 403 lockout that couldn't be attributed to a specific caller). The complete design and chained-PR plan is reproduced below.

**Delivery status:**
- **PR 1** (this repo) — schema + in-process `ApiKeyCache` + dual-accept `require_lml_key`. *Open.*
- **PR 2** (this repo) — `scripts/api_keys` seed/revoke/list CLI. *Open, stacked on PR 1.*
- **PR 3** (tubafrenzy) — add the LML key to `REQUIRED_VARS` in `install-webapps.sh` (closes a silent-fail gap). *Not started — independent, can land anytime.*
- **Rollout** — HITL, one consumer at a time (wxyc-canary → operator script → Backend-Service → tubafrenzy), live secrets. *Not started.*
- **PR 4** (this repo) — remove the legacy fallback + docs cleanup. *Gated on all four consumers showing zero `legacy_shared_key_used` for 7 consecutive days.*

This issue stays open until the rollout completes and PR 4 lands; the code PRs below are the first two installments, not the whole plan.

---

# LML per-consumer API keys — replace the shared `LML_API_KEY` with a `lml_cache.api_keys` table

No GitHub issue yet — this plan gets filed as one (with this doc as the issue body) once it clears review, per the org's Planning & PRs convention. Motivated by the 2026-07-23 `LML_API_KEY` rotation incident: rotating LML's own copy left Backend-Service's EC2 `.env` and wxyc-canary's Secrets Manager copy stale, producing a multi-hour 403 lockout that was misdiagnosed twice (thought to be Backend-Service's `backend` container, then `enrichment-worker`) before being traced to canary's independently-stale copy — because nothing in the current scheme can attribute a request to a specific caller. (Backend-Service staying 403 after the rotation had a second, distinct contributor — a `TARGET`/`RESTART_TARGET` shell-variable-shadowing bug in `set-ec2-env-var.yml`'s SSH restart step meant `backend` never actually restarted with the rotated key; that bug is already fixed in the same file and isn't part of what this plan addresses.)

## Problem recap (verified against code)

[...]

PR #916, opened 2026-07-23
feat(auth): per-consumer lml_cache.api_keys with dual-accept require_lml_key

jakebromberg:
Part of #915. First of two chained PRs (PR 2 — the seed/revoke/list CLI — is stacked on this branch).

## What this does

Replaces the single shared `LML_API_KEY` bearer-token compare in `core/auth.py:require_lml_key` with a table of per-consumer credentials, so a future rotation becomes issue-new → confirm → revoke-old instead of a synchronized cutover across every caller — the structural gap the 2026-07-23 `LML_API_KEY` rotation incident exposed (updating LML's own copy left Backend-Service and wxyc-canary's independent copies stale, with no overlap window and no way to attribute a rejected request to a specific caller).

- **Schema** (`entity/api_keys.py` + `entity/api_keys.sql`) — `lml_cache.api_keys` (hashed token, one row per consumer, `revoked_at` soft-delete), an LML-owned `lml_cache.*` table bootstrapped from the lifespan with `CREATE ... IF NOT EXISTS`, no alembic — matching the six sibling `entity/*` modules.
- **In-process cache** (`core/api_key_cache.py`) — `ApiKeyCache` keeps the request path DB-free: an in-memory `{key_hash: caller_name}` snapshot refreshed on a background interval, mirroring the `core/event_loop_lag.py` sampler + accessor/reset seam. A transient discogs-cache PG blip during refresh keeps the last-known-good snapshot rather than clearing it. `touch_last_used` is fire-and-forget, anchored in a module-level `_background_tasks` set (the strong-reference pattern already used in three other modules here).
- **Dual-accept auth** (`core/auth.py`) — a resolving table-backed key passes (caller tagged on the request's Sentry span, `last_used_at` touched); the still-live legacy key also passes but logs at WARNING (`reason=legacy_shared_key_used`) so not-yet-migrated traffic stays visible for the length of the rollout. The misconfiguration 500 now fires only when neither the legacy key nor any active table-backed key is configured.
- **Wiring** (`main.py`) — schema DDL as one more `bootstraps` entry; the cache as a separate best-effort start/stop pair reusing the block's `source`, with only the stop handle threaded to shutdown.
- **Config + docs** — `lml_api_key_cache_refresh_seconds` (default 300, `ge=1`) + `LML_API_KEY_CACHE_REFRESH_SECONDS` documented in `docs/env-vars.md`.

## Testing

TDD throughout (red → green → refactor). New unit coverage for `hash_token`/`generate_token`, all `require_lml_key` branches (including the changed 500 condition and revoked-treated-as-miss), `ApiKeyCache.resolve`/`refresh` (a raising fetch leaves the prior snapshot intact), and the `touch_last_used` background task running to completion under a real loop. Integration (`-m pg`): idempotent bootstrap, seed → refresh → resolve round-trip, `touch_last_used` hourly throttle. Local: ruff + mypy clean, 4444 unit passed, 523 `-m pg` passed.

This is strictly additive — `LML_REQUIRE_AUTH` remains the master kill switch, and the legacy key keeps working until the rollout confirms every consumer has migrated (removed later in PR 4).


jakebromberg:
Pushed review-hardening (amended into the schema+cache+auth commit):

- **In-process touch throttle** — `ApiKeyCache.touch_by_hash` now skips scheduling the `last_used_at` write when the key was touched within the last hour, so the steady state is at most one background write per key per window instead of a no-op UPDATE on the contended discogs-cache pool (LML#706) per authenticated request. The SQL `WHERE` throttle stays as belt-and-braces across restarts/workers.
- **Constant-time legacy compare** — the shared key is now checked with `secrets.compare_digest` over the bytes form (constant-time, and no `TypeError`/500 on a non-ASCII bearer token; a unit test pins that).
- **Hash once per request** — the bearer token is hashed a single time and reused for both the lookup and the touch.
- **DDL parity test** — a new unit test asserts `entity/api_keys.sql` and the runtime `_DDL_*` constants stay in lockstep.
- **De-duplicated auth-log fields** — both log helpers now share `_auth_log_fields`.

Local: ruff + mypy clean, 4450 unit + 523 `-m pg` green.

Issue #927, opened 2026-07-27
Discogs-semaphore class priority: reserve interactive headroom below the shared 5-permit gate (durable #924)


jakebromberg:
## Parent

Sub-issue of [WXYC/Backend-Service#1819](https://github.com/WXYC/Backend-Service/issues/1819).

## What to build

Give the shared 5-permit Discogs concurrency semaphore class-aware priority so interactive `POST /api/v1/lookup` never head-of-line-blocks behind bulk/enrichment `/api/v1/lookup/bulk` work. This is the **durable** version of the env-only interim [WXYC/library-metadata-lookup#924](https://github.com/WXYC/library-metadata-lookup/issues/924) (which caps `LML_BULK_GLOBAL_MAX_CONCURRENT` below the interactive lane) and supersedes it once shipped.

Evidence (trailing 14d prod spans, in the #1819 thread): the Discogs **quota** gate `lml.discogs.rate_limiter` waits p95 ≈ 32 ms — not the constraint — while the 5-permit **concurrency** gate `lml.discogs.semaphore` waits p95 ≈ 117 s. `/lookup/bulk` owns ~69% of acquisitions and holds permits for tens of seconds; interactive `/lookup` acquires in ≤36 ms 90% of the time but blows to p99 ≈ 22.7 s when it queues behind bulk. Critically, this collision is on the Discogs semaphore that sits **below** the entry gates named in the PRD's Root cause #4 ([`lookup/router.py#L459-L572`](https://github.com/WXYC/library-metadata-lookup/blob/dbd04d4097e861560833a95b447038be610f1e23/lookup/router.py#L459-L572), [`core/bulk_concurrency.py#L1-L37`](https://github.com/WXYC/library-metadata-lookup/blob/dbd04d4097e861560833a95b447038be610f1e23/core/bulk_concurrency.py#L1-L37)), so class priority must reach down to the Discogs semaphore itself — reserving or partitioning permits per traffic class — not stop at admission control at the entry.

## Multi-worker interaction (LML#747 / Lever B) — must be designed in

[...]

Issue #924, opened 2026-07-26
Reserve interactive Discogs-semaphore headroom: cap LML_BULK_GLOBAL_MAX_CONCURRENT below the interactive lane in prod


jakebromberg:
## Problem

Interactive metadata lookups (`POST /api/v1/lookup`) queue behind bulk/enrichment work on LML's shared 5-permit Discogs concurrency semaphore (`discogs_max_concurrent`, built in `discogs/ratelimit.py`). Measured on LML prod (trailing 14d, Sentry spans):

- `lml.discogs.rate_limiter` (the 50/min Discogs **quota** budget) p95 = **32 ms** — the quota is not the constraint; large headroom.
- `lml.discogs.semaphore` (the 5-permit **concurrency** gate) p95 = **117 s** — this is the bottleneck (~3,700x the quota wait).
- By transaction: `/api/v1/lookup/bulk` = 42,148 semaphore acquisitions, avg **54 s** wait; `/api/v1/lookup` (interactive) = 17,907, p90 36 ms but **p99 22.7 s** — head-of-line-blocked behind the bulk lane.

The interactive p99 of 22.7 s crosses the ~8 s BS/caller and canary budgets, which is what surfaces as lookup / `proxy-library-search` canary trips during enrichment and backfill activity.

**Current prod config makes this worse than the code defaults would.** `LML_DISCOGS_POOL_MAX_SIZE=8` and `LML_PG_POOL_MAX_SIZE=8` are set (overriding the code default of 5); `LML_LOOKUP_MAX_CONCURRENT`, `LML_BULK_GLOBAL_MAX_CONCURRENT`, and `DISCOGS_MAX_CONCURRENT` are unset. Effective topology:

- realtime `/lookup` in-flight gate = `min(8, pool) = 8`
- bulk global gate = `pool = 8`
- Discogs semaphore = **5** (unchanged default)

The entry gates admit up to 8 + 8 = 16 concurrent requests that funnel into 5 Discogs permits, and the bulk lane (8) is not capped below the interactive lane (8) — so bulk can occupy all 5 permits with nothing reserved for interactive traffic.

## Desired end state

[...]

Issue #716, opened 2026-07-04
Cross-request in-flight bound for bulk-family endpoints sharing the event loop + PG pool


jakebromberg:
## Problem

The LML#706 PR3 cap ([PR #714](https://github.com/WXYC/library-metadata-lookup/pull/714), `LML_LOOKUP_MAX_CONCURRENT`) bounds **single `/api/v1/lookup` requests only**. The other `perform_lookup` / pool consumers still have per-request bounds but no cross-request bound, and they share the same single event loop and asyncpg pool (default 5 connections; env-tunable via `LML_DISCOGS_POOL_MAX_SIZE` since [PR #745](https://github.com/WXYC/library-metadata-lookup/pull/745)):

- `POST /api/v1/lookup/bulk` — `LML_BULK_MAX_CONCURRENT` (default 10) bounds items **inside one batch**; N concurrent batches admit N×10 in-flight items.
- `POST /api/v1/identity/bulk-resolve-libraries` — pool-sized semaphore per request (default tracks `LML_DISCOGS_POOL_MAX_SIZE` since #745; `LML_BULK_MAX_CONCURRENT` overrides it outright); unbounded concurrent requests.
- `POST /api/v1/cache/refresh-for-identities` — same shape.
- `POST /api/v1/streaming-check` — per-client rate limits only.

So the pool-starvation regime the #706 cap closes for single-lookup storms is still reachable: e.g. an operator runs a 35k-album bulk drain (or two concurrent batches = 20 in-flight items) while organic `/lookup` traffic flows — the loop + 5-conn pool starve exactly as in the 2026-06-25 incident, with only ≤8 single lookups in flight. The scope note in `lookup/router.py` (above `_LOOKUP_MAX_CONCURRENT_ENV_VAR`) and the `docs/env-vars.md` entry document this residual explicitly.

## End state

A cold storm composed of bulk-family requests cannot starve the shared loop/pool: total in-flight per-item work across concurrent bulk-family requests is bounded process-wide, without throttling the single-`/lookup` path (which has its own cap).

## How (design space — pick during implementation)

Options, roughly in order of fit:
1. A shared process-global permit pool for bulk-family **items** (one semaphore consulted inside each dispatcher's per-item runner), separate env knob, so N concurrent batches share one budget instead of multiplying.
2. A cap on concurrent bulk-family **requests** (coarser; simpler; a single long drain would still hold its full per-batch width).
[...]

PR #951, opened 2026-07-28
Route X-Caller-Class=5 onto the low-priority bulk lane

### LINKED ISSUE #928: Low-priority lane policy for batch/cron/backfill LML callers
## Parent

Sub-issue of [WXYC/Backend-Service#1819](https://github.com/WXYC/Backend-Service/issues/1819).

## What to build

Establish and enforce an explicit policy that batch/cron/backfill callers of LML use a lower-priority traffic class with explicit concurrency limits, so scheduled work cannot compete equally with live interactive search and lookup. Named jobs the PRD calls out: popularity free-text resolution, rotation identity backfills, artist genre enrichment, streaming URL upgrades, and historical metadata backfills. Backfill/drain callers already keep `allow_release_resolution_fallback` off to protect offline work ([`lookup/router.py#L686-L692`](https://github.com/WXYC/library-metadata-lookup/blob/dbd04d4097e861560833a95b447038be610f1e23/lookup/router.py#L686-L692)); this slice generalizes that into a documented low-priority lane plus a guard so a new job cannot accidentally use the interactive lane. The env-only interim [WXYC/library-metadata-lookup#924](https://github.com/WXYC/library-metadata-lookup/issues/924) caps the bulk lane today; this makes the policy durable and self-documenting.

## Acceptance criteria

- [ ] A documented low-priority traffic class exists for batch/cron/backfill LML callers with explicit concurrency limits.
- [ ] The named jobs above are routed onto the low-priority lane.
- [ ] A guard (test/CI check) fails when a new batch/cron caller uses the interactive lane by default.
- [ ] Policy documented (CLAUDE.md or docs/) so future jobs choose the right class.
- [ ] Interaction with the #924 interim and the Discogs-semaphore priority slice documented.

## Blocked by

None — can start immediately.

jakebromberg:
## Summary

Closes #928.

- Adds an optional `X-Caller-Class` header (integer 1-5, forwarded by Backend-Service per BS#1843) to both `POST /api/v1/lookup` and `POST /api/v1/lookup/bulk`, generalizing the bulk drain's implicit low-priority placement (#924) into an explicit, caller-declared policy.
- Class 5 (batch/cron/backfill) additionally routes a single `/lookup` request onto the existing #716 process-global bulk permit (`LML_BULK_GLOBAL_MAX_CONCURRENT`, `core/bulk_concurrency.py`) -- the same low-priority lane `/lookup/bulk` items already share.
- Classes 1-4, an absent header, and any unparseable/out-of-range value are a no-op on both endpoints: today's lane placement is unchanged. This is a hard requirement, not just a default -- callers that predate BS#1843 must see byte-identical behavior.
- Security invariant (carried from the BS#1843 review pinned on the issue): the header is down-rank-only. It can never grant a protected/interactive lane beyond a caller's existing entitlement, and it is honored only on the authenticated Backend-Service-to-LML channel (`require_lml_key` gates the `lookup` router before either handler body runs). On `/lookup/bulk`, no class value -- including 1-4 -- can escalate a caller out of the low-priority lane either, since every item already runs under the global permit unconditionally.

## AC3 CI guard

Implemented as an LML-side routing-invariant test suite (the caller->class registry itself lives in Backend-Service's `shared/lml-client/src/policy.ts`, out of scope here):

- `tests/unit/test_caller_class.py` -- pure-function pins for `resolve_caller_class`, `is_low_priority_caller_class`, and `maybe_acquire_bulk_global_permit` in isolation from the FastAPI wiring.
- `tests/unit/test_lookup_router.py::TestCallerClassLowPriorityLane` -- parametrized over classes 1-5, absent, and invalid: class 5 enters the low-priority lane, everything else stays on today's interactive-only behavior.
- `tests/unit/test_bulk_lookup_endpoint.py::TestCallerClassAntiUpRankInvariant` -- parametrized regression guard proving no class value can escalate a bulk caller out of the low-priority lane.

## Docs

`docs/env-vars.md`'s `LML_BULK_GLOBAL_MAX_CONCURRENT` entry documents the new header, and clarifies scope against two related, separate tickets: this operates at the entry-level admission gate only; it does not reach into the shared 5-permit Discogs concurrency semaphore itself, which is the separate, still-open #927 slice (the durable replacement for the #924 interim at that deeper gate).

## Test plan

- [x] `ruff check .` and `ruff format --check .` clean
- [x] `mypy . --ignore-missing-imports` clean
- [x] `pytest tests/unit/` -- 4271 passed (unit tier; `pg`/`external_api` tiers require infra not available locally, matching CI's job split)


jakebromberg:
CI note: the \`Codegen Freshness\` job is red because \`generated/api_models.py\` hasn't caught up to a recent \`wxyc-shared\` api.yaml doc-only update (commit 01dfaf2, \"document X-Caller-Class + X-Caller-Reason headers alongside X-Caller-Budget-Ms\") -- unrelated to this PR's diff, which never touches \`generated/\`. Confirmed by regenerating locally against both the GitHub copy and the local \`wxyc-shared\` checkout: identical one-line drift, reproducible independent of this branch.

Already tracked and fixed in #952 (\`chore(codegen): sync api_models.py with wxyc-shared api.yaml header prose\`, currently green and mergeable). This check should go green here after #952 merges and this branch rebases onto \`main\`. Every job that actually exercises this PR's diff -- Lint & Format, Type Check, CI Marker Sync, External API Tests, Default Tests, PG Tests -- is green.

PR #964, opened 2026-07-28
refactor: reuse the canonical SpotifyClient in scripts

### LINKED ISSUE #963: Three-way duplication of Spotify client-credentials token/init logic
## Problem

Spotify's client-credentials token acquisition and client init are duplicated three times:

- **Canonical**: \`clients/streaming/spotify.py\` — \`SpotifyClient.__init__\` / \`_ensure_token\`. This is the client actually wired into the running service (\`streaming/dependencies.py\`).
- **Copy 1**: \`scripts/spotify_artist_catalog.py\` — a local \`SpotifyClient\` with its own \`__init__\` / \`_get_token\`, byte-identical to Copy 2.
- **Copy 2**: \`scripts/revalidate_spotify.py\` — \`SpotifyValidator\` with its own \`__init__\` / \`_get_token\`, byte-identical to Copy 1.

A June structural audit flagged the canonical \`SpotifyClient\` and one copy. A second audit pass turned up the third copy (\`SpotifyValidator\` in \`revalidate_spotify.py\`), which had drifted into existence independently but re-implements the identical Basic-auth client-credentials POST to \`https://accounts.spotify.com/api/token\`, using the same \`SPOTIFY_CLIENT_ID\`/\`SPOTIFY_CLIENT_SECRET\` env vars and the same \`https://api.spotify.com/v1\` base URL as the canonical client.

## Fix

Both scripts should subclass the canonical \`clients.streaming.spotify.SpotifyClient\` for token/init, and keep only their script-specific extra behavior (\`get_artist_albums\` pagination; \`validate_url\` single-resource lookup) as thin additions. This is behavior-preserving: same token endpoint, same auth header shape, same env vars, same API base — the only material difference is that the canonical client proactively refreshes the token based on \`expires_in\` rather than only reactively on a 401, which is strictly safer.

Several other scripts in this repo (\`scripts/revalidate_misses.py\`, \`scripts/search_unmatched_compilations.py\`, \`scripts/recheck_streaming_after_mojibake.py\`, \`scripts/track_streaming/__main__.py\`, \`scripts/streaming_availability/__main__.py\`) already import and construct the canonical client directly — this brings the last two duplicated copies in line with that established pattern.

jakebromberg:
Closes #963

## What

Removes two byte-identical re-implementations of Spotify client-credentials token acquisition + init, leaving the canonical `clients/streaming/spotify.py:SpotifyClient` as the single source of truth:

- `scripts/spotify_artist_catalog.py`: local `SpotifyClient` (`__init__` + `_get_token`) → `ArtistCatalogClient(CanonicalSpotifyClient)`, keeping only the script-specific `get_artist_albums` pagination.
- `scripts/revalidate_spotify.py`: `SpotifyValidator` (`__init__` + `_get_token`) → `SpotifyValidator(SpotifyClient)`, keeping only the script-specific `validate_url` single-resource lookup.

Both now import `SpotifyClient` from `clients.streaming.spotify` and construct it with `(client_id, client_secret)`, matching the pattern already used by `scripts/revalidate_misses.py`, `scripts/search_unmatched_compilations.py`, `scripts/recheck_streaming_after_mojibake.py`, `scripts/track_streaming/__main__.py`, and `scripts/streaming_availability/__main__.py`.

## Why

A June structural audit flagged the canonical `SpotifyClient` and one duplicate; a follow-up pass found a third copy (`SpotifyValidator`) that had independently drifted into existence with the identical token/init logic. Three copies of the same Basic-auth client-credentials flow is exactly the kind of drift risk (env var rename, base URL change, auth flow change) this consolidation removes.

## Behavior preserved

Verified before refactoring that the canonical client and both scripts agreed on every observable detail:
- Same token endpoint (`https://accounts.spotify.com/api/token`), same Basic-auth header shape (`base64(client_id:client_secret)`), same `grant_type=client_credentials` form body.
- Same env vars (`SPOTIFY_CLIENT_ID` / `SPOTIFY_CLIENT_SECRET`).
- Same API base (`https://api.spotify.com/v1`) and same `Bearer <token>` header shape for authenticated requests.

One intentional, safe difference: the canonical client proactively refreshes the token based on the response's `expires_in` (with a 60s buffer), where the old script copies cached the token forever and only refreshed reactively on a 401. The 401-triggered refresh path is preserved in both scripts (now resetting the canonical client's `_access_token`) as a defensive fallback, so behavior on an actual 401 is unchanged; the proactive-refresh path only reduces the odds of ever hitting one.

Both scripts also drop their own `httpx.AsyncClient` management in favor of the canonical client's internal singleton-managed client (`_get_client()` / `close()`), matching how the app and other scripts already use it.

## Testing

Added focused characterization tests (`tests/unit/test_spotify_artist_catalog.py`, `tests/unit/test_revalidate_spotify.py`) that pin the token-fetch request shape (URL, Basic-auth header, form body), configured attributes, and the extra script-specific behavior (pagination + bearer header; single-resource lookup + bearer header + 401 refresh). Confirmed these passed against the pre-refactor duplicated implementations first, then adjusted the call shape to match the new subclassing (fewer args, since the HTTP client is now managed internally) and confirmed still green post-refactor. Also added a regression guard asserting the script classes don't redefine `__init__`/`_get_token`/`_ensure_token`, so the duplication can't silently creep back.

Local checks, all green:
- `ruff check .`
- `ruff format --check .`
- `mypy . --ignore-missing-imports`
- `pytest` (4624 passed, 1 skipped, 551 deselected, 2 xfailed)

## Test plan

- [x] `ruff check .`
- [x] `ruff format --check .`
- [x] `mypy . --ignore-missing-imports`
- [x] `pytest`
- [ ] Manual smoke of `scripts/spotify_artist_catalog.py --dry-run` and `scripts/revalidate_spotify.py --dry-run` against real Spotify credentials (no CI coverage for these scripts' `main()`; recommend before running against prod data)

### DIFF
--- a/scripts/revalidate_spotify.py
+++ b/scripts/revalidate_spotify.py
@@ -12,7 +12,6 @@ from __future__ import annotations
 
 import argparse
 import asyncio
-import base64
 import json
 import logging
 import os
@@ -22,6 +21,8 @@ import aiosqlite
 import httpx
 from dotenv import load_dotenv
 from rapidfuzz import fuzz
+
+from clients.streaming.spotify import SpotifyClient
 
 load_dotenv()
 
@@ -33,33 +34,18 @@ logging.basicConfig(
 log = logging.getLogger(__name__)
 
 SQLITE_PATH = "streaming_availability.db"
-SPOTIFY_TOKEN_URL = "https://accounts.spotify.com/api/token"
-SPOTIFY_API_BASE = "https://api.spotify.com/v1"
 
 
-class SpotifyValidator:
-    def __init__(self) -> None:
-        self._client_id = os.environ["SPOTIFY_CLIENT_ID"]
-        self._client_secret = os.environ["SPOTIFY_CLIENT_SECRET"]
-        self._token: str | None = None
-        self._semaphore = asyncio.Semaphore(3)
+class SpotifyValidator(SpotifyClient):
+    """Adds single-resource validation lookups on top of the canonical client's token/init.
 
-    async def _get_token(self, client: httpx.AsyncClient) -> str:
-        if self._token:
-            return self._token
-        creds = base64.b64encode(f"{self._client_id}:{self._client_secret}".encode()).decode()
-        resp = await client.post(
-            SPOTIFY_TOKEN_URL,
-            headers={"Authorization": f"Basic {creds}"},
-            data={"grant_type": "client_credentials"},
-        )
-        resp.raise_for_status()
-        self._token = resp.json()["access_token"]
-        return self._token
+    Token acquisition and client-credentials init are inherited from
+    ``clients.streaming.spotify.SpotifyClient`` (LML dedup) — only
+    ``validate_url``, which the canonical client doesn't expose, is
+    script-specific.
+    """
 
-    async def validate_url(
-        self, client: httpx.AsyncClient, url: str, expected_artist: str, expected_title: str
-    ) -> tuple:
+    async def validate_url(self, url: str, expected_artist: str, expected_title: str) -> tuple:
         """Returns (is_valid, actual_title) or (None, reason) for errors."""
         match = re.search(r"/(album|track|artist)/([A-Za-z0-9]+)", url)
         if not match:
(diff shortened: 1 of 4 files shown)


jakebromberg:
Note on CI: \`Codegen Freshness\` is red, but it's unrelated to this diff. That job only runs on \`pull_request\` (see \`ci.yml\`), regenerates \`generated/api_models.py\` from the currently-published \`wxyc-shared\` schema, and diffs it against the committed copy. The drift it's flagging is the \`degraded\` / \`degraded_reason\` \`LookupResponse\` fields and \`DegradedReason\` enum (LML#930/#931) — main's committed \`generated/api_models.py\` is stale against upstream \`wxyc-shared\` independent of anything here. This PR doesn't touch \`generated/\` or any API contract, and per repo convention \`generated/\` shouldn't be hand-edited outside its regeneration script. Every check that actually exercises this diff (Default Tests, External API Tests, PG Tests, Lint & Format, Type Check, CI Marker Sync) is green.

PR #967, opened 2026-07-28
refactor(release): drop misleading underscore on ResolverDependencies

### LINKED ISSUE #966: refactor(release): drop misleading underscore on cross-module ResolverDependencies
`release/orchestrator.py` defines `@dataclass class _ResolverDependencies`. The leading underscore marks it private, but the class is imported and used across module boundaries:

- `release/dependencies.py` imports it, uses it as a return type, and constructs it
- `release/router.py` imports it and uses it as a `Depends(...)` parameter type
- `release/orchestrator.py` itself uses it as a parameter type in two internal functions
- `tests/unit/release/test_orchestrator.py` and `tests/unit/release/test_router.py` import and construct it directly

A single-underscore prefix is a module-privacy convention in Python. This class is de-facto public API within the `release` package -- three other modules and two test files depend on it directly. The underscore is misleading: it signals "do not import this from outside this file" while the codebase does exactly that.

## The fix

Rename `_ResolverDependencies` to `ResolverDependencies` everywhere it appears: the class definition, every import, every type annotation, every constructor call, and any docstring or comment that names it. This is a pure, mechanical, behavior-preserving rename -- no logic changes.

jakebromberg:
## Summary

`release/orchestrator.py` defines `@dataclass class _ResolverDependencies`, but the class is imported and used across module boundaries: `release/dependencies.py` (import, return type, construction), `release/router.py` (import, `Depends(...)` param type), and both `tests/unit/release/test_orchestrator.py` and `tests/unit/release/test_router.py` (import, construction). The leading underscore signals private-to-this-module, which is no longer true.

This renames `_ResolverDependencies` to `ResolverDependencies` everywhere it appears -- class definition, imports, type annotations, constructor calls, and docstrings. Pure rename, no behavior change. `release/__init__.py` is empty and there is no `__all__` to update.

## Test plan

- [x] `ruff check .` -- clean
- [x] `ruff format --check .` -- clean
- [x] `mypy . --ignore-missing-imports` -- clean (would have flagged any missed reference)
- [x] `pytest` -- 4630 passed, 1 skipped, 2 xfailed (pre-existing, unrelated)
- [x] `pytest tests/unit/release/` -- 115 passed

Closes #966

### DIFF
--- a/release/dependencies.py
+++ b/release/dependencies.py
@@ -12,7 +12,7 @@ from core.dependencies import get_discogs_service
 from discogs.service import DiscogsService
 from entity.store import EntityStore
 from identity.dependencies import get_entity_store
-from release.orchestrator import _ResolverDependencies
+from release.orchestrator import ResolverDependencies
 from streaming.dependencies import (
     get_apple_music_client,
     get_bandcamp_client,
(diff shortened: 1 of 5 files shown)


jakebromberg:
Superseded by #955, which removes the POST /api/v1/releases/resolve endpoint and the entire release/ resolve implementation — release/orchestrator.py, release/router.py, release/dependencies.py, and the release tests, i.e. the exact files this rename touches. With that subsystem being archived there is no lasting target for the ResolverDependencies rename, so closing this in favor of #955.

Issue #959, opened 2026-07-28
TRACK_ON_COMPILATION surfaces rows whose specific WXYC pressing lacks the track


jakebromberg:
> **Status (2026-07-28) — decision + tracking issue.** Near-term fix = **option 1 (tighten the carve-out)**, split out as the implementable sub-issue #973 — the only option with no ETL prerequisite. Options 2 and 3 stay here as future precision work: option 2 (track/artist-credit gate) is unblocked by WXYC/discogs-etl#332; option 3 (structural per-row release map) is net-new schema. This issue keeps the root-cause analysis and the options trade-off; implementation + the precision/recall corpus live in #973.

## Problem

`TRACK_ON_COMPILATION` can surface a WXYC library row whose **specific pressing does not contain the matched track**, because LML validates the track against a *different*, similarly-titled Discogs compilation rather than against WXYC's own copy.

Concrete case (`lookup "i only have eyes for you by the flamingos"`): library row 58775 "Various - Greatest hits of the 50s & 60s" (vinyl, `Z-G 0/19`) is surfaced as a match. WXYC's vinyl copy is the Plaza House 2×LP (Discogs 13332759), whose tracklist does **not** include "I Only Have Eyes for You". The row was surfaced only because `process_release` fuzzy-matched the library title to a *different* track-bearing comp — "Greatest Hits Of The 50's" (Discogs 605487, a Disky CD box) — via the `>= 80` title-ratio carve-out, and validated the track on that release.

Path detail: the query is song+artist with no album, so `album_fallback_should_fire` is false and this runs the **strict** branch of `process_release`. The carve-out that fires is [`track_on_compilation.py:497-500`](https://github.com/WXYC/library-metadata-lookup/blob/main/lookup/strategies/track_on_compilation.py#L497-L500) — `artist_matches_item` fails (row artist "Various" vs query "the flamingos"), `is_compilation_artist("Various")` is true, and `_fuzz.ratio(...) >= 80` admits the row. Validation at [`:551-563`](https://github.com/WXYC/library-metadata-lookup/blob/main/lookup/strategies/track_on_compilation.py#L551-L563) then passes against Discogs 605487, which does carry the track.

## Root cause

Two compounding factors in `search_compilations_for_track` -> `process_release` (`lookup/strategies/track_on_compilation.py`):

[...]

Issue #717, opened 2026-07-04
Album-title fallback surfaces wrong-artist library row, preempting non-library resolution


jakebromberg:
## Problem

`/api/v1/lookup` returns the **wrong album** (or nothing) for a non-library release when the caller includes an album title. Surfaced by the iOS app showing incomplete metadata (no release year, artist bio, or Apple Music / Discogs / Wikipedia links) for **Lone – Galaxy Garden**, which recently played.

The metadata resolves perfectly in LML — but only when the album title is *omitted*. Adding the album title makes the answer worse.

## Reproduction (live prod, authenticated `/api/v1/lookup`, `extended: true`)

| Query sent to LML | Result |
|---|---|
| `artist=Lone, album=Galaxy Garden` (no track) | **0 results** — no non-library fallback for album-only queries |
| `artist=Lone, album=Galaxy Garden, song=Crystal Caverns 1991` | **Wrong album**: *Galaxy 2 Galaxy / "Galaxy to Galaxy"* (library id 70229) |
| `artist=Lone, song=Crystal Caverns 1991` (no album) | ✅ **Correct + fully enriched**: Lone / Galaxy Garden, 2012, Discogs + bio + Wikipedia + Apple Music all present (row-less, id 0) |

"Galaxy Garden" is not in the WXYC library catalog (WXYC shelves two *other* Lone albums — "Once in a While" and "Always inside your head"). LML's non-library resolver (`_resolve_nonlibrary_release`, `LML_RESOLVE_NONLIBRARY_RELEASE=true` in prod) can resolve it from Discogs — and does, when the album title is absent.

## Root cause

The album-title fallback in `search_compilations_for_track` runs `process_release(..., skip_artist_match_filter=True)` — [orchestrator.py#L2304-L2316](https://github.com/WXYC/library-metadata-lookup/blob/2eb9c4c5c51e53b40c6836f34c03650e4f2eb3f3/lookup/orchestrator.py#L2304-L2316). That flag intentionally drops the strict library-side artist prefix filter ([orchestrator.py#L2201](https://github.com/WXYC/library-metadata-lookup/blob/2eb9c4c5c51e53b40c6836f34c03650e4f2eb3f3/lookup/orchestrator.py#L2201)) so reordered/collaborative credits still reach `validate_track_on_release`, deferring artist gating to that fuzzy validator (PR #236).

[...]

PR #974, opened 2026-07-28
Tighten the TRACK_ON_COMPILATION title-ratio carve-out with an extra-scope guard

### LINKED ISSUE #973: Tighten the TRACK_ON_COMPILATION title-ratio carve-out to reject wrong-pressing VA-comp rows (option 1)
Implements the near-term option from #959 — read it for the full root cause. In short: `TRACK_ON_COMPILATION` surfaces a WXYC Various-Artists compilation row whose *specific pressing* lacks the matched track, because the `>= 80` title-ratio carve-out admits a differently-titled comp and the track is then validated against a *different* Discogs release. This ticket is **option 1 (tighten the carve-out) only**. The ETL-dependent options — option 2 (track/artist-credit gate, needs WXYC/discogs-etl#332) and option 3 (structural per-row release map) — remain in #959 as future precision work.

## Problem (concrete, reproducible)

`lookup "i only have eyes for you by the flamingos"` surfaces library row **58775** ("Various - Greatest hits of the 50s & 60s", vinyl `Z-G 0/19`), whose Plaza House pressing (Discogs 13332759) does **not** contain the track. It's admitted because `_fuzz.ratio("greatest hits of the 50's", "greatest hits of the 50s & 60s") >= 80` and the row is a V/A comp — but that 80% title match is to a *different* comp ("Greatest Hits Of The 50's", Discogs 605487) than the one WXYC actually holds, and the track was validated on 605487, not on 58775's pressing.

## Scope — both carve-out sites

The 80-floor logic appears **twice** in `lookup/strategies/track_on_compilation.py`; both must be handled consistently (or the fallback explicitly scoped out, with rationale in-code):

1. **Strict branch** — [`track_on_compilation.py:499`](https://github.com/WXYC/library-metadata-lookup/blob/2f0a042e74dbc9d3bbec241b42d414a7dc0e2670/lookup/strategies/track_on_compilation.py#L499) (`title_score >= 80` inside `process_release`). This is the site the concrete case hits.
2. **Fallback branch** — [`:519-529`](https://github.com/WXYC/library-metadata-lookup/blob/2f0a042e74dbc9d3bbec241b42d414a7dc0e2670/lookup/strategies/track_on_compilation.py#L519-L529) (`_fallback_row_acceptable`, same `title_score >= 80`), reached on the album-title-fallback path.

## Approach (implementer's choice — decided by the corpus, not guessed)

Candidate mechanisms:
- Raise the ratio floor (blunt; measure the recall cost).
- Require a stronger/normalized match — e.g. `token_sort_ratio` on normalized titles, or an 80-floor **plus a length-delta guard** so "…50's" vs "…50s & 60s" fails (the extra "& 60s" is a different comp).
- Reject when the library row's title *also* strongly matches a **different** (non-validated) release — i.e. the title is ambiguous across comps, so a title match alone isn't evidence this is the track-bearing comp.

**Do not reopen #717** (wrong-artist rows preempting non-library resolution): the `is_compilation_artist` guard on this branch exists for that reason — keep it. `is_compilation_artist` is False for coincidental wrong-artist collisions, which is what protects #717; don't remove that gate.

## Corpus requirement — the gate that makes this safe

Deliberately two tiers, so the precision/recall claim can't be hand-waved:

**Tier 1 — hermetic regression fixture (REQUIRED to merge; self-verifiable in CI).** A checked-in test that mocks `DiscogsService` (as the existing strategy tests do) and asserts per labeled case:
- must-**DROP**: row 58775 for "I Only Have Eyes for You" / The Flamingos.
- must-**KEEP**: rows 50963, 8865 for the same query (genuine comp hits).
- must-**KEEP** (recall-risk class): **≥3** VA-comp rows whose WXYC catalog title *differs* from the Discogs comp title but which genuinely contain the track — the exact case a naive floor-raise would break. Assemble as fixtures; source candidates from prod TRACK_ON_COMPILATION lookups or the `compilation_track_artist` disambiguation output.

The change must pass every labeled case. No merge otherwise.

**Tier 2 — prod recall measurement (HITL gate before any prod threshold/flag flip; NOT AFK-verifiable).** Before the tightened floor governs prod, run the change over a sample of real TRACK_ON_COMPILATION lookups and report the precision/recall delta. This needs prod discogs-cache / library data a sandboxed agent cannot reach — it is explicitly a human/analyst step. An autonomous run must **not** claim this from a sandbox; leave it as an open checkbox for a human.

## Constraints

- Repo requires **TDD** (CLAUDE.md / `docs/testing.md`): write the failing Tier-1 case first (58775 currently surfaces → assert it must not), confirm it's red, then implement.
- Keep both carve-out branches in sync, or document in-code why the fallback is exempt.
- No prod-governing threshold change ships until Tier 2 has run.

## Acceptance criteria

- [ ] Tier-1 hermetic fixture exists and is green: 58775 dropped; 50963 & 8865 preserved; ≥3 divergent-title recall cases preserved.
- [ ] Both carve-out sites (`:499` strict, `:519-529` fallback) handled consistently, or the fallback exemption justified in-code.
- [ ] #717 guard intact — a wrong-artist non-comp collision is still rejected (add/keep a test).
- [ ] TDD order followed (failing test precedes implementation).
- [ ] Tier-2 prod recall measurement left as an explicit pre-prod HITL checkbox — not silently skipped or fabricated.

## Related

- #959 — parent decision/tracking issue (full root cause + options analysis).
- WXYC/discogs-etl#332 — unblocks option 2 (the track/artist-credit gate), a complementary precision lever tracked in #959; **not** required for this ticket.

---

jakebromberg:
Implements option 1 from #959 (tighten the carve-out) — the near-term, no-ETL-prerequisite fix. Closes #973; cross-references #959, which keeps the full root-cause analysis and the options trade-off (options 2 and 3 remain future precision work there, out of scope here).

## Problem

`TRACK_ON_COMPILATION`'s title-ratio carve-out (`process_release`'s strict branch and its `_fallback_row_acceptable` album-title-fallback counterpart) admitted a Various-Artists library row into results on `fuzz.ratio(release_album, library_title) >= 80` alone, with no check that the matched library row is the *specific pressing* Discogs actually validated the track on. Concretely: `lookup "i only have eyes for you by the flamingos"` surfaced library row 58775 ("Various - Greatest hits of the 50s & 60s") — its own pressing lacks the track — because its title fuzz-matched (ratio ~87) a *different* Discogs compilation ("Greatest Hits Of The 50's", release 605487) that Discogs validated the track on.

## Fix

Added `_library_title_claims_extra_scope(release_album, library_title)`: reject the carve-out when the library title asserts a significant word the validated release's title doesn't have (e.g. "60s"), while leaving the far more common abbreviation direction — the library title merely *omitting* words the Discogs title has (a dropped subtitle, edition tag, leading article) — untouched, since that claims nothing beyond what was actually validated.

Both carve-out sites (`process_release`'s strict branch; `_fallback_row_acceptable`) route through one shared `_title_ratio_carveout_admits()` verdict function, so they can't drift apart the way a duplicated ad hoc `title_score >= 80` literal would.

This is a deliberately blunt, title-text-only heuristic per #959's options analysis — it will not catch every wrong-pressing collision. See "Known leaks" below for the specific gaps and which ones this PR does vs. doesn't close.

## Reversibility — default-OFF flag

The extra-scope guard is gated behind a new setting, **`LML_TRACK_ON_COMPILATION_SCOPE_GUARD`** (`config/settings.py`, `lml_track_on_compilation_scope_guard`), **default `False`**. With the flag off, both carve-out sites are byte-for-byte the pre-#973 ratio-only behavior — merging this PR alone changes no prod behavior. The gate lives inside the shared `_title_ratio_carveout_admits()` helper, so both call sites are gated consistently by construction. A human flips it on in Railway only *after* the Tier-2 prod recall measurement below; flipping it back off is an instant, fully reversible revert to the ratio-only carve-out. `TestScopeGuardFlagGating::test_flag_off_admits_58775_old_behavior` pins the off-by-default behavior explicitly.

## Encoding-bug fixes (from code review)

Two real defects in the significant-word extraction, independent of the flag (they affect the guard's correctness whenever it's eventually enabled):

- **Curly-apostrophe bug.** The original `title.replace("'", "")` stripped only the ASCII apostrophe (U+0027); a typographic apostrophe (U+2019 — the common "curly" form, e.g. copy-pasted from Discogs) survived to the punctuation-stripping regex and got treated as a token boundary, splitting `"50’s"` into throwaway tokens `"50"`/`"s"` instead of the single significant token `"50s"` the ASCII spelling produced — the same semantic title pair could flip ADMIT/DROP purely on which apostrophe glyph was used. Fixed by stripping all four common apostrophe variants (`'`, `‘`, `’`, `ʼ`) before tokenizing.
- **Diacritics.** The extraction hand-rolled `.lower()` instead of reusing this module's existing `to_match_form` normalizer, so `"Café"` and `"Cafe"` read as different significant words. Now composes `to_match_form` (already used elsewhere in this file) with the apostrophe-stripping fix above.

Both are covered by new hermetic unit tests (`TestSignificantTitleWordsEncoding`) asserting a typographic-apostrophe pair and a diacritic pair behave identically to their ASCII/straight-apostrophe equivalents.

## Housekeeping: module extracted for the line-budget guardrail

The encoding fixes plus the flag gate plus the reason-carrying verdict pushed `track_on_compilation.py` 3 lines over its `tests/unit/test_module_budgets.py` ceiling (850). Per that guardrail's own documented policy ("the answer is to extract a module, not to append"), the self-contained carve-out logic (`_significant_title_words`, `_library_title_claims_extra_scope`, `_title_ratio_carveout_admits`, and their constants) moved to a new file, `lookup/strategies/track_on_compilation_carveout.py` (163 lines, budgeted at 250 per the module's calibration formula). Pure relocation — no behavior change — `track_on_compilation.py` now imports `_title_ratio_carveout_admits` from it.

## Log reason (from code review)

The strict branch's existing debug log printed only the (passing-looking, `>= 80`) `title_score` on reject, with no indication of *why* — indistinguishable from an ordinary below-floor reject once the scope guard could also reject at `>= 80`. The fallback branch logged nothing at all on reject. Both sites now log a `reason=` field (`below_ratio_floor` or `extra_scope`) so a human can grep Railway logs to size the Tier-2 recall impact. Covered by `test_reject_log_names_the_scope_guard_reason` (strict) and `test_fallback_reject_log_names_the_scope_guard_reason` (fallback).

## Corpus (Tier 1 — hermetic, CI-verified)

`tests/unit/test_track_on_compilation_title_carveout.py`, 18 tests. All DROP/KEEP fixtures below explicitly enable `LML_TRACK_ON_COMPILATION_SCOPE_GUARD` (via the `scope_guard_enabled` fixture) since the guard is default-off — that's the only state where the new logic executes.

| Case | Verdict | Provenance |
|---|---|---|
| Library row 58775 ("...50s & 60s") vs. validated Discogs release 605487 ("...50's") — strict branch and fallback branch | DROP | Real prod data — ids, titles, and the query itself are taken verbatim from the #959/#973 issue bodies |
| Library rows 50963, 8865 (same query) | KEEP | ids sourced from #973; **titles/artists and the Discogs releases they bind to are SYNTHETIC** — this sandbox has no access to prod `library.db`/discogs-cache to look up their real text. A human should confirm these two against the real rows before treating them as a faithful pin |
| 3× divergent-title recall class (dropped apostrophe/comma; dropped subtitle tag; dropped leading article + hyphen) | KEEP | **Entirely SYNTHETIC** — invented pairs representing the DJ card-catalog shorthand pattern a naive floor-raise would break. Not sourced from any real WXYC row |
| Wrong-artist non-compilation row ("Galaxy 2 Galaxy" / Lone collision) | DROP (unchanged, guard on) | Adapted from the existing #717 regression fixture in `tests/unit/test_album_title_fallback.py`, re-asserted on the strict branch with the guard enabled to confirm the `is_compilation_artist` gate is untouched |
| Curly-apostrophe / diacritic pairs | Identical verdict to ASCII equivalent | Synthetic, direct unit tests of `_significant_title_words` / `_library_title_claims_extra_scope` |
| Flag off | 58775 admitted (old behavior) | Pins reversibility |

## Known leak, now visible (was previously masked)

`tests/unit/test_orchestrator_gaps.py::TestSearchCompilationsForTrack::test_validates_only_library_matching_releases` is now marked **`xfail(strict=True)`**, with the guard explicitly enabled and its original fixture restored: library title `"Trax Records 20th Anniversary Collection"` vs. Discogs `"Trax Records: The 20th Anniversary Edition"` (ratio 85.4 on `main`, admitted there). An earlier pass at this PR had silently edited that title to a pure-subset variant so it would pass the new guard — masking the suite's only witness to this regression. It's restored now specifically so the regression stays *visible and documented* rather than hidden: a library title that **substitutes** a word the Discogs title doesn't share (`"Edition"` → `"Collection"`) reads as asserting new scope and is wrongly dropped. This is a real, inherent limitation of a title-text-only heuristic — **not fixed by this PR** (the encoding fixes above are unrelated bug classes: apostrophe glyph and diacritics, not semantic word substitution). `strict=True` means this test will fail loudly (XPASS) if a future change happens to close the gap, rather than going stale silently.

## Known leaks — explicitly out of scope

Per direction: don't attempt to close these; they're inherent to option 1 (title-text-only) and are future precision work under #959 options 2/3 (track/artist-credit gate — blocked on discogs-etl#332 — or the structural per-row release map):

- The broad substitution-word recall class beyond the two encoding bugs fixed here (see the `xfail` above).
- Numeric/volume/disc divergence leaking through (e.g. `"Vol 1"` vs `"Vol 2"` would not be caught by this guard as currently scoped).
- The subset-direction precision hole: a library title that only *omits* words is always let through, even in the rare case where the omission itself changes meaning.

## Tier 2 — prod recall measurement (HITL, NOT run)

- [ ] Run the tightened carve-out over a sample of real prod `TRACK_ON_COMPILATION` lookups and report the precision/recall delta **before flipping `LML_TRACK_ON_COMPILATION_SCOPE_GUARD` to `True`** in any environment.

This needs prod `discogs-cache`/`library.db` data this sandbox cannot reach — explicitly left as an open, unchecked item for a human. No prod-governing threshold or flag was flipped; the guard ships default-off.

## Open questions for the reviewer

- Rows 50963 and 8865: I could not source their real titles/artists (no prod DB access from this sandbox) — their fixture text is invented. Please confirm against real `library.db` data, or supply the real titles so the fixture can be tightened from "a plausible synthetic hit" to a faithful pin.
- The 3 divergent-title recall-class fixtures are also fully synthetic. If you have real prod `TRACK_ON_COMPILATION` divergent-title hits (or `compilation_track_artist` disambiguation output) handy, swapping in real examples would strengthen this corpus.
- Is `LML_TRACK_ON_COMPILATION_SCOPE_GUARD` the right name/location, or should it live alongside the other `lml_resolve_*` flags under a different naming convention?

## Local checks

- `uv run --no-sync pytest tests/unit/ tests/integration/` — 4591 passed, 551 deselected (pg/external_api), 3 xfailed (2 pre-existing + the new documented regression above)
- `uv run --no-sync ruff check .` / `ruff format --check .` — clean
- `uv run --no-sync mypy lookup/strategies/track_on_compilation.py lookup/strategies/track_on_compilation_carveout.py tests/unit/test_track_on_compilation_title_carveout.py tests/unit/test_orchestrator_gaps.py tests/unit/test_module_budgets.py config/settings.py --ignore-missing-imports` — no issues

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -495,7 +495,9 @@ class Settings(BaseSettings):
             "tests/unit/test_orchestrator_gaps.py); do not flip True until the "
             "LML#973 Tier-2 prod recall measurement quantifies that cost against "
             "real TRACK_ON_COMPILATION traffic. Fully reversible: flip False in "
-            "Railway to instantly revert to the ratio-only carve-out. See "
+            "Railway to revert to the ratio-only carve-out; the change lands on "
+            "the next deploy (get_settings() is @lru_cache'd, so a var flip takes "
+            "effect on process restart, not instantly). See "
             "WXYC/library-metadata-lookup#973 and #959."
         ),
     )
--- a/config/settings.py
+++ b/config/settings.py
@@ -474,6 +474,31 @@ class Settings(BaseSettings):
             "source-stamped rows + flip False. See WXYC/library-metadata-lookup#850."
         ),
     )
+    lml_track_on_compilation_scope_guard: bool = Field(
+        default=False,
+        description=(
+            "When True, TRACK_ON_COMPILATION's >=80 title-ratio compilation "
+            "carve-out (process_release's strict branch and its "
+            "_fallback_row_acceptable fallback -- both route through the shared "
+            "_title_ratio_carveout_admits) additionally rejects a library row "
+            "whose title asserts a significant word the *validated* Discogs "
+            "release's title doesn't have (LML#973's extra-scope guard), fixing "
+            "the WXYC library row 58775 false positive (LML#959: a title-ratio "
+            "match alone can't tell 'same comp, shortened title' from 'different "
+            "comp, similar title'). When False (default), the pre-#973 ratio-only "
+            "carve-out behavior holds byte-for-byte -- merging this flag changes "
+            "no prod behavior. The guard is a deliberately blunt, title-text-only "
+            "heuristic that also drops some genuine divergent-title compilation "
+            "hits (a library title that *substitutes* a word the Discogs title "
+            "doesn't share, e.g. 'Edition' -> 'Collection', reads as asserting "
+            "new scope -- see the xfail'd regression in "
+            "tests/unit/test_orchestrator_gaps.py); do not flip True until the "
+            "LML#973 Tier-2 prod recall measurement quantifies that cost against "
+            "real TRACK_ON_COMPILATION traffic. Fully reversible: flip False in "
+            "Railway to instantly revert to the ratio-only carve-out. See "
+            "WXYC/library-metadata-lookup#973 and #959."
+        ),
+    )
     lml_emit_server_timing: bool = Field(
         default=True,
         description=(
--- a/lookup/strategies/track_on_compilation.py
+++ b/lookup/strategies/track_on_compilation.py
@@ -69,6 +69,7 @@ from lookup.rowless import (
     _make_rowless_item,
     _resolve_nonlibrary_release,
 )
+from lookup.strategies.track_on_compilation_carveout import _title_ratio_carveout_admits
 from lookup.strategies.track_release_matching import search_album_fuzzy
 from services.parser import ParsedRequest
 
@@ -148,87 +149,6 @@ def _log_album_title_fallback(
             transaction.set_data("album_title_fallback", payload)
     except Exception as e:
         logger.warning("Failed to project album_title_fallback onto Sentry transaction: %s", e)
-
-
-# LML#973: the compilation title-ratio carve-out floor. `process_release`'s
-# strict branch and `_fallback_row_acceptable` (its album-title-fallback
-# counterpart) both gate a Various-Artists library row admitted purely on
-# title fuzz through this one floor plus the scope guard below -- named once
-# at module scope so the two call sites can't drift the way a duplicated
-# ad hoc `title_score >= 80` literal did pre-#973.
-_COMPILATION_TITLE_RATIO_FLOOR = 80
-
-# Below this length a token can't carry independent meaning EXCEPT for the
-# compilation-title vocabulary this guard cares about -- decade tokens ("50s",
-# "60s") and edition markers ("vol") are exactly 3 characters, so the floor
-# here is one lower than the `len(w) > 3` threshold this module's own
-# keyword-search significance filter uses a few lines down in
-# `search_compilations_for_track`.
-_SIGNIFICANT_TITLE_WORD_MIN_LEN = 3
-
-
-def _significant_title_words(title: str) -> set[str]:
-    """Content-bearing words in a compilation title, for the #973 scope guard.
-
-    Strips apostrophes before tokenizing so "50's" and "50s" collapse to the
-    same token -- left alone, the apostrophe would otherwise split "50's"
-    into the throwaway tokens "50" and "s" -- then drops punctuation,
-    ``STOPWORDS``, and words too short to carry meaning on their own.
-    """
-    stripped = title.replace("'", "")
-    words = re.sub(r"[^\w\s]", " ", stripped.lower()).split()
-    return {w for w in words if len(w) >= _SIGNIFICANT_TITLE_WORD_MIN_LEN and w not in STOPWORDS}
-
-
-def _library_title_claims_extra_scope(release_album: str, library_title: str) -> bool:
-    """Does the library row's title assert content the *validated* Discogs
-    release title doesn't?
-
-    Root cause (LML#959): a title-ratio match alone can't tell "same comp,
-    filed under a shortened title" apart from "different comp, similarly
-    worded title" -- library row 58775 ("...50s & 60s") fuzz-matched the
-    Discogs release actually validated to carry the track ("...50's") at
-    ratio ~87, but the two are different pressings, and 58775's own copy
-    lacks the track. The ratio alone can't separate them (both same-comp
-    abbreviations and different-comp near-misses can land at any ratio
-    >= 80); what does is that the library title adds a *new* significant
-    word ("60s") the validated release's title doesn't have -- it asserts
-    scope beyond what was actually validated.
-
-    Deliberately asymmetric: a library title that only *omits* words the
-    Discogs title has (the far more common card-catalog pattern -- dropping
-    a subtitle, an edition tag, a leading article) is unaffected, since it
-    doesn't claim anything the validation didn't cover. Only new words on
-    the library side trip this guard.
-
-    A blunt, title-text-only heuristic (see #959's options analysis): it
-    won't catch every wrong-pressing collision, and a word-level
-    abbreviation that isn't a shared token (e.g. "Volume" -> "Vol") can
-    misread as a "new" word. The precision follow-ups -- #959 option 2's
-    track/artist-credit gate (blocked on discogs-etl#332) and option 3's
-    structural per-row release map -- are what close that gap; #973 is
-    "tighten the carve-out" only, by design.
-    """
-    extra = _significant_title_words(library_title) - _significant_title_words(release_album)
-    return bool(extra)
-
-
-def _title_ratio_carveout_admits(release_album: str, library_title: str) -> tuple[bool, float]:
-    """Shared verdict for the two `>= 80` title-ratio compilation carve-out
-    sites (`process_release`'s strict branch and `_fallback_row_acceptable`).
-    One function so LML#973's fix can't drift between the two call sites.
-
-    Returns ``(admitted, title_score)`` -- callers keep the score for their
-    own debug logging.
-    """
-    from rapidfuzz import fuzz as _fuzz
-
-    title_score = _fuzz.ratio(release_album.lower(), library_title.lower())
-    if title_score < _COMPILATION_TITLE_RATIO_FLOOR:
-        return False, title_score
-    if _library_title_claims_extra_scope(release_album, library_title):
-        return False, title_score
-    return True, title_score
 
 
 async def search_compilations_for_track(
@@ -573,15 +493,14 @@ async def search_compilations_for_track(
                     if artist_matches_item(match, lib_artist):
                         filtered_matches.append(match)
                     elif discogs_is_compilation and is_compilation_artist(match.artist or ""):
-                        admitted, title_score = _title_ratio_carveout_admits(
-                            release_album, match.title or ""
-                        )
-                        if admitted:
+                        verdict = _title_ratio_carveout_admits(release_album, match.title or "")
+                        if verdict.admitted:
                             filtered_matches.append(match)
                         else:
                             logger.debug(
                                 f"Rejected '{match.title}' for '{release_album}' "
-                                f"(title_score={title_score:.0f})"
+                                f"(title_score={verdict.title_score:.0f}, "
+                                f"reason={verdict.reason})"
                             )
                 matches = filtered_matches
             elif matches and lib_artist and skip_artist_match_filter:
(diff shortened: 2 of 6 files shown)


jakebromberg:
## Code review summary

Reviewed at max depth (initial pass at `847f847`, re-review at `cae96ea`). The initial review surfaced a masked recall regression, two encoding bugs, and an ungated behavior change. All were addressed; the re-review confirmed the fixes are real and complete, with no new blocker.

**Resolved:**
- **Masked regression → honest xfail.** The initial `#973` commit had edited a `test_orchestrator_gaps.py` fixture from a *substitution* title pair (library `"…Anniversary Collection"` vs Discogs `"…Anniversary Edition"`, admitted on `main` at `fuzz.ratio` 85.4) to a subset variant, silently removing the suite's only witness to a real recall regression the guard introduces. Restored to the original substitution pair as `@pytest.mark.xfail(strict=True)`, with the guard explicitly enabled inside the test — it reports xfailed (not xpassed), so it is load-bearing and will XPASS loudly if a future fix closes the gap.
- **Curly-apostrophe bug.** `title.replace("'", "")` stripped only ASCII U+0027, so a typographic apostrophe (U+2019) flipped the verdict (`"…50's"` vs `"…50s"` → DROP with a straight quote, ADMIT with a curly one). Now normalizes U+0027/2018/2019/02BC before `to_match_form`.
- **Diacritics.** The significant-word extraction hand-rolled normalization; it now composes the already-imported `to_match_form`, so `"Café"`/`"Cafe"` fold to one token. Verified neither fix over-corrects: row 58775 still drops (driven by `60s`), and plain-ASCII and Café/Cafe pairs yield no false drop.
- **Ungated → default-OFF flag.** New `LML_TRACK_ON_COMPILATION_SCOPE_GUARD` (default `False`) gates the extra-scope rejection at both carve-out sites via the shared `_title_ratio_carveout_admits` helper (grep-confirmed no residual `>= 80` bypass survives). With the flag off, behavior and results are identical to pre-`#973` — verified by running the guard: flag-off reproduces `main`'s ratio-only verdict exactly. **Merging changes no prod behavior until the flag is enabled.**

**Gate before enabling in prod:** the `#973` Tier-2 recall measurement over real TRACK_ON_COMPILATION traffic must quantify the divergent-title recall cost before `LML_TRACK_ON_COMPILATION_SCOPE_GUARD` is flipped True. It is documented and left unchecked in the PR body.

**Known limitations (flag-gated, deferred to `#959` options 2/3 — not addressed here):** the guard is a deliberately blunt, title-text-only heuristic. With the flag *on* it still (a) drops genuine hits where the library title *substitutes* a word the Discogs title doesn't share (`Edition`→`Collection`, `Volume`→`Vol`), and (b) fails to catch wrong-pressing collisions where the distinguishing token is numeric (`Vol 1` vs `Vol 2`, `Disc 1` vs `Disc 4`) or the library title is a *subset* of the validated release. These matter only once the flag is enabled and are the reason the Tier-2 measurement gates that step.

**Housekeeping:** the carve-out logic was extracted to `track_on_compilation_carveout.py` to stay under the enforced module line-budget; verified behavior-preserving (identical computation, private symbols only, single importer, no circular import).

Merge-ready as a reversible, dormant change.

jakebromberg:
Superseded by #1010, which is the implementation we're landing for #973.

Per a side-by-side review: this PR is ~30 commits behind `main` and edits the inline `process_release` carve-out that `main` has since refactored into `_filter_release_matches` (so its green CI is on a stale base and it would conflict on rebase); it gates the fix behind a **default-OFF** flag (so merging it would not fix prod until a human flips it, leaving #973's acceptance unmet); and it carries a self-acknowledged xfail recall regression (Edition→Collection word substitution) that #1010's length-ratio guard preserves.

#1010 is surgical (+265/-4), on current `main`, fixes the row-58775 false positive on both the strict and fallback branches, and ships a hermetic CI fixture. We're landing #1010 with an added kill-switch flag (default ON, revertible via env) so the deferred prod-recall measurement can run without a deploy-to-revert. Closing this one to avoid a duplicate.

Issue #972, opened 2026-07-28
Library FTS5 search silently degrades to LIKE-AND for any query containing "." (and other FTS metacharacters) — hurts recall for initials/punctuated artist names


jakebromberg:
> **AFK-ready.** Everything needed to implement this without conversation context is below: exact code locations (permalinked to `origin/main`), a verified fix approach, the edge case that trips up the naive version, the test harness to reuse, and copy-paste local check commands. TDD is required in this repo (see `CLAUDE.md`).

## Problem

`LibraryDB._search_uncached` builds an FTS5 `MATCH ?` query directly from the raw query string. Any query containing an FTS5 metacharacter — most commonly `.` (periods in initials), but also `"`, `*`, `:`, `(`, `)` — raises `fts5: syntax error near "…"`. That error is swallowed by the broad `except Exception` in the FTS branch and silently degraded to `_fallback_like_search`, which requires ALL significant tokens to co-occur in a single row (an `AND` across `title` / `artist` / `alternate_artist_name`). This narrows recall for a whole class of artist names.

Concrete, reproduced: passing `query="M.I.A."` (or `"J. Mascis"`, `"St. Vincent"`, `"C. Spencer Yeh"`) into the FTS `MATCH` throws `fts5: syntax error near "."` and never runs a real FTS match — it drops to LIKE-AND, which then requires every token (e.g. "spencer") to appear on one row. Verified locally against the exact production tokenizer:

```
[FAIL] 'M.I.A.'          -> MATCH 'M.I.A.'          => ERROR: fts5: syntax error near "."
[FAIL] 'J. Mascis'       -> MATCH 'J. Mascis'       => ERROR: fts5: syntax error near "."
[FAIL] 'St. Vincent'     -> MATCH 'St. Vincent'     => ERROR: fts5: syntax error near "."
```

Surfaced during the "C. Spencer Yeh / In the Blink of an Eye" lookup investigation. It is a distinct, latent root cause from the credit-binding bug filed as its sibling (#971): even after that fix, punctuated-artist queries still degrade off the FTS path. The degrade is silent, so it hides recall loss across many lookups — any initials/punctuated artist name (M.I.A., J. Mascis, C. Spencer Yeh, St. Vincent, Godspeed You! Black Emperor, etc.), not just this one.

## Where (permalinks pinned to `origin/main` @ `356f29d`)

[...]

PR #976, opened 2026-07-28
fix(#971): bind library row by Discogs release credit in TRACK_ON_COMPILATION

### LINKED ISSUE #971: TRACK_ON_COMPILATION drops a title-matched library row when the WXYC credit is a band name / Discogs ANV that doesn't prefix-match the typed artist (C. Spencer Yeh / In the Blink of an Eye)
## Problem

An artist+song lookup can locate both the correct Discogs release AND the correct WXYC library row, then discard the library row at a strict artist-prefix gate — returning empty library results for a release WXYC physically holds.

Reproduced with `lookup "in the blink of an eye by c spencer yeh"` (parsed: artist="C. Spencer Yeh", song="In the Blink of an Eye", no album). Library results come back empty; the pipeline falls through to TRACK_ON_COMPILATION with 0 results.

Verified facts:

- The library row exists: `library.db` id 57833 — `artist="Burning Star Core"`, `alternate_artist_name="C.S. Yeh"`, `title='"In The Blink of an Eye" 7-inch'`, genre "Rock". "Burning Star Core" is C. Spencer Yeh's band; "C.S. Yeh" is the abbreviated per-release alias.
- On Discogs this is release 2789357, canonical artist "C. Spencer Yeh" (id 378138), credited on this release via the Artist Name Variation "C.S. Yeh"; track A is "In The Blink Of An Eye". The entity's `namevariations` include "C.S. Yeh", "CS Yeh", "Spencer Yeh".
- LML's own live Discogs track search (`type=release&track=In The Blink Of An Eye&artist=C. Spencer Yeh`) returns exactly release 2789357 — the Discogs side resolves the query correctly.

Root cause: in `lookup/strategies/track_on_compilation.py`, `search_compilations_for_track` → inner `process_release`, `search_album_fuzzy(db, "In The Blink Of An Eye")` finds library row 57833 by title, but the strict artist filter then drops it — `artist_matches_item(match, lib_artist="C. Spencer Yeh")` is False (the row is credited to the band / abbreviated alias, and the prefix match fails), and the row is not a VA compilation so the compilation carve-out (which keeps a title-matched compilation row on an 80+ title score) does not apply. The correctly-located row is discarded before `validate_release_for_track` ever runs.

This is the same class as #327 (validate_track_on_release rejects band tracks credited to individual members) and adjacent to #801 (artist+track recall) — a band-name / member-name / abbreviation identity gap.

## Desired end state

For an artist+song lookup where LML resolves and validates the Discogs release the track sits on, the matching WXYC library row is surfaced even when the WXYC credit uses the band name or an abbreviated ANV — provided the row credit matches the resolved release's own credited artist and the typed artist is validated as performing the track on that release.

## Where

- `lookup/strategies/track_on_compilation.py` — `search_compilations_for_track` → inner `process_release`, the strict artist-filter block (rows must survive it to reach `validate_release_for_track`).
- `lookup/matching.py` — `artist_matches_item` (strict prefix match against `artist` / `alternate_artist_name`).
- `discogs/matching.py` — `strip_discogs_suffix` (currently strips only `(N)`; does NOT strip the ANV `*`).

## Suggested approach

Extend the strict filter so a title-matched library row is also kept when it prefix-matches the Discogs release's own credited artist (`release_info.artist`), not only the typed `lib_artist`. Keep it a strict `artist_matches_item` prefix match — no fuzzy loosening. The existing `validate_release_for_track(..., parsed.artist, ...)` on the typed artist remains the safety belt, so binding on the release credit is safe: a row only survives if the typed artist is also validated as performing the track on that release.

Load-bearing detail: `release_info.artist` here is the string "C.S. Yeh*" — the Discogs ANV marker `*` is present (it comes from `_parse_title` splitting the search-result title "C.S. Yeh* - In The Blink Of An Eye"). No existing normalizer strips a trailing `*`: `strip_discogs_suffix` strips only `(N)`; `to_match_form` / `normalize_for_comparison` keep both `*` and periods. Confirmed: `artist_matches_item(row, "C.S. Yeh")` = True but `artist_matches_item(row, "C.S. Yeh*")` = False and `artist_matches_item(row, "C. Spencer Yeh")` = False. So the fix must strip the trailing ANV `*` (and `(N)`) from the release credit before comparing.

## Constraints

- Do not loosen matching globally or introduce a fuzzy artist floor on this path — the strict gate is the #400 contamination guard. Bind only on an exact credit prefix-match plus the existing typed-artist release validation.
- Surgical, well under ~150 lines including tests. Do not touch ARTIST_PLUS_ALBUM (library-only; no release credit to match against; a separate gap).

## Acceptance criteria

- [ ] TDD: a failing test first — `search_compilations_for_track` surfaces library row 57833 for artist="C. Spencer Yeh", song="In the Blink of an Eye".
- [ ] A guard test: a title-matched row whose credit matches neither the typed artist nor the release credit is still dropped.
- [ ] ruff check + format clean (line length 100); mypy clean (the required CI job pre-commit/ruff don't cover).
- [ ] No regressions in existing `track_on_compilation` tests.

## Related

- WXYC/library-metadata-lookup#327 — band tracks credited to individual members (same class; closed)
- WXYC/library-metadata-lookup#801 — artist+track recall (adjacent; closed)
- FTS punctuation recall bug — sibling finding from the same investigation (cross-referenced below once filed)

jakebromberg:
## Summary

An artist+song lookup for artist="C. Spencer Yeh", song="In the Blink of an Eye" returned no library results even though WXYC owns the release. The library row is filed under the band name "Burning Star Core" with `alternate_artist_name` "C.S. Yeh", while Discogs credits release 2789357 to the Artist Name Variation "C.S. Yeh*". In `search_compilations_for_track` → `process_release`, the album-title search finds the row, but the strict artist filter only compared it against the typed artist ("C. Spencer Yeh") — which prefix-matches neither the band name nor the alias — and the row isn't a V/A compilation, so the compilation carve-out couldn't rescue it either. The correctly-located row was discarded before `validate_release_for_track` ever ran.

## Change

The strict artist filter now also keeps a title-matched row when it prefix-matches the release's own Discogs credit (`release_info.artist`), gated specifically on the ANV `*` marker via the new `credit_uses_anv` predicate. `validate_release_for_track` (on the typed artist) runs immediately after and remains the safety belt against a bad bind.

The gate is deliberately ANV-`*`-specific rather than "any Discogs decoration": a bare numeric disambiguator (`"Sun (2)"`) is not an alias signal, and because `validate_release_for_track` only validates the release against the typed artist — never the surfaced library row — treating a stripped generic base name as an independent match-back path would reopen prefix-collision false positives (a "Sun (2)" credit binding "Sunburned Hand of the Man"; a "Various (2)" credit binding a "Various Artists" row while bypassing the title-score gate).

The new `strip_release_artist_suffix` helper strips the trailing ANV `*` and the `(N)` disambiguator for building the comparison string; it is kept separate from `strip_discogs_suffix` so its `*`-stripping doesn't widen that function's contract across its catalog-enrichment callers.

## Testing

- New strategy-level tests through `search_compilations_for_track`: the repro (library row surfaces; validation is called with the typed artist), a validation-failure guard (a credit match is not a validation bypass), an over-admission guard (a row matching neither signal stays dropped), and two numeric-disambiguator regression tests (`"Sun (2)"` and `"Various (2)"` must not bind unrelated rows). The numeric-case tests were confirmed load-bearing by monkeypatching the gate to the pre-fix behavior and observing the wrong row surface.
- Unit tests for `credit_uses_anv` and `strip_release_artist_suffix`.
- Full `tests/unit` + `tests/integration` (excluding `pg` / `external_api`): 4595 passed, 2 pre-existing xfails, no regressions. `ruff check` / `ruff format --check` / `mypy` clean.

## Scope / follow-ups

- Only rescues the case where Discogs itself flags the alias with the ANV `*`. A release credited plainly to a band name with no Discogs decoration is a materially broader problem, addressed on the data side by enriching `library.db` from the WXYC catalog's own cross-reference table (separate discogs-etl ticket).
- The FTS-metacharacter recall degrade surfaced in the same investigation is tracked separately in #972.
- `ARTIST_PLUS_ALBUM` has the same shape of gap in principle but is library-only (no release credit to bind against) and is intentionally out of scope here.

Closes #971

### DIFF
--- a/discogs/matching.py
+++ b/discogs/matching.py
@@ -35,6 +35,11 @@ from wxyc_etl.text import to_match_form as normalize_for_comparison
 # Discogs disambiguation suffix: (2), (22), etc.
 _DISCOGS_SUFFIX_RE = re.compile(r"\s*\(\d+\)$")
 
+# Discogs Artist Name Variation (ANV) marker: a trailing '*' on a release
+# credit signals the name shown differs from that artist's canonical/primary
+# name (e.g. "C.S. Yeh*" credits "C. Spencer Yeh" via the alias "C.S. Yeh").
+_ANV_MARKER_RE = re.compile(r"\*+\s*$")
+
 
 def strip_discogs_suffix(name: str) -> str:
     """Remove Discogs numeric disambiguation suffixes like '(2)' or '(22)'.
@@ -45,6 +50,50 @@ def strip_discogs_suffix(name: str) -> str:
     if not name:
         return name
     return _DISCOGS_SUFFIX_RE.sub("", name)
+
+
+def strip_release_artist_suffix(name: str) -> str:
+    """Remove Discogs release-credit decorations from an artist string.
+
+    Strips the trailing Artist Name Variation marker ('*') and the numeric
+    disambiguation suffix ('(2)', '(22)', ...). No real artist name legitimately
+    ends in '*', so this is safe to apply unconditionally to a release-credit
+    string. Safe to apply to names with neither decoration (no-op).
+
+    Only the canonical Discogs ordering ``Name (N)*`` is handled -- the
+    reverse ``Name* (N)`` leaves a residual '*', since ``_ANV_MARKER_RE`` is
+    anchored to the end of the string. Not worth a more permissive regex:
+    ``credit_uses_anv`` (the caller's gate for whether to trust this string at
+    all) is anchored the same way, so a reverse-order credit simply never
+    reaches this function in practice.
+
+    Distinct from ``strip_discogs_suffix``: that function's callers span
+    several catalog-enrichment scripts operating on artist/label names sourced
+    from the Postgres discogs-cache (not split from a raw search-result
+    title), so widening its contract to also strip '*' there would be an
+    unreviewed blast-radius increase. This helper is scoped to the
+    ``ReleaseInfo.artist``-shaped strings produced by
+    ``DiscogsService._parse_title`` splitting a "Credit* - Album" search
+    result, where the ANV marker is a known, well-understood artifact.
+    """
+    if not name:
+        return name
+    return strip_discogs_suffix(_ANV_MARKER_RE.sub("", name)).strip()
+
+
+def credit_uses_anv(name: str | None) -> bool:
+    """True when ``name`` (a raw Discogs release-credit string) ends in the
+    Artist Name Variation marker ('*').
+
+    Deliberately narrower than "did stripping change anything": a numeric
+    disambiguator alone (``"Sun (2)"``) is NOT an alias signal -- it only says
+    "the Nth Discogs artist named Sun", carrying no information that the
+    credited name means a *different* canonical artist. Callers that want to
+    know "is this credit trustworthy as an alternate name for the searched
+    artist" must gate on this predicate specifically, not on whether
+    ``strip_release_artist_suffix`` altered the string.
+    """
+    return bool(_ANV_MARKER_RE.search(name or ""))
 
 
 def normalize_for_track_comparison(text: str | None) -> str:
--- a/lookup/strategies/track_on_compilation.py
+++ b/lookup/strategies/track_on_compilation.py
@@ -36,6 +36,7 @@ from core.search import (
     song_not_found_with_artist_and_song,
 )
 from discogs.lookup import lookup_releases_by_track
+from discogs.matching import credit_uses_anv, strip_release_artist_suffix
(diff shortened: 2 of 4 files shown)


jakebromberg:
Superseded by #980 (merged), which fixes #971 via the artist name-variation identity bridge. #976's ANV-`*` approach was proven a no-op on the production cache path: the discogs-cache stores the release's canonical credit (no `*` display marker), so the marker this PR keyed on never reaches the matcher on the dominant Wave-A cache hit. Closing without merging.

Issue #1019, opened 2026-07-30
[V/A] Recall index: lml_cache.compilation_track_location (per-track shelf location + precomputed art)


jakebromberg:
Sub-issue of WXYC/library-metadata-lookup#271. **Foundation for the comprehensive multi-location union (#1022).** Freeze-safe (no `lookup/` changes) — build now.

> **Rescoped 2026-07-30** after a grill-through of #271. This is no longer the per-track *identity* store from #271 §1. Under Option B (decouple), the cheap **recall/location index** — which serves the interactive `/lookup` union — is split out from the expensive per-track *identity* matcher (now deferred; see #1020/#1021). This ticket is the index; it has no external-ID resolution and no Discogs-API matcher.

## Problem

The interactive `/lookup` union (#1022) needs to answer "which library shelf locations contain track *T* credited to artist *A*?" deterministically — today it can only gamble on a live Discogs keyword search that is title-distinctiveness-dependent and silently partial (the #271 live-repro). There is no persistent reverse `(track_artist, track_title) → library_id` index.

## Desired end state

A lifespan-bootstrapped, LML-owned table:

```
lml_cache.compilation_track_location
  library_id         integer   -- the WXYC library.id (shelf location)
  track_position     varchar
  track_artist       text      -- credited track artist (reverse-lookup key)
  track_title        text      -- reverse-lookup key
  credit_role        text      -- 'primary' | 'featured' | 'extra' (ranking signal; from release_track_artist.extra/role)
  discogs_release_id integer   -- the comp's matched Discogs release (art re-resolution key)
  artwork_url        text      -- PRECOMPUTED cover URL (per-comp; denormalized onto each row or joined)
  PRIMARY KEY (library_id, track_position, track_artist)
```

- **Schema home — resolved: `lml_cache.*`** (LML-owned, lifespan-bootstrapped, no alembic dance). Only LML's `/lookup` reads it — no external reader post-pivot. The `entity.*` question belongs to the deferred identity table (#1020), not here.
- **Reverse index:** btree on the normalized `(track_artist, track_title)` key, using the same `wxyc_etl.text` normalization `/lookup` uses → exact-normalized O(log n) probe. `EXPLAIN` in the PR must show index use.

## Sourcing (reuse, don't reinvent)

[...]

Issue #271, opened 2026-05-10
[V/A] Compilation track identity support in bulk-resolve-libraries


jakebromberg:
> **Status update (2026-06-23): the cited contract has landed — this is now an unblocked implementation ticket.**
>
> Both original blockers are cleared:
> - The `wxyc-shared` `api.yaml` v0.7 `kind: compilation` contract **shipped**: `BulkResolveResultKind` (`single_artist` / `compilation` / `unresolved`), `BulkResolveResult.main: ReconciledIdentity | None` (null for compilation), and `BulkResolveResult.tracks: list[BulkResolveTrackIdentity] | None` ("Set only for `kind: compilation`") are in `api.yaml` and `generated/api_models.py` (classes `BulkResolveResultKind`, `BulkResolveTrackIdentity`, `BulkResolveResult`, `BulkResolveProvenanceEntry`).
> - The main LML bulk-resolve endpoint **shipped** via #272 / PR #273. `identity/bulk_resolve.py` already returns `kind: compilation` with `tracks: []` for V/A rows — exactly the stub this ticket fills in.
>
> Remaining scope is purely **implementation** and ready to start: per-track storage (§1), the per-track matcher (§2), populating `tracks[]` instead of `[]` (§3), per-track composition rules (§4–5), and script migration (§6). The "Blocked by" entries below are struck through for history.

---

## Update 2026-07-30 — decomposed into sub-issues + corrections

This ticket is now a **parent**; the implementation is split into children.

> **Revised 2026-07-30 (post-grill).** A grill-through of the design (a) **decoupled** the cheap recall index from the expensive identity matcher (Option B), (b) **deferred Layer 2** (the identity matcher; #1020/#1021) until BS#801 pulls, and (c) folded in the now-**closed #1018** decision: `/lookup` is WXYC's operational physical-library index → **comprehensive multi-location union**, not just Case-B fallback. The DAG below and the "CDC consumer"/`entity.*` specifics in the original note are superseded by this.

**Final shape:**

| Child | Layer | Scope | Depends on | Status |
|---|---|---|---|---|
| **#1019** | 1 | Recall index `lml_cache.compilation_track_location` (per-track shelf location, all credits + `credit_role`, precomputed `artwork_url`, incremental daily build) | — | **build now** (freeze-safe) |
[...]

PR #1025, opened 2026-07-31
Add the comprehensive multi-location union to /lookup

### LINKED ISSUE #1022: [V/A] Comprehensive multi-location union in /lookup (also_available_on)
Sub-issue of WXYC/library-metadata-lookup#271. The **user-facing deliverable** — the comprehensive multi-location union in `/lookup`. Blocked by the recall index (#1019) and the contract (wxyc-shared#270).

> **Rescoped 2026-07-30** after grilling #271, and after #1018 closed. This is no longer "Case-B fallback recall." Per the #1018 resolution, `/lookup` is WXYC's operational physical-library index: it surfaces the primary match **plus** every other library shelf location for the track, even when the artist's own release already matched (Case A). Fallback → **union**.

## Problem

Two failure modes make a WXYC-held compilation/soundtrack track invisible to `/lookup` (the #271 live-repro):
- **Case A (gate):** `"tommib by squarepusher"` matches the artist's own release and stops — the comp copy is never offered, even though the primary might be checked out.
- **Case B (discovery):** `"ikebana by brian reitzell"` fires `TrackOnCompilation` but returns nothing — live Discogs keyword discovery is title-distinctiveness-dependent and silently partial on common titles / joint credits.

#1018's decision: surface **all** library locations comprehensively, own release first, comps after — built on the precomputed recall index (#1019), **not** live Discogs (which is silently partial — worse than a labelled complete list).

## Desired end state

On a **track query** (a `song` in the request) with `include_locations: true`, `/lookup` returns the normal primary result **plus** `also_available_on`: a ranked list of `LibraryLocation` entries (one per other shelf copy of the track). Fires **even when the own release matched** (Case A). Contract in wxyc-shared#270.

## Design (locked via the #271 grill)

- **Source = the recall index only.** Single indexed btree probe on `lml_cache.compilation_track_location`'s normalized `(track_artist, track_title)` key. The existing ~6-call live-Discogs probe pass on this path is **removed**, not augmented.
- **Concurrent + gated.** Launch the probe concurrently with the main resolution (not serial); gate on `song`-present **AND** `include_locations`. Album-only lookups and non-flagged callers (incl. BS enrichment / the CDC firehose) pay nothing. On a flagged track query the probe hides under the main resolution latency; a miss is microseconds.
- **All credits, ranked.** The index admits every credit (primary/featured/remixer); precision is carried by **ranking: credit-tier → title-ratio → library_id**, not by filtering.
- **Lean entries, precomputed art.** Each `LibraryLocation` carries `library_id`, shelf `artist`/`album_title`, `track_position`, `track_artist` (credited-as), `credit_role`, `discogs_release_id`, and a **precomputed `artwork_url`** — full album art for recognition with **zero live artwork/identity resolution**.
- **Flag:** request-level `include_locations` opt-in **+ a server-side kill-switch env var** (disable the union globally in an incident without a client deploy). Response byte-identical when off.
- **BS stays single-best.** Multi-location is DJ-facing only (dj-site catalog, request-o-matic); BS enrichment does not set the flag.

## Constraints

- **Touches `lookup/strategies/track_on_compilation.py`** — inside the project-32 (`lookup/`) freeze. **Freeze sign-off granted by the #32 lead (2026-07-30).** The change is additive + flag-gated (byte-identical off); no hardening-surface change.
- **Merge order:** wxyc-shared#270 merges first → LML regen → this. Regen drift-fails until the contract lands.
- **`/lookup` latency budget:** measure p50/p95 on the flagged track-query path before/after — the union must not regress the critical path (the concurrent probe should be effectively free); non-flagged and album-only paths unaffected.

## Acceptance criteria

- [ ] `include_locations: true` track queries return ranked `also_available_on` (credit-tier → title-ratio → id) sourced from the recall index; live-repro assertions pass: `"ikebana by brian reitzell"` → LiT soundtrack (id 60654); `"tommib by squarepusher"` → Go Plastic **plus** the LiT location.
- [ ] Fires even when the own release matched (Case A union), gated on `song`-present + flag; concurrent probe (`EXPLAIN` single indexed probe); no live-Discogs probe on this path.
- [ ] Album art present per location via precomputed `artwork_url`; no live artwork resolution.
- [ ] Kill-switch env var disables the union; flag-off response byte-identical.
- [ ] `/lookup` latency measured before/after — no critical-path regression; BS single-best path unchanged.
- [ ] TDD: failing recall/union tests first, then implementation.

## Related

- WXYC/library-metadata-lookup#271 — parent.
- WXYC/library-metadata-lookup#1019 — the recall index; **blocks this**.
- WXYC/wxyc-shared#270 — the `api.yaml` contract; **blocks this**.
- WXYC/library-metadata-lookup#1018 — the product decision (comprehensive union), **CLOSED**.
- #210, #237 — related V/A `/lookup` recall gaps.

jakebromberg:
## Summary

- Adds a new orchestrator-level concurrent probe (`lookup/location_union.py`) that answers "which OTHER WXYC library shelf locations carry this track?" from the LML#1019 recall index (`lml_cache.compilation_track_location`) alone — no live Discogs call on this path.
- Gated on the request's `include_locations` opt-in + a `song` present + the new server-side `LML_LOCATION_UNION_ENABLED` kill switch. `perform_lookup` launches the probe as an `asyncio` task right after `parsed` is built, racing the search pipeline (step 3) instead of running serially after it.
- Once the pipeline resolves, the ranked candidates (credit-tier `primary` > `featured` > `extra` -> title-ratio -> `library_id`) are narrowed to *every other* location by excluding whichever library ids the pipeline is already surfacing as the primary result — so the union fires even when the artist's own release already matched (Case A), not only when the pipeline came up empty (Case B).
- Adds the read side of the recall index (`get_compilation_track_locations`, mirroring `get_library_release_overrides`'s best-effort PG pattern) and `LibraryDB.get_items_by_ids` to bind a recall-index row's bare `library_id` back to a displayable shelf artist/album title.
- `LookupResponse.also_available_on` carries lean `LibraryLocation` entries with the precomputed `artwork_url`; omitted (not empty) whenever the request didn't opt in or the kill switch is off, so the response stays byte-identical to before this ticket on every other path.

## Scoping note

`lookup/strategies/track_on_compilation.py` is intentionally untouched. The issue's constraints section requires any change there to be "additive + flag-gated (byte-identical off); no hardening-surface change," which a literal removal of its live-Discogs discovery pass would violate for existing non-flagged callers (that strategy's Case-B primary-match role runs unconditionally, not gated on `include_locations`). Every acceptance criterion — including both live-repro assertions ("ikebana by brian reitzell" -> LiT soundtrack id 60654; "tommib by squarepusher" -> Go Plastic plus the LiT location) — is satisfiable via the new orchestrator-level probe alone, so this PR leaves that file's behavior unchanged pending explicit confirmation that a deeper rework was actually intended.

## Test plan

- [x] `ruff check .` / `ruff format --check .` clean
- [x] `mypy . --ignore-missing-imports` clean
- [x] `pytest` (unit + integration, `pg` included) — all green, no regressions
- [x] New failing-first unit tests for: the recall-index read helper, `LibraryDB.get_items_by_ids`, `lookup/location_union.py`'s gate/rank/exclude logic, and orchestrator-level acceptance tests (flag-off byte-identical, kill-switch-disables, Case A live-repro, Case B live-repro)
- [x] New `pg`-marked integration tests for the read helper against a real PostgreSQL connection

Closes #1022


jakebromberg:
**Review gate (ship-dag Opus pass): clean on correctness/security; merging on green.** Verified load-bearing items: the union probe is genuinely concurrent (task created before the search pipeline await), gated independently of `song_not_found` so it fires in Case A, parameterized SQL using the identical `wxyc_etl.text.to_match_form` normalization the #1019 index was built with, correct primary-location exclusion, precomputed artwork (zero live artwork/identity resolution), and both the request flag and the server kill-switch produce a flag-off response byte-identical to `main`.

**One CONFIRMED deviation (tracked as a follow-up, not blocking here):** the locked design (#271 grill) said the ~6-call live-Discogs `TRACK_ON_COMPILATION` discovery pass on this path is **removed, not augmented**. This PR adds the recall-index union *alongside* it — `lookup/strategies/track_on_compilation.py` is unchanged, so on a flagged track query the orchestrator still runs the live pass concurrently with the index probe. Consequence: the Case-B primary match still comes from the unreliable live search, and the ~6 Discogs calls aren't eliminated. Decision (maintainer): ship the additive `also_available_on` now; remove the live pass + make the index drive the Case-B primary in a scoped follow-up with the before/after latency + recall validation the ticket calls for.

**Non-blocking nits (folded into the follow-up):**
1. `lookup/location_union.py` — the final `LibraryLocation(**)` construction/sort in `resolve_also_available_on` is outside the try/except, so the "never raises → degrades to []" contract relies on DDL constraints (credit_role CHECK, NOT NULL) rather than code; a future schema relaxation could turn a recall-index data defect into a hard `/lookup` 500. Wrap it.
2. `/lookup/bulk` pins `warm_cache` off per item but not `include_locations`; pin it off too, per the bulk-amplification-guard convention (no current caller sets it, so not live harm).
3. Docs (`config/settings.py`, `docs/env-vars.md`) say the flag-off `also_available_on` is "omitted", but FastAPI serializes it as `null` (no `response_model_exclude_none`); byte-identical to main, but the wording is inaccurate.
4. The orchestrator-level Case-A/Case-B repro tests mock `resolve_also_available_on`; the real probe→rank→shelf-join is covered by the integration test. Consider one end-to-end assertion through `perform_lookup`.

Issue #1052, opened 2026-08-01
Non-library (rowless) playcuts get no verified Spotify link on the bulk enrichment path


jakebromberg:
## Parent

Sub-issue of [WXYC/library-metadata-lookup#830](https://github.com/WXYC/library-metadata-lookup/issues/830) (Epic: Comprehensive verified streaming-link coverage). The epic's existing children (#831 Spotify drain, #832 Bandcamp drain, #833 YouTube Music spike) all warm the **library catalog** via offline drains. This is the sibling gap for **non-library** playcuts, which the drains cannot reach by construction.

> **Scope reconciled 2026-08-02** after an empirical re-check against the code + prod mint data (see the re-check comment below). The original framing led with Bandcamp; the prod data inverted that. **This ticket is now: Spotify parity on the bulk enrichment path, implemented as Shape 1.** Bandcamp is **out of scope** (deferred to the offline pipeline / #832 — near-zero runtime yield); Apple is **already covered** by the existing synchronous probe. The earlier Shape-2 "warmer-decoupled" design-decision comment is **superseded** — its mechanism (`refresh-for-identities`, #525) is Discogs-only and can never write a streaming URL.
>
> **Implementation-review round 2026-08-03** folded in five findings (verified against `main` @ 926c968): where the exemption seam actually lives (not the bulk handler), a dedicated bulk-rowless kill-switch, the blank-album pre-flight measurement, the BS#642 backfill rollout gate, and the AC-1 timing clarification (the triggering row never heals — BS#1747).

## Problem

[...]

Issue #548, opened 2026-06-13
feat(cache): wire bandcamp leg into POST /api/v1/cache/refresh-for-identities (LML#525 follow-up)


jakebromberg:
> **Re-scoped 2026-06-21.** Piece 1 of the original body ("build a PG-backed Bandcamp album cache") is **done** — [#573](https://github.com/WXYC/library-metadata-lookup/issues/573) PR-3 supplied the LML-owned substrate (`lml_cache.album_streaming_url_cache`, `service='bandcamp'`). This issue is now narrowed to the **offline pre-warmer**: the `bandcamp` leg of `POST /api/v1/cache/refresh-for-identities` that fills that cache ahead of live traffic. Do **not** build new cache infra.

## Context

[LML#525](https://github.com/WXYC/library-metadata-lookup/issues/525) introduced `POST /api/v1/cache/refresh-for-identities`, a source-agnostic dispatcher that warms LML's per-source release caches from `entity.release_identity` provenance. `bandcamp` is currently in [`_NOT_IMPLEMENTED_SOURCES`](https://github.com/WXYC/library-metadata-lookup/blob/1699c0d341a8f2e79a3e5c1908b743bd27a4959c/cache/dispatch.py#L178), so `_dispatch_source` returns `release_outcome = "not_implemented"` for it — only `discogs_release` runs a real leg.

Since then, [#573](https://github.com/WXYC/library-metadata-lookup/issues/573) PR-3 landed the **live** Bandcamp cache entry: a full [`STREAMING_URL_CACHE_CONFIG["bandcamp"]`](https://github.com/WXYC/library-metadata-lookup/blob/1699c0d341a8f2e79a3e5c1908b743bd27a4959c/lookup/streaming_url_postprocess.py#L157) backed by [`lml_cache.album_streaming_url_cache`](https://github.com/WXYC/library-metadata-lookup/blob/1699c0d341a8f2e79a3e5c1908b743bd27a4959c/entity/streaming_url_cache.py#L93), populated read-through during `/lookup`. The substrate this issue writes into therefore already exists; what's missing is the dispatcher leg that pre-warms it offline.

[...]

Issue #1069, opened 2026-08-01
Bandcamp: discover label/imprint/VA-hosted releases the artist-subdomain drain can't reach


jakebromberg:
## Problem

The Bandcamp full-matching drain (#832) and the runtime post-process probe (LML#573) both resolve **artist-subdomain-first**, and that design structurally cannot reach releases hosted on a **label / reissue-imprint / VA-compilation** Bandcamp page — even when the album is live on Bandcamp and a plain search finds it instantly.

The two-phase resolution in `scripts/bandcamp_pipeline.py` → `clients/bandcamp.py`:

1. **Phase 1 (slug discovery):** `BandcampClient.search_artist(artist)` hits the autocomplete API with `param="b"` — **band results only** (`clients/bandcamp.py:141`, and `:150` drops any `item["type"] != "b"`) — and keeps a hit only if `fuzz.token_sort_ratio(artist, result_name) >= MATCH_THRESHOLD_ARTIST` (75) (`scripts/bandcamp_pipeline.py:41,154-155`).
2. **Phase 2 (album match):** scrape `{slug}.bandcamp.com/music` — the artist's **own** catalog — and fuzzy-match the album title.

If a release lives on a label's subdomain rather than the artist's, Phase 1 finds no band whose name token-matches the artist ≥75, no slug is minted, Phase 2 never runs, and `/lookup` falls through to the synthesized search-URL floor (`lookup/enrichment/item.py:660-663`).

### Concrete example

- **George Theodorakis — "The Rules of the Game: Original Studio Recordings (1978–1996)"** is live at `https://into-the-light.bandcamp.com/album/the-rules-of-the-game-original-studio-recordings-1978-1996` — *Into the Light* is a Greek reissue **label**, not the artist's page.
- `/lookup` currently surfaces `https://bandcamp.com/search?q=George%20Theodorakis%20The%20Rules%20of%20the%20Game...` — the deferred search-URL fallback.
- The irony: that fallback URL lands on a Bandcamp search-results page that **contains the correct album**. The pipeline synthesizes a link to the answer it declined to scrape.

This is a **class** of misses (anything on a label/imprint/VA Bandcamp), not a one-off, and it is currently masked by search-URL fallbacks so it does not show up as a "miss" in coverage counts.

## Desired end state

[...]

PR #1087, opened 2026-08-03
feat(streaming): enable Bandcamp warm on /lookup/bulk behind a default-off flag

jakebromberg:
## Problem

The iOS flowsheet metadata path is: BS CDC enrichment worker → LML `POST /api/v1/lookup/bulk` (`extended=true`, `allowReleaseResolutionFallback=true`). On the **bulk** path LML could not resolve Bandcamp **at all** for a cold / non-library / rotation album, for two structural reasons:

1. The bulk handler pinned `bandcamp=None` (no Bandcamp client injected into `perform_lookup`).
2. The bulk handler calls `set_suppress_streaming_warm(True)`, so `apply_streaming_url_postprocess` runs cache-**read-fill only** — on a miss it neither probes nor schedules a live warm.

Net effect: a cold bulk lookup returned `bandcamp_url = null`, and Backend-Service synthesized a `bandcamp.com/search?q=` **search** fallback that the app displayed instead of the direct album URL. (The Bandcamp resolver itself is fast and correct: `BandcampClient.find_album_match("Olof Dreijer","Loud Bloom")` returns the exact URL in ~1.4s at 100% confidence.)

## Change

Adds a dedicated, **default-off** kill switch, `lml_bulk_bandcamp_streaming_warm`, AND-gated with the existing `lml_persist_streaming_urls` (master) and `lml_persist_streaming_url_bandcamp` (per-service) flags. When **all three** are on:

- **`lookup/router.py`** (`handle_bulk_lookup`) injects the Bandcamp client (`Depends(get_bandcamp_client)`) and forwards it to `perform_lookup` — `bandcamp=bandcamp if settings.lml_bulk_bandcamp_streaming_warm else None`. With the flag off it forwards `None`, byte-identical to today.
- **`lookup/streaming_url_postprocess.py`** exempts **Bandcamp only** from the bulk warm suppression via a new `_bulk_bandcamp_warm_exempt(service, settings)` helper: the `elif suppress_warm:` branch becomes `elif suppress_warm and not _bulk_bandcamp_warm_exempt(...)`, so a genuine Bandcamp miss falls through to the existing bounded, deduplicated, off-response-path background warm.

The warm resolves the **direct** album URL and writes `lml_cache.album_streaming_url_cache`, so **subsequent** bulk lookups of the same normalized `(artist, album)` read-fill it. The **triggering** response still returns `null` by design — the warm is off the response path. **Apple and Spotify bulk behavior is unchanged either way.**

### Flag

| Flag | Default | Role |
|---|---|---|
| `lml_persist_streaming_urls` | True | master kill switch (existing) |
| `lml_persist_streaming_url_bandcamp` | False | per-service Bandcamp (existing) |
| **`lml_bulk_bandcamp_streaming_warm`** | **False** | **new** — bulk-path Bandcamp warm exemption |

Bandcamp warms on bulk **only when all three are true**. Flipping the new flag off restores exactly today's bulk behavior.

### Why default-off (BS#642)

An active prod backfill drain (**BS#642**, ~748k rows at ~50/min) drives this same bulk batcher. Bandcamp is rate-limited to **1 req/s** and shares the `LML_STREAMING_WARM_CONCURRENCY` pool with the interactive-path warms, so enabling Bandcamp bulk warms **during** that drain could flood the limiter and starve interactive warms — exactly the starvation the bulk suppression was built to prevent. Shipping dark lets us enable it **only after** the drain completes, watching the `cache_miss_enqueued` / warm counters, and flip it off instantly if it misbehaves without touching the proven interactive path.

### Reconciliation with #1052 / #548

This **extends** [#1052](https://github.com/WXYC/library-metadata-lookup/issues/1052)'s Shape 1 to include Bandcamp. #1052 re-scoped itself to **Spotify-only** on the bulk path and **deliberately kept the `bandcamp=None` pin**, citing [#548](https://github.com/WXYC/library-metadata-lookup/issues/548)'s finding that runtime Bandcamp warming had near-zero identity-reconciliation yield (~10 mints/month even where the interactive warm already runs). The new justification for including Bandcamp is orthogonal to that identity-yield lens:

- the resolver is **fast** (~1.4s, well under the 9s Bandcamp warm ceiling), and
- a missing **direct** Bandcamp URL is a **visible user-facing symptom** on brand-new / rotation plays (the freeform-station core case), not just an identity statistic.

This PR does **not** touch #1052's Spotify scope; the change is Bandcamp-specific and flag-gated. Relates to #1052 (does not close it — this is a scope extension).

## Tests (TDD)

New unit tests in `tests/unit/test_lookup_streaming_url_postprocess.py` (`TestBulkBandcampWarmExemption`):

- `test_flag_on_bandcamp_miss_enqueues_warm_despite_suppression` — flag on + suppressed + genuine miss → warm enqueued, `cache_miss_enqueued` projected, returns `unresolved`.
- `test_flag_off_bandcamp_stays_suppressed_and_inert` — flag off + suppressed + Bandcamp injected → no probe, no warm (`cache_miss_unwarmed`).
- `test_flag_on_does_not_exempt_apple_or_spotify[apple_music_album|spotify_album]` — the exemption is Bandcamp-only; Apple/Spotify stay suppressed.
- `test_flag_only_consulted_under_suppression` — interactive path (no suppression) still warms Bandcamp regardless of the flag.

New / updated router tests in `tests/unit/test_bulk_lookup_endpoint.py`:

- `test_bulk_forwards_bandcamp_client_when_flag_on` — flag on → `perform_lookup` receives the injected client.
- `test_bulk_omits_bandcamp_client_when_flag_off` (renamed from `test_bulk_does_not_forward_bandcamp_client`) — flag off → `bandcamp=None`.

## Local checks

- `uv run --no-sync pytest tests/unit -q` → **5015 passed**
- `ruff check` / `ruff format --check` → clean
- `mypy config/settings.py lookup/streaming_url_postprocess.py lookup/router.py --ignore-missing-imports` → clean

Note: the repo-wide "Codegen Freshness" check is failing on `main` due to unrelated upstream `api.yaml` drift (`discogsUnavailable` family) and is not affected by this PR.

Relates to #1052


jakebromberg:
## Review fix: scoped the warm exemption to rowless items only (commit ad301f1)

Addresses the review finding that `_bulk_bandcamp_warm_exempt` was too broad — it exempted **any** bulk Bandcamp cache-miss, not just rowless/non-library ones. Because the #1069 drain put direct Bandcamp URLs on only ~9k of ~65k library albums, most library-album bulk lookups miss the PG bandcamp cache and would each have enqueued a 1 req/s Bandcamp warm — a ~65k-album, ~18h serialized backfill (the exact load the `bandcamp=None` pin guarded against, and redundant since the offline drain already owns library albums). It also diverged from #1052's rowless-only Spotify precedent.

**Fix:** threaded the rowless signal through a new `is_rowless` keyword on `apply_streaming_url_postprocess`, required in the exemption. The warm now fires only for **Bandcamp AND a rowless item AND the flag on**.

- **Rowless predicate used:** `item.id == ROWLESS_LIBRARY_ID` (from `lookup/rowless.py`, `ROWLESS_LIBRARY_ID == 0`) — the existing predicate `enrich_one` already branches on (e.g. `lookup/enrichment/item.py:104`, the librarian-override gate). Not a new signal.
- **Threading:** `lookup/enrichment/item.py` passes `is_rowless=item.id == ROWLESS_LIBRARY_ID` at the post-process call site (where `item` is in scope), exactly as #1052's plan prescribed (the rowless bit is unknowable in the router).
- Library rows stay suppressed on bulk (cache read-fill only), unchanged. Flag, AND-gate with the master + per-service kill switches, flag-off equivalence, and the BS#642 rollout gate are all intact.

**New/updated tests:**
- `test_lookup_streaming_url_postprocess.py::TestBulkBandcampWarmExemption` — `test_flag_on_rowless_bandcamp_miss_enqueues_warm_despite_suppression` (rowless → warms), `test_flag_on_nonrowless_library_item_stays_suppressed` (**new guard**: library id != 0 → suppressed, no warm), plus the flag-off test now uses a rowless item to prove the flag dominates the rowless signal.
- `test_enrichment.py::TestBandcampStreamingUrlPostprocessWiring::test_is_rowless_threaded_from_item_id[0-True|42-False]` — **new wiring test** that `enrich_one` threads `is_rowless` correctly for both a rowless (id==0) and a library (id!=0) item.

Local: `pytest tests/unit -q` → 5018 passed; ruff check/format + mypy clean on changed files.

Issue #1094, opened 2026-08-04
Streaming-URL warm slot has no bulk/interactive isolation: a bulk sweep monopolizes LML_STREAMING_WARM_CONCURRENCY and 429-storms Bandcamp


jakebromberg:
## Problem

The streaming-URL warm subsystem has **no isolation between latency-sensitive interactive warms and high-volume bulk warm sweeps**. All warms — every service (Spotify / Apple / Bandcamp) and every caller of the interactive `/lookup` path — contend for one process-global semaphore sized by `LML_STREAMING_WARM_CONCURRENCY`, which prod pins to **1** (the code default is 4). When a bulk sweep drives a sustained stream of warm tasks, it monopolizes that single slot: a freshly-enqueued interactive warm head-of-line-blocks behind the backlog for **minutes**, and the sweep's request volume drives **sustained Bandcamp 429 rate-limiting** that further ties up the slot with 3s-backoff retries.

This is the warm-slot analog of the Discogs-semaphore contention already fixed for the *query* path by [#927](https://github.com/WXYC/library-metadata-lookup/issues/927) / [#924](https://github.com/WXYC/library-metadata-lookup/issues/924) (class-aware priority so interactive never head-of-line-blocks behind bulk). The warm semaphore never received that treatment. [#1052](https://github.com/WXYC/library-metadata-lookup/issues/1052)'s own Constraints section already names this exact latent risk ("all warms share `LML_STREAMING_WARM_CONCURRENCY`... a backfill-rate stream of distinct rowless albums queues tasks and head-of-line-blocks interactive warms, exactly the starvation the bulk suppression was built to prevent") — this ticket extracts it into a first-class issue with live prod evidence.

**Important framing:** the Bandcamp warm leg is **not broken**. It resolves fresh URLs continuously in prod (see evidence). The defect is throughput/fairness under contention, not resolvability.

## Evidence (prod deploy logs, 2026-08-03 PT)

[...]

Issue #1040, opened 2026-08-01
refactor(discogs): decompose the 285-line _request_with_retry into an admission stack; share the 429 loop with the Bandcamp client


jakebromberg:
## Problem

`discogs/service.py::_request_with_retry` is a ~285-line method folding six concerns, interleaved line-by-line: (a) breaker admit/shed, (b) live-request telemetry counter, (c) bulk sub-semaphore acquire/release + its Sentry projection (#927), (d) shared-semaphore acquire + queue-depth projection, (e) rate-gate acquire (#879), (f) the 429 retry loop with retry-budget-deadline arithmetic and terminal breaker `record_*` accounting. Every incident fix (breaker #755, rate gate, bulk reservation) edits the same method — high collision and review cost.

The layer is also not reusable: `clients/bandcamp.py` re-implements its own 429 loop (`_request_with_retry`, `for attempt in range(...)` with `RETRY_BASE_DELAY * (2**attempt)` — no jitter, no cap, own `Retry-After` parse), parallel to `service.py::_compute_retry_delay` (jittered, 60s-capped).

## Desired end state

A composable admission stack, behavior-preserving:

```python
async with breaker.admit(), acquire_discogs_permits(bulk_reserved, shared), rate_gate.acquire():
    return await retry_429(request_fn, compute_delay=..., budget_deadline=...)
```

- Each permit context manager owns its own Sentry projection (queue-depth, bulk-reserved-capped) internally.
- `retry_429` is the one shared 429 loop; `clients/bandcamp.py` adopts it with its own delay policy (keeping Bandcamp's current constants unless a change is explicitly approved).
- Zero telemetry-key changes; zero ordering changes in acquire sequence (breaker → bulk sub-semaphore → shared semaphore → rate gate — verify exact current order from the code before assuming this listing).

## Constraints (read before starting)

- This is the hottest, most incident-hardened path in the service, squarely inside the Post-launch service hardening project's "stop adding features in the touched files" zone (org project #32). Execute as dedicated, mechanically-verified refactor PR(s) with the existing breaker/ratelimit/live-counter suites as the harness — **or explicitly defer until #32 winds down**. Do not bundle with feature work.
- The `locked()`-precheck and cancellation-safety comments in the current body are load-bearing documentation — port them with the code they describe.
- The retry-budget-deadline contextvar seam (`set_retry_budget_deadline` / `reset_retry_budget_deadline`, consumed from `core/search.py`) must keep its exact semantics.

[...]

PR #1113, opened 2026-08-04
fix(streaming): bound the Spotify 429 retry sleep by the warm probe's remaining budget

### LINKED ISSUE #1108: Streaming warm probe cannot succeed under Spotify 429: 5s minimum Retry-After sleep exceeds the 4s probe budget
## Problem

Once Spotify starts returning HTTP 429, **every** background streaming-URL warm probe for `spotify_album` is guaranteed to fail. This is not a tail-latency or fairness problem — the two deadlines are arithmetically incompatible:

- The Spotify client's 429 handler sleeps for the full `Retry-After`, with a **floor of 5 seconds**: [`clients/streaming/spotify.py#L128-L138`](https://github.com/WXYC/library-metadata-lookup/blob/d997b0fcab74ba61475a75c64e3fdadf834a95a0/clients/streaming/spotify.py#L128-L138) — `retry_after = int(resp.headers.get("Retry-After", "5"))` then `await asyncio.sleep(retry_after)`.
- The warm probe that wraps it is bounded at **4.0 seconds**: [`lookup/streaming_url_postprocess.py#L217`](https://github.com/WXYC/library-metadata-lookup/blob/d997b0fcab74ba61475a75c64e3fdadf834a95a0/lookup/streaming_url_postprocess.py#L217) (`_DEFAULT_PROBE_TIMEOUT_S = 4.0`), applied at [`#L564-L565`](https://github.com/WXYC/library-metadata-lookup/blob/d997b0fcab74ba61475a75c64e3fdadf834a95a0/lookup/streaming_url_postprocess.py#L564-L565).

5s minimum sleep against a 4s budget means the probe is cancelled mid-sleep 100% of the time. The client's own log line formats the delay in hours (`hours = retry_after / 3600`), so multi-hour `Retry-After` values are expected in practice — the gap is not marginal.

Each doomed probe still holds a warm-semaphore permit for the full 4s before dying, so under sustained 429s the warm slot is consumed entirely by probes that cannot succeed.

## Evidence

[LIBRARY-METADATA-LOOKUP-1B](https://wxyc.sentry.io/issues/LIBRARY-METADATA-LOOKUP-1B) — `TimeoutError`, **33,095 lifetime occurrences, 11,724 in the last 7 days**, still firing at 2026-08-04 08:08 PDT. It is roughly 80% of all WXYC org Sentry error volume in the window.

The captured event proves the mechanism directly:

- Culprit frame is `lookup/streaming_url_postprocess.py`, line 565, in `_warm_streaming_url_cache`.
- Local variable in that frame: `semaphore: <asyncio.locks.Semaphore [locked, waiters:232]>` — **232 queued warm tasks**.
- The chained `CancelledError` unwinds through `clients/streaming/spotify.py`, line 138, `await asyncio.sleep(retry_after)` — i.e. the probe deadline fired while the client was in its 429 back-off sleep.
- Every event carries `lml.caller_reason: flowsheet-metadata-backfill` and `lml.low_priority: true`.

## Root cause, in two parts

**1. Deadline inversion (the correctness defect).** Spotify is the only 429 path in the codebase that never migrated to the shared retry helper. `discogs.admission.retry_429` ([`discogs/admission.py#L160`](https://github.com/WXYC/library-metadata-lookup/blob/d997b0fcab74ba61475a75c64e3fdadf834a95a0/discogs/admission.py#L160)) is used by `clients/bandcamp.py` (via LML#1040) and `discogs/service.py`, both of which bound their delay. `clients/streaming/spotify.py` does not import it and open-codes a raw unbounded `asyncio.sleep`. It has no awareness of the caller's remaining budget.

**2. Unbounded warm queue depth (the amplifier).** [`_enqueue_streaming_warm`](https://github.com/WXYC/library-metadata-lookup/blob/d997b0fcab74ba61475a75c64e3fdadf834a95a0/lookup/streaming_url_postprocess.py#L509) calls `asyncio.create_task` with no admission check on queue depth. The semaphore bounds *concurrency* but nothing bounds *how many tasks may be waiting for it*. The semaphore is also acquired **outside** the `wait_for` (`async with semaphore:` at L564 encloses the `wait_for` at L565), so queue wait time is not covered by any deadline. Per WXYC/library-metadata-lookup#1094 prod pins `LML_STREAMING_WARM_CONCURRENCY=1` (code default is 4) — at 1 permit and 4s per doomed probe, a 232-deep queue is a ~15-minute backlog of work that will all fail.

## Relationship to #1094 — sibling, not duplicate

WXYC/library-metadata-lookup#1094 covers the same semaphore and also discusses 429 storms, but it is scoped to **bulk/interactive fairness for Bandcamp**: its thesis is explicitly that "the Bandcamp warm leg is not broken… the defect is throughput/fairness under contention, not resolvability."

This ticket is the opposite claim for a different service: under Spotify 429s the warm leg **is** broken, unconditionally, and no amount of class-aware priority or slot isolation fixes it. #1094's fairness work would give interactive Spotify warms a clean slot in which they would still fail. The two should land together but neither subsumes the other.

## Desired end state

A Spotify warm probe under 429 either completes within its budget or is abandoned *without burning a permit for the full budget*, and the warm queue cannot grow without bound.

## Acceptance criteria

- [ ] The Spotify 429 back-off is bounded by the caller's remaining deadline — a sleep that cannot fit in the remaining budget aborts immediately rather than sleeping into a guaranteed cancellation.
- [ ] Preferably: `clients/streaming/spotify.py` delegates its 429 loop to `discogs.admission.retry_429`, matching the LML#1040 treatment of `clients/bandcamp.py`. If a bespoke loop is kept, document why.
- [ ] The warm-probe timeout and the client's minimum back-off can no longer be configured into a state where the former is unconditionally shorter than the latter — either enforce the invariant at startup or derive one from the other.
- [ ] `_enqueue_streaming_warm` bounds pending-task depth; enqueues beyond the bound are shed and counted rather than queued. Sensible bound is a small multiple of `LML_STREAMING_WARM_CONCURRENCY`.
- [ ] Time spent waiting for the warm semaphore is either covered by a deadline or observable — a probe should not be able to sit unbounded before its 4s clock starts.
- [ ] Unit tests: (a) a 429 with `Retry-After` exceeding the remaining budget aborts without sleeping the full interval; (b) enqueueing past the depth bound sheds rather than queues.
- [ ] Sentry volume on LIBRARY-METADATA-LOOKUP-1B drops to near zero while `flowsheet-metadata-backfill` is running.

## Constraints

- The warm path is fire-and-forget by design — every exception must stay swallowed and must never propagate to the event loop (see the module docstring's "Side effects degrade gracefully" contract). Shedding must not change the `/lookup` response shape.
- Do not widen `_DEFAULT_PROBE_TIMEOUT_S` past the 429 floor as the sole fix. That trades a guaranteed-fail for a slow-fail and makes the head-of-line blocking in #1094 worse.
- `lookup/orchestrator.py` and `lookup/artwork.py` are frozen under the [Post-launch service hardening](https://github.com/orgs/WXYC/projects/32) project; this work should not need to touch either.

## Notes for implementer

The traffic driver is Backend-Service's `flowsheet-metadata-backfill` cron, running hourly at `10 * * * *` UTC. It is **supposed** to be running — do not read WXYC/Backend-Service#1064's "keep the cron stopped" stop-gap as current. That applied to the job's former role as the nightly historical drain, which completed and was retired under WXYC/Backend-Service#1011. The job was re-launched on 2026-07-28 in a new role as the Epic C C6 hourly gap-recovery sweep (WXYC/Backend-Service#895), and its schedule moved from the never-fires sentinel `0 0 31 2 *` to `10 * * * *` in Backend-Service commit `8f0fc79e`.

What this reframes: the sweep is a **safety net** that should only pick up rows the CDC enrichment consumer missed inside a 6h window — a small hourly worklist, not a bulk drain. 11,724 warm-probe timeouts in 7 days from a job of that shape is disproportionate to its intended volume, which makes the unbounded queue depth in part 2 above more suspicious rather than less. Worth measuring warm enqueues per row processed as part of the fix.

## Related

- WXYC/library-metadata-lookup#1094 — warm-slot bulk/interactive isolation (sibling; same semaphore, different defect)
- WXYC/library-metadata-lookup#1052 — rowless playcuts get no verified Spotify link (likely a downstream symptom of this)
- WXYC/library-metadata-lookup#1040 — the shared `retry_429` migration that Bandcamp received and Spotify did not
- WXYC/library-metadata-lookup#903 — precedent for taking a probe off the critical path with a bounded deadline
- WXYC/Backend-Service#1064 — the backfill cron's ~21% LML timeout rate, viewed from the caller side

jakebromberg:
## Problem

Once Spotify starts returning HTTP 429, every background streaming-URL warm probe for `spotify_album` was guaranteed to fail. The two deadlines were arithmetically incompatible:

- Spotify's 429 handler (`clients/streaming/spotify.py`) slept for the full `Retry-After`, floored at 5s when the header is absent, with no knowledge of the caller's remaining budget.
- The background warm probe (`lookup/streaming_url_postprocess.py`) wraps every client call in `asyncio.wait_for(timeout=4.0)`.

5s minimum sleep against a 4s budget meant the probe was cancelled mid-sleep on every attempt. This produced `LIBRARY-METADATA-LOOKUP-1B` — 33k+ lifetime `TimeoutError` occurrences, ~80% of all WXYC org Sentry volume in the affected window — and each doomed probe held the warm-concurrency semaphore permit for the full ceiling before dying.

## Fix

- `clients/streaming/base.py` exposes a shared probe-deadline `ContextVar` (`set_probe_deadline` / `get_probe_deadline` / `reset_probe_deadline`), mirroring `discogs.service`'s caller-budget-deadline pattern (LML#758) but scoped to the streaming-client contract so any client can opt in without the post-process needing a client-specific import.
- `lookup/streaming_url_postprocess.py`'s `_warm_streaming_url_cache` arms that deadline with, at or just before, the value `wait_for` will enforce, right before the probe call, and resets it in a `finally`. The semaphore-acquire wait and the probe itself are timed and exception-handled separately, so a probe failure is never misattributed in the log as a semaphore-wait failure.
- `clients/streaming/spotify.py`'s 429 **and** 5xx branches both consult the deadline (via a shared `_delay_exceeds_probe_budget` helper) and give up — **by raising `SpotifyRetryBudgetExceededError`, never by returning `[]`** — when the next planned sleep can't fit in what's left. The give-up is binary (abort immediately, no partial "clamp" sleep), the same shape `discogs.admission.retry_429`'s own LML#758 budget-deadline check uses. Outside a warm probe (interactive `/lookup`, offline scripts), the deadline is `None` and the pre-existing attempt-count-only bound is unchanged.

Why the exception matters, not just the deadline check: returning `[]` on a give-up reads as a confirmed "no match" to every caller. `search_album` and `search_track` re-raise `SpotifyRetryBudgetExceededError` past their own broad `except Exception` (which would otherwise swallow it back into that same `[]`), so it propagates out of `find_album_match` to `entity.streaming_url_cache.resolve_streaming_url_with_cache`'s existing `except Exception`, which degrades to `live_error` and — critically — writes **no** cache row. An early draft of this fix returned `[]` on give-up instead, which durably UPSERTed a `url=None` row with a 7-day `miss_ttl` for albums that were merely rate-limited, not confirmed absent — trading the old guaranteed-fail (loud, but safe: a `CancelledError` is a `BaseException` and always wrote nothing) for a guaranteed silent false negative. `TestBudgetGiveupDoesNotPoisonCache` exercises the real `resolve_streaming_url_with_cache` integration point end to end and asserts zero PG writes; it's mutation-verified against that exact regression. Raising rather than returning `[]` also means `search_album`'s unquoted fallback no longer fires on a give-up — a second live request to a service that just signaled backoff.

The retry loop stays bespoke rather than delegating fully to the shared `discogs.admission.retry_429` (the LML#1040 precedent `clients/bandcamp.py` followed): Spotify's loop also retries 5xx with its own exponential backoff sharing the same attempt budget, and `retry_429` returns immediately on any non-429 status including 5xx. Full delegation would mean either dropping that 5xx resilience or nesting an independent sub-loop that could multiply total attempts under a mixed 429/5xx failure mode — a bigger, riskier change than this deadline-inversion fix calls for. Documented inline.

## Also closes the two named amplifiers

- **Unbounded warm queue depth.** `_enqueue_streaming_warm` now sheds an enqueue (rather than queueing it) once the pending-plus-running warm count reaches a multiple of `LML_STREAMING_WARM_CONCURRENCY` — LIBRARY-METADATA-LOOKUP-1B showed a 232-deep backlog. A shed still reports the same `unresolved` status a healthy enqueue would (the `/lookup` response shape is unchanged) but projects a distinguishable Sentry outcome (`cache_miss_shed`) and a cache-stats counter (`streaming_warm_depth_shed`, seeded into `lookup/router.py`'s `_LML_CACHE_STATS_EXTRA_KEYS` so the shed rate is an alertable baseline series). This policy is extracted into a new `lookup/streaming_warm_admission.py` module to keep `streaming_url_postprocess.py` under its module-budget guardrail (recalibrated 750 → 800 for the review's split-exception-handling requirement).
  - The multiplier defaults to **16** (up from an initial 4) and is env-tunable via `LML_STREAMING_WARM_QUEUE_DEPTH_MULTIPLIER`, read fresh per call. 4 was too tight against prod's real shape: at the prod `LML_STREAMING_WARM_CONCURRENCY=1` pin, a bound of 4 sheds under two ordinary concurrent cold `/lookup` calls (each can enqueue up to 3 keys — Apple + Spotify + Bandcamp), not just a storm. 16 gives headroom for roughly 5 concurrent 3-key cold lookups while still being a >14x reduction from the incident peak; combined with the deadline fix (a doomed warm now fails near-instantly instead of holding its permit for the full ceiling), even a genuine storm should drain this depth far faster than the incident did. WXYC/library-metadata-lookup#1094 forbids raising `LML_STREAMING_WARM_CONCURRENCY` itself as a runtime lever, which makes this multiplier the only no-redeploy lever left for the depth bound — hence env-tunable rather than a fixed constant.
- **Unobserved semaphore queue wait.** The warm now times its own wait to acquire the concurrency semaphore and folds it into the outcome log line, so a growing queue is visible without adding a second hard deadline (which would start to overlap WXYC/library-metadata-lookup#1094's fairness scope).

## Scope notes

- Does not widen `_DEFAULT_PROBE_TIMEOUT_S` — that would trade a guaranteed-fail for a slow-fail, which the issue explicitly rules out.
- Does not touch `lookup/orchestrator.py` or `lookup/artwork.py` (frozen under the Post-launch service hardening project).
- Does not implement WXYC/library-metadata-lookup#1094's bulk/interactive fairness or priority-lane work — that ticket is a sibling covering the same semaphore for a different defect (throughput/fairness under contention, not the arithmetic guaranteed-fail this ticket fixes). The depth bound added here is orthogonal (a hard cap + shed, not fairness) and should compose with #1094's future priority-lane work rather than collide with it.
- On the "invariant enforcement" acceptance criterion (warm-probe timeout vs. the client's minimum backoff): rather than a static startup assertion — which would either be redundant given the dynamic bound, or wrong-headed (asserting `probe_timeout >= floor` would effectively re-impose the widen-the-timeout anti-pattern) — the deadline-bound retry loop satisfies this dynamically. To be precise about the mechanism (corrected from an earlier draft of this description): it is a **binary give-up**, not a fractional clamp — a sleep either fits inside the remaining budget in full, or the loop aborts before sleeping at all. It does not sleep a reduced amount and retry.

## Tests

TDD throughout, including two rounds of mutation-verification (temporarily reverting the fix under test and confirming the new test fails). New/updated:
- `tests/unit/test_spotify_retry_budget_deadline.py` — 429 and 5xx give-ups raise `SpotifyRetryBudgetExceededError` (deadline exceeded / already passed / a realistic multi-step backoff sequence that would have reproduced the incident signature), proceed normally with enough budget, ignore the budget outside a warm probe; `search_album`/`search_track`/`find_album_match` all propagate the give-up past their own broad exception handlers; `search_album` does not run its unquoted fallback on a give-up; `TestBudgetGiveupDoesNotPoisonCache` exercises the real cache-resolution integration point and asserts zero PG writes.
- `tests/unit/test_lookup_streaming_url_postprocess.py` — `TestWarmQueueDepthBound` (shed at the bound, succeeds one below it, dedup hits aren't sheddable, distinct Sentry outcome + cache-stats counter, env-tunable multiplier) and `TestProbeDeadlinePropagation`, rewritten to observe from inside the same warm task (`asyncio.create_task` copies the calling context, so a parent-scope read after the task completes was trivially `None` regardless of whether the reset actually ran).
- `tests/unit/test_lookup_router.py` — the depth-shed stat key is seeded via the real `init_cache_stats`/`get_cache_stats()` path, not a mock.

`ruff check`, `ruff format --check`, `mypy . --ignore-missing-imports`, and the full default-tier `pytest` suite (5414 passed, 1 skipped, 2 xfailed — both pre-existing) all pass locally.

**Codegen Freshness will stay red on this PR** — unrelated pre-existing `generated/api_models.py` drift, tracked in WXYC/library-metadata-lookup#1117, not introduced or fixable here.

## Related

- WXYC/library-metadata-lookup#1094 — warm-slot bulk/interactive isolation (sibling; same semaphore, different defect — not implemented here, see Scope notes above)
- WXYC/library-metadata-lookup#1040 — the shared `retry_429` migration `clients/bandcamp.py` received (precedent for the pattern mirrored here, and for why Spotify's loop stays bespoke, and for the `BandcampRateLimitedError` shape `SpotifyRetryBudgetExceededError` mirrors)
- WXYC/library-metadata-lookup#1117 — the unrelated pre-existing Codegen Freshness drift

Closes #1108


jakebromberg:
The red **Codegen Freshness** check on this PR is not caused by this change. It is pre-existing drift on `main`: `generated/api_models.py` lags `wxyc-shared` api.yaml 1.29.0 after commit `cae6030` (BS#1965 CTA export) merged 2026-08-03. The same check is red on the unrelated WXYC/library-metadata-lookup#1110 for the same reason.

Tracked in WXYC/library-metadata-lookup#1117 — a standalone regeneration. This PR should go fully green once that lands and this branch is rebased. Every other check here (Default Tests, External API Tests, Lint & Format, PG Tests, Type Check, CI Marker Sync) is already passing.

Issue #1116, opened 2026-08-04
Bandcamp breaker coverage parity with Discogs: default-mode traffic, in-flight re-check, stale-outcome cool-down, trial sizing, env-tunability


jakebromberg:
> **Rescoped 2026-08-04**, after #1106/#1098 merged (PR #1110) and four rounds of review changed the breaker underneath this ticket. `record_transport_failure` did not exist when this was written; the trip-signal set, the watchdog sizing, and the exception hierarchy have all moved. The original three items survive — one of them with sharper teeth — and two new ones are added. Line references below are against `main` at merge.

## Problem

`BandcampProbeBreaker` observes only `find_album_match(..., fail_fast=True)` traffic. `clients/bandcamp.py:429` returns early — `if not fail_fast: return await self._find_album_match_impl(...)` — **before** `get_bandcamp_probe_breaker()` is ever called at `:432`. Every live Bandcamp caller in production today uses the default path:

- `streaming/orchestrator.py` — `POST /api/v1/streaming-check`
- `lookup/streaming_url_postprocess.py::_warm_streaming_url_cache` — the background warm
- `scripts/bandcamp_pipeline.py` — the offline drain

So during an #1094-style bulk warm sweep those callers 429-storm Bandcamp for minutes while the breaker sits CLOSED, having seen none of it. The first `fail_fast` probe is then admitted into a guaranteed 429 and only trips after three of its *own* consecutive sheds.

Discogs has no such carve-out: `discogs/service.py` wraps its single shared request path in `admit_or_shed` unconditionally.

**What changed, and why this now matters more.** #1115 (PR #1121) makes `search_artist` / `fetch_artist_catalog` / `search_albums` raise `BandcampTransportError` in the **default** mode too, so a couldn't-ask is no longer silently degraded to `[]`/`None` on the warm and `/streaming-check` paths. That is the right fix for cache poisoning, but it means default-mode callers now fail **loudly and repeatedly** during an outage with nothing shedding them. The failures became visible without becoming bounded.

### 1. The breaker observes only fail-fast traffic

[...]

Issue #1106, opened 2026-08-04
Bandcamp client fail-fast mode + live-probe saturation breaker (split from #1098)


jakebromberg:
## Parent

Split out of #1098 (Bandcamp live probe on the enrichment path), which is a sub-issue of #1100 under #830. #1098 proved too large to land in one pass; this issue carries the **client-and-cache-layer plumbing**, and #1098 retains the **enrichment-path probe** that consumes it.

The seam is where the two halves genuinely separate: everything here is testable against `BandcampClient` and `resolve_streaming_url_with_cache` in isolation, with no dependency on `enrich_one`, the router, or the response shape.

## Problem

The synchronous Bandcamp live probe #1098 introduces cannot reuse the existing Bandcamp request path as-is, for two reasons:

1. **`_request_with_retry` retries 429s with a backoff loop.** On a response-path probe that stacks a 3s+ sleep on top of the probe's own `wait_for` ceiling. The probe needs exactly one attempt.
2. **A 429 currently degrades to `None`,** which is indistinguishable from a genuine no-match. On the probe path that distinction is load-bearing: a rate-limit shed must never be written to the cache as a negative result, or a transient throttle becomes a permanently cached "this album isn't on Bandcamp".

There is also no saturation guard on the Bandcamp path. Bandcamp rate-limits at ~1 req/s and is 429-prone; a response-path probe without a breaker turns a Bandcamp throttle into latency on every enrichment request.

## Desired end state

`BandcampClient.find_album_match(..., fail_fast=True)` is a single-attempt, breaker-gated call that raises rather than silently degrading on a shed, and `resolve_streaming_url_with_cache` can forward that mode without recording a negative-cache row for a shed. No caller uses the new mode yet — #1098 is the first consumer.

## Scope

**In:**

[...]

PR #1121, opened 2026-08-04
fix(streaming): stop caching Bandcamp default-path transport failures as known-misses

### LINKED ISSUE #1115: Non-429 Bandcamp failures are cached as 7-day known-misses on the default path
## Problem

A Bandcamp request that fails for any reason **other than a 429** degrades to a clean `None`, and `resolve_streaming_url_with_cache` UPSERTs that as a known-miss row with `DEFAULT_MISS_TTL = 7 days` ([`entity/cache_toolkit.py:72`](https://github.com/WXYC/library-metadata-lookup/blob/main/entity/cache_toolkit.py#L72)). "We couldn't ask" is frozen as "not on Bandcamp" for a week.

The degrade chain, all on the default (non-`fail_fast`) path:

- `clients/bandcamp.py` — `_request_with_retry` returns the response verbatim or `None`; only a 429 under `fail_fast` raises. A 503, a Cloudflare 403/1015, a connect timeout, or any other non-200 returns normally.
- `search_albums` → `None` on non-200 · `search_artist` → `[]` on non-200 · `fetch_artist_catalog` → `None`
- `find_album_match_via_search` → `BandcampSearchUnavailableError`, which `_find_album_match_impl` **swallows** → `None`
- `resolve_streaming_url_with_cache` → `if match is None:` → `set_cached_streaming_url(..., url=None)` → `ResolveOutcome(source="live_miss")`

Note `discogs.admission.retry_429` retries **only** 429 — 5xx and transport failures are single-attempt in both modes — so a transient 503 gets exactly one shot before being cached as absent.

**Failure scenario:** a two-minute Bandcamp 503 window during a warm sweep or a `/streaming-check` batch writes every album touched as `url=NULL`, read back as `cache_miss_recent` for seven days. Those albums then serve the synthesized `bandcamp.com/search?q=` fallback to the flowsheet and to iOS for the whole TTL, and — because Backend-Service freezes the first enrichment response (the BS#1747 pattern) — a flowsheet row that lands during the window keeps the search URL indefinitely, not just for seven days.

## Scope

This is the **default-path half** of a finding split out of #1106. #1106 fixed the `fail_fast` path only, under an explicit no-behavior-change-for-existing-callers constraint; the pre-existing default-path hole was deliberately deferred here rather than widened into that PR.

Affected live callers, none of which pass `fail_fast`:

- `streaming/orchestrator.py` — `/api/v1/streaming-check`
- `lookup/streaming_url_postprocess.py` — `_warm_streaming_url_cache` (the background warm)
- `scripts/bandcamp_pipeline.py` — the offline drain

## Desired end state

A Bandcamp request that failed to *reach a verdict* is distinguishable from one that reached a verdict of "no match", on the default path as well as the fail-fast path, and only the latter writes a known-miss row. A transient outage costs a retry, not a week of cached absence.

## Suggested approach

#1106 already built the fail-fast half: `BandcampRateLimitedError` for a 429 shed, and (per that PR's review fixes) a distinguishable channel for non-429 failures that neither writes the negative row nor reports success to the breaker. The work here is extending that distinction to the default path and deciding the write policy per outcome class:

| Outcome | Live call | Cache write today | Should be |
|---|---|---|---|
| 200, no match | reached verdict | known-miss null (7d) | unchanged — correct |
| 429 | no verdict | known-miss null | no write |
| 5xx / 403 / transport | no verdict | known-miss null | no write |
| Cancellation | no verdict | no write | unchanged — correct |

Because this changes behavior for three live callers, it wants its own before/after measurement — in particular how many existing `url IS NULL` rows in `lml_cache.album_streaming_url_cache` are outage artifacts rather than genuine misses. A one-off audit re-probing a sample of null rows would size the retro-cleanup, which may be worth doing separately from the forward fix.

## Constraints

- TDD per `CLAUDE.md`.
- Changes behavior for existing callers — unlike #1106, that is the point here, so it needs its own review of each call site's failure posture.
- Do not weaken the deliberate negative caching: a genuine 200-with-no-match **must** still write the known-miss row. That TTL machinery is what keeps repeat plays of an off-Bandcamp album from re-probing on every spin, and #1098's steady-state volume estimate depends on it.
- No retro-deletion of existing cache rows without explicit approval (org data-safety rule); propose the cleanup separately from the forward fix.

## Acceptance criteria

- [ ] A non-429 Bandcamp failure on the **default** path produces no known-miss cache row.
- [ ] A genuine 200-with-no-match still writes the known-miss row with the existing TTL.
- [ ] Each of the three live callers has its failure posture reviewed and pinned by a test.
- [ ] Sized: how many current null rows are plausibly outage artifacts, and whether a retro-cleanup is worth proposing.

## Related

#1106 (fixed the fail-fast half; this is the deferred default-path half) · #1098 (the live probe that consumes the fail-fast path) · #573 / #595 (the streaming-URL cache this writes to) · #1094 (warm-slot contention — the conditions that produce 429/5xx bursts) · #832 / #1069 (the offline Bandcamp drain) · BS#1747 (why a frozen response makes a cached miss outlive its TTL downstream)

jakebromberg:
Closes #1115

## Summary

A Bandcamp request that failed for any reason other than a 429 used to degrade to a clean `None` on the default (non-`fail_fast`) path -- `search_artist` collapsed a transport failure to `[]`, `fetch_artist_catalog` and `search_albums` collapsed it to `None`, and `find_album_match_via_search`'s `BandcampSearchUnavailableError` was swallowed back to `None` inside `_find_album_match_impl`. That clean `None` then looked exactly like a genuine no-match to `resolve_streaming_url_with_cache`, which UPSERTs a 7-day known-miss row on `match is None` -- so "couldn't ask" got frozen as "not on Bandcamp" for a week.

#1106 already built a distinguishable-failure channel for this exact class of bug on the `fail_fast` path (`BandcampRateLimitedError` for a 429, `BandcampTransportError` for everything else). This PR extends that channel to the default path instead of inventing a parallel one:

- `search_artist` / `fetch_artist_catalog` / `search_albums` now raise `BandcampTransportError` on a transport failure in **both** modes (previously `fail_fast`-only).
- `_find_album_match_impl` no longer catches `BandcampSearchUnavailableError` to degrade to `None` -- a couldn't-ask outcome at any phase now propagates out of `find_album_match` uncaught, mirroring the short-circuit-without-trying-the-fallback posture `fail_fast` already had.
- `find_album_match_via_search` keeps translating a default-mode `search_albums` failure to `BandcampSearchUnavailableError` (same type, now sourced from `BandcampTransportError`) so the offline album-search drain's existing catch, and the live Bandcamp smoke test, keep working unchanged.

No code changes were needed in `entity/streaming_url_cache.py` or `streaming/orchestrator.py`: `resolve_streaming_url_with_cache`'s exception handling already skipped the cache write for any raised exception in default mode, and `check_streaming_availability`'s per-service `except Exception` already folded any raise into `errored_sources` rather than a hard failure or a false negative. Both are now exercised end-to-end by new tests using a real `BandcampClient` instead of only by generic-exception stand-ins.

The one real caller-side change is `scripts/bandcamp_pipeline.py`'s `phase_lookup`, which calls `fetch_artist_catalog` directly (not through `find_album_match`) and now catches `BandcampTransportError`, folding it back to the pre-existing `None` sentinel so a slug stays `pending` across a network blip instead of aborting the whole multi-hour drain. `phase_search`'s `search_artist` failures were already caught generically by `process_batched`'s `return_exceptions=True` and needed no change -- this incidentally fixes a latent bug where a transient search failure was previously written as a permanent `bandcamp_slug=''` not-found.

A genuine 200-with-no-match is untouched and still writes the known-miss row with its existing TTL (the deliberate negative caching the ticket says not to weaken).

## Call site failure postures (reviewed per the ticket's acceptance criteria)

| Caller | Posture | Why |
|---|---|---|
| `streaming/orchestrator.py` (`/api/v1/streaming-check`) | Degraded response: folds into `errored_sources`, escalates `on_streaming` to `None` absent other positive evidence | Pre-existing generic `except Exception` per service leg; no code change needed, pinned with a new Bandcamp-specific test |
| `lookup/streaming_url_postprocess.py` (`_warm_streaming_url_cache`, background warm) | Logged and swallowed, no mint, no cache write | `resolve_streaming_url_with_cache` already swallows any default-mode exception before returning; no code change needed, pinned end-to-end with a real `BandcampClient` |
| `scripts/bandcamp_pipeline.py` (`phase_lookup`, offline drain) | Row stays `pending`/re-runnable, `fetch_failed` tally increments | Needed a code change: `fetch_artist_catalog` no longer returns `None`, so the call site now catches `BandcampTransportError` and converts it back to the `None` sentinel the rest of the loop already handles correctly |

## Data safety

No cache rows were deleted, reset, or rewritten. Sizing how many existing `lml_cache.album_streaming_url_cache` null rows are outage artifacts (vs. genuine misses) is explicitly out of scope per the ticket -- that's a separate proposal, not attempted here.

## Stacking

Stacked on `lml-1106-bandcamp-fail-fast-breaker` (#1110) -- **must merge after it**. GitHub will auto-retarget this PR to `main` once #1110 merges. Sibling PR #1116 (breaker coverage of default-mode traffic) touches the same file in parallel; this PR stays out of `find_album_match`'s breaker wrapper and `clients/bandcamp_breaker.py` entirely, per the agreed ownership split.

## Test plan

- [x] `uv run --no-sync ruff check .`
- [x] `uv run --no-sync ruff format --check .`
- [x] `uv run --no-sync mypy . --ignore-missing-imports`
- [x] `uv run --no-sync pytest -m 'not pg and not external_api' -q` -- 5451 passed, 1 skipped, 2 xfailed
- [x] New unit tests pin: default-path `BandcampTransportError`/`BandcampSearchUnavailableError` raises at each layer of `clients/bandcamp.py`; end-to-end `resolve_streaming_url_with_cache` behavior with a real `BandcampClient` (both the failure -> no write and the genuine-miss -> write cases); the `/streaming-check` orchestrator's degraded-response posture; the background warm's swallow-and-skip posture; the offline drain's `phase_lookup` leave-pending posture


jakebromberg:
Review round 3 surfaced that the `is_error` row provides no backpressure on the default `/lookup` warm path: a fresh error row routes to the enqueue branch, and `bypass_error_row=True` then tells that warm to ignore the row that triggered it. The per-key warm rate during an outage is therefore unchanged from `main`, plus one extra UPSERT per failure.

Resolved deliberately rather than by changing behavior here. The column's real consumer is the fail-fast probe path — `bandcamp_probe.py` maps `cache_error_recent` to `unresolved` instead of `absent`, which is load-bearing for the LML#1100 probe engine once `lml_bandcamp_live_probe` enables after the BS#642 drain. This PR keeps that and corrects the four docs plus five code comments that claimed default-path backpressure, so the documentation states what the mechanism actually does.

Making the row throttle the default path is filed as #1141, which also carries the Spotify `Retry-After` question: `SpotifyRetryBudgetExceededError` mints `is_error` rows too and its backoff can run hours against a 5-minute TTL. That mismatch is unreachable today precisely because the default path bypasses the row, and it becomes live the moment #1141 ships.

Issue #904, opened 2026-07-23
Raise the Apple Music rate limit behind an env knob (LML_APPLE_MUSIC_RATE_PER_MIN) — self-throttle nulls 56% of probes


jakebromberg:
## Problem

The `apple_music.search` (`find_track_url`) probe on the `/api/v1/lookup` hot path is throttled to **1 req/s** by LML's **own** `AsyncLimiter(60, 60)` — a **self-imposed** rate, *not* an Apple-published limit. Apple publishes no official Apple Music API rate limit; the "60 req/sec" figure in `docs/plans/apple-music-non-library-persistence.md:142` is a misread of that `(60, 60)` = 60-per-60s config. The community estimate is ~20 req/s per token (unverified).

At 1 req/s the probe queues behind the `asyncio.wait_for(..., 4s)` ceiling and a **majority of probes time out and return null**. 7-day Sentry (`span.op:apple_music.search`, n≈17,785):

| `apple_music.search.result` | count | share |
|---|---:|---:|
| **null (cancelled at the 4s `wait_for`)** | 9,907 | **56%** |
| hit | 6,309 | 35% |
| miss | 1,533 | 9% |
| error | 36 | 0.2% |
| **429 (any kind)** | **0** | **0%** |

Two facts fall out: (1) **zero 429s in 7 days** at 1 req/s + 60-burst — Apple never throttled us once, so there is real headroom to raise; (2) the raw Apple HTTP GET is ~338 ms, so the rest of the 4s is pure 1-req/s acquire-wait. **The self-throttle — not Apple, not the network — is nulling most probes today**, and each null is then persisted downstream as a missing `apple_music_url` (the LML#782 / BS#1192 dead-button shape). This probe is also the dominant ~3.6s p50 span on `/lookup` (LML#803 saturation-era investigation), so widening it cuts both latency and the null rate.

## Desired end state

The Apple Music `AsyncLimiter` per-minute ceiling is an **env knob** (`LML_APPLE_MUSIC_RATE_PER_MIN`, default 60 = the historical 1 req/s), read at runtime so it can be widened from Railway with **no redeploy and instant rollback** (mirroring `LML_STREAMING_WARM_CONCURRENCY`). Merge is **zero-behavior-change** until the var is set. Ops then rolls the ceiling up in steps (60 → 300 → 600/min = 1 → 5 → 10 req/s) watching the 429-count + null-rate, staying under the ~20 req/s community estimate.

[...]

Issue #719, opened 2026-07-04
Streaming matcher: token_set subset-inflation lets Various Artists queries cache unrelated (Suno) tracks


jakebromberg:
## Problem

The Apple **track** probe scores its axes with `token_set_ratio` ([`find_track_metadata`, `apple_music.py:407-423`](https://github.com/WXYC/library-metadata-lookup/blob/2eb9c4c5c51e53b40c6836f34c03650e4f2eb3f3/clients/streaming/apple_music.py#L407-L423), against the shared `_APPLE_MUSIC_MATCH_FLOOR = SCORE_MATCH_ACCEPTANCE_FLOOR = 80.0`). `token_set_ratio` returns **100 whenever the shorter string's tokens are a subset of the longer's**, regardless of how much extra material the longer side carries. On a **Various Artists** request — which has no discriminating artist name — both axes can clear the floor against wholly unrelated catalog content, and the resolved (wrong) URL is then cached hit-forever.

Observed in production during the LML#638 two-week telemetry review (5 span events, 2026-07-01):

| axis | query | matched candidate | `token_set_ratio` |
|---|---|---|---|
| artist | `Various Artists - Blues` | `Various artists Suno` | **85.7** (shared `various artists` cluster) |
| title | `Hound Dog` | `Black leather (The hound dog mix)` | **100** (`hound dog` ⊂ candidate) |

Both clear 80/80, so the matcher accepts an **AI-generated (Suno) catalog-spam track** for a blues-compilation request and caches its streaming URL. Two independent weaknesses compound:

1. **`token_set_ratio` subset-inflation on the title axis.** A short, generic query title (`Hound Dog`) that appears as a token-subset inside a long unrelated candidate title scores a perfect 100. The title axis — normally the disambiguator (see the LML#592 / #638 analysis) — carries no signal here.
2. **Various Artists queries are un-disambiguable on the artist axis.** A `Various Artists …` request shares the `various artists` token cluster with *anything* else credited to VA, so the artist axis clears the floor against unrelated VA-credited content (here, Suno spam self-credited as "Various artists Suno").
3. **(Aggravating) Suno-generated spam is now present in Apple's catalog** — a new class of junk candidate that keyword-matches real requests.

[...]

PR #1143, opened 2026-08-05
Reject V/A↔V/A Apple track matches that never cleared the album axis

### LINKED ISSUE #1139: Streaming matcher: 'Various Artists - <Genre>' prefix clears the artist floor against unrelated V/A buckets (62% of marginal false positives)
## Problem

WXYC catalogs Various-Artists compilations under a **shelf-genre convention**, not an artist name: `Various Artists - Blues`, `Various Artists - Latin`, `Various Artists - Jazz`. The streaming matcher scores that string as if it were an artist name.

Because `Various Artists - ` is a constant ~19-character prefix, **any two WXYC V/A rows score ~85 against each other regardless of content** — comfortably over the 80 acceptance floor. The shared prefix does all the scoring work and carries zero identifying information.

Measured over 30d of `matcher.match` spans (the LML#592 telemetry, 2026-07-06 → 2026-08-05):

| query_artist | matched_artist | artist_score | n |
|---|---|---|---|
| `Various Artists - Blues` | `Various Artists - Document Records` | **85.00** | 14 |
| `Various Artists - Latin` | `Various Artists - Azzurra Music` | **85.00** | 4 |

Those are the **two highest-volume marginal clears in the entire band** and account for **18 of 29 false-positive occurrences (62%)** in a hand-labeled 40-pair / 166-occurrence sample.

## Where the false positives actually occur (2026-08-05 re-examination)

The span-level evidence (full table in the issue comment below) relocated and reshaped this ticket. Three facts:

1. **Every measured V/A false positive is on the Apple Music track path.** All 21 marginal V/A clears in the trailing 30d carry surface `apple_music:track` or `apple_music:track_album_fallback`, emitted from [`_select_best_track_candidate`](https://github.com/WXYC/library-metadata-lookup/blob/83190beca42d0e98e26a91f7680272b2e45a7ea5/clients/streaming/apple_music.py#L563) — an inline `token_set_ratio` loop. **Zero** occurred at `find_best_match` / `find_best_typed_match` in `clients/streaming/matching.py`, which serve the album path (all 5 services) and the Discogs artwork path. An earlier draft of this ticket named those two sites; that was wrong.
2. **Title scores on the FP pairs are 95–100.** `I'm On My Way` = `I'm On My Way`, `Desafinado` = `Desafinado`, `Early in the Mornin'` ≈ `Early In the Morning`. These are generic blues/latin standards recorded on hundreds of compilations: the track title *genuinely agrees* and still points at the wrong record. Since [`is_acceptable_match`](https://github.com/WXYC/library-metadata-lookup/blob/83190beca42d0e98e26a91f7680272b2e45a7ea5/clients/streaming/matching.py#L447-L451) already requires **both** axes ≥ 80, the earlier draft's mechanism — "require the title axis to carry acceptance" — is a no-op against every measured FP. Abandoned.
3. **The dominant surface is the LML#782 album-dropped fallback** ([apple_music.py#L502-L512](https://github.com/WXYC/library-metadata-lookup/blob/83190beca42d0e98e26a91f7680272b2e45a7ea5/clients/streaming/apple_music.py#L502-L512)): the album-constrained pass ran, **every candidate failed the album axis**, and the fallback re-admitted the winner on vacuous-artist + generic-title. For a real artist, artist+track without album was #782's accepted trade. For V/A↔V/A the artist axis is empty, so the fallback accepts on a generic track title **alone**.

## This cannot be fixed by moving the floor

The LML#592 marginal sample was labeled in full. False positives span artist scores **83.87–85.71**; true positives span **80.0–89.55**. The FP band is not merely overlapping the TP band — it is **entirely contained inside it**. No artist-score threshold removes any false positive without destroying true positives.

The proof case:

| query_artist | matched_artist | artist_score | correct? | n |
|---|---|---|---|---|
| `Hank Williams, **Sr.**` | `Hank Williams` | 83.87096774193549 | ✅ yes | 17 |
| `Hank Williams, **Jr.**` | `Hank Williams` | 83.87096774193549 | ❌ no | 3 |

Same matched artist, identical score to 14 significant figures, opposite correctness. This confirms LML#638's decision to close no-change on the artist-axis floor — that decision was correct and should **not** be revisited.

## Desired end state

For a V/A↔V/A pair on the Apple track path, the **album axis is the only real evidence** that a track-title clear is the right recording. Such a pair is acceptable **only when the album-constrained pass supplied the winner**: the LML#782 fallback pass and album-less track queries reject V/A↔V/A candidates. A V/A compilation whose **album** genuinely agrees still matches; V/A↔V/A stops matching on a shared prefix plus a generic standard's title. The album path (all 5 services) and the Discogs artwork typed path are unchanged.

## Mechanism

**1. Predicate** — pure, unit-testable, in `clients/streaming/matching.py` beside `title_subset_is_degenerate`:

```python
def va_artist_axis_is_uninformative(query_artist: str, candidate_artist: str) -> bool:
    """True when both sides are V/A credits, so the artist score is carried
    entirely by the shared 'Various Artists' prefix and identifies nothing."""
```

Both sides checked via `wxyc_etl.text.is_compilation_artist` (already the org's canonical V/A detector; LML imports it elsewhere, e.g. [`lookup/matching.py#L13`](https://github.com/WXYC/library-metadata-lookup/blob/83190beca42d0e98e26a91f7680272b2e45a7ea5/lookup/matching.py#L13)). Feed it **raw** strings, not `normalize_for_comparison` output — it lowercases internally, and its leading-anchored rule means `Various Artists - Blues` and `Various Artists & Pan Ron` hit while real artists like `Various Production` don't.

**2. Guard** — in [`_select_best_track_candidate`](https://github.com/WXYC/library-metadata-lookup/blob/83190beca42d0e98e26a91f7680272b2e45a7ea5/clients/streaming/apple_music.py#L563): add a `raw_artist` param (mirrors the existing `raw_song` / `raw_album`), and in the candidate loop, alongside the LML#719/#721 guards:

```python
if norm_album is None and va_artist_axis_is_uninformative(raw_artist, attrs.get("artistName") or ""):
    continue
```

`norm_album is None` covers both the #782 fallback pass and album-less-from-the-start queries — the whole rule in one clause. Per-candidate (not a winner veto, not query-side fallback suppression): only V/A↔V/A candidates are struck; any other candidate stays eligible on its own merits. That matters because the artist axis here is `token_set_ratio` ([#L607](https://github.com/WXYC/library-metadata-lookup/blob/83190beca42d0e98e26a91f7680272b2e45a7ea5/clients/streaming/apple_music.py#L607)), kept deliberately for leader→ensemble subsets — non-V/A candidates can legitimately clear.

**3. No feature flag** — ships unconditional, matching LML#719's precedent.

**4. Cache purge (same deploy)** — [`entity/track_streaming_url_cache.py`](https://github.com/WXYC/library-metadata-lookup/blob/83190beca42d0e98e26a91f7680272b2e45a7ea5/entity/track_streaming_url_cache.py) (LML#893 L1) is **hit-only and TTL-less**, peeked on the happy path *before* the live probe — so pre-guard wrong V/A deep-links would otherwise serve forever and the guard never re-adjudicates them, on exactly the repeat-play traffic that recurs most. Ship a `scripts/` one-shot (dry-run default, SELECT-first): `DELETE FROM track_streaming_url_cache WHERE service = 'apple_music' AND` the artist-normalized key matches the V/A forms (mirror `is_compilation_artist`: leading `various artists` / `v/a` / `v.a` / `soundtracks`, exact `various` / `soundtrack` / `compilation`). Accepted over-deletion: correct album-cleared V/A deep-links are purged too; they re-fill via the album-constrained pass on next play (~4.8s warm probe each).

## Constraints

- **Failure mode stays safe.** A guard-struck pair drops the Apple URL for that item; never persist a match this guard rejects.
- **Do not raise `SCORE_MATCH_ACCEPTANCE_FLOOR`.** See the proof above; LML#638 settled this.
- **V/A matching must survive via the album-constrained lane and the album path.** The guard must not block a V/A candidate whose album axis genuinely clears (≥ 80, non-degenerate under LML#721). *(This replaces the earlier draft's "do not simply reject the pair": on the album-less track surface we now do reject V/A↔V/A outright — the old constraint assumed FP title scores were low; they are 95–100, so no non-rejecting rule can satisfy the acceptance criteria.)*
- The album path (`find_best_source_match`, 5 services), `find_best_typed_match`, `title_subset_is_degenerate`, and the title floor are all untouched.

## Acceptance criteria

- [ ] Regression, real pair from the 30d sample: query (`Various Artists - Blues`, song `I'm On My Way`, album=<catalog album>) against a fixture where the album-constrained pass clears nobody and a `Various Artists - Document Records` candidate would previously win the fallback → `find_track_metadata` returns `None`.
- [ ] Likewise `Various Artists - Latin` / `Various Artists - Azzurra Music` (`Aguas de Março`).
- [ ] A V/A↔V/A candidate that clears the album-constrained pass (track ≥ 80, album ≥ 80 non-degenerate) still resolves and still persists its URL.
- [ ] Album-less V/A↔V/A track query rejects — pin the real pair (`Various Artists - Asia`, `Jombang Jet`) → (`Various Artists & Pan Ron`, title 100) as a **deliberate kill**, with a comment documenting the accepted recall cost (the LML#719 idiom).
- [ ] Non-V/A pairs completely unaffected — pin the labeled-sample true positives (`Ice-T`→`Ice T` @ 80.0, `Aha`→`a-ha` @ 85.71, `Altin Gun`→`Altın Gün` @ 88.89, `N.E.R.D.*`→`N.E.R.D` @ 87.5) as a no-regression corpus.
- [ ] `va_artist_axis_is_uninformative` unit-tested directly: both-sides matrix including the trailing-credit negative (`CAGAYANO VARIOUS ARTISTS` — leading-anchored, does not fire) and a real-artist negative (`Various Production`).
- [ ] `SCORE_MATCH_ACCEPTANCE_FLOOR` is unchanged at 80.0.
- [ ] Purge script: dry-run prints the candidate count; `--execute` deletes only the `service='apple_music'` V/A net; run result recorded on the PR.
- [ ] Post-deploy: the `matcher.match` marginal-clear telemetry keeps firing and the V/A class disappears from the marginal band vs. the baseline below (absence-vs-baseline — the same verification LML#719 used).

## Out of scope (documented residuals)

The remaining labeled false positives are deliberately out of scope — each needs a different guard:

- **Album-axis-fooled V/A pairs** — 2 of the 18 occurrences (`Various Artists - Latin` → Azzurra on surface `apple_music:track`, 2026-07-11) cleared the album gate: a generic album title fuzzy-agreed ≥ 80, so the guard's album-constrained lane admits them. Would need an album-genericness signal; much smaller volume.
- **One-sided / trailing V/A credits** — `Various Artists - Xmas` → `CAGAYANO VARIOUS ARTISTS` @ 81.08: `is_compilation_artist` is leading-anchored, so the both-sides predicate misses it. n=1 class.
- **Generational suffix** — `Hank Williams, Jr.` → `Hank Williams` (3 occurrences).
- **Surname-only collision** — `Al Martino` → `Pat Martino`, `Andrew Parker` → `Andrew Pekler`, `Ali Farka Toure` → `Vieux Farka Touré` (2 each).
- **Lineup superset** — `Crosby, Stills & Nash` → `Crosby, Stills, Nash & Young` (2). **Recommend leaving alone**: many legitimate matches share this exact shape.
- **Already-persisted wrong URLs in Backend-Service** — WXYC/Backend-Service#2000 (blocked on this ticket; must run only after the guard **and** the purge deploy).

## Baseline for post-deploy comparison

The marginal band is ~0.9% of all matches (22 marginal vs 2,414 non-marginal in 24h), with a ~17.5% false-positive rate by occurrence within the band. This guard addresses 16 of the 18 V/A occurrences by construction (the 2 album-axis-fooled ones remain), which should cut the band's FP rate to roughly 6–7% without moving the band's size.

## Related

- WXYC/library-metadata-lookup#638 — closed no-change on tightening the artist-axis floor. This ticket supplies the labeled evidence #638 called for and **confirms** its decision; the fix here is structural, not a floor change.
- WXYC/library-metadata-lookup#719 — the title-axis half of the same V/A insight (`title_subset_is_degenerate`). Placement and fail-safe model for the guard here.
- WXYC/library-metadata-lookup#782 — the album-dropped fallback this guard scopes for V/A queries; #721 — the album-axis degenerate guard whose clears define "genuinely agrees".
- WXYC/library-metadata-lookup#893 — the L1 track cache the purge targets.
- WXYC/library-metadata-lookup#592 — the `matcher.match` telemetry whose marginal sample this labeling consumed.
- WXYC/Backend-Service#2000 — downstream remediation of BS-persisted rows (blocked by this ticket).

jakebromberg:
Closes #1139

## Why

WXYC files Various-Artists compilations under a **shelf-genre convention** rather than an artist name — `Various Artists - Blues`, `Various Artists - Latin`. Because `Various Artists - ` is a constant ~19-character prefix, **any two such credits score ~85 against each other regardless of content**, clearing the 80 floor while carrying zero identifying information. In the trailing-30d `matcher.match` sample those were the two highest-volume marginal clears in the entire band (85.00, n=14 and n=4) — 18 of 29 labeled false-positive occurrences.

**Neither axis can be tightened to fix it.** The FP artist-score band (83.87–85.71) is not merely overlapping the TP band (80.0–89.55), it is *entirely contained inside* it: `Hank Williams, Sr.` and `Hank Williams, Jr.` both score `83.87096774193549` against `Hank Williams` — identical to 14 significant figures, opposite correctness. `SCORE_MATCH_ACCEPTANCE_FLOOR` is unchanged at 80.0 and #638's decision stands. And every measured FP scored **95–100** on the *title* axis, because these are generic blues/latin standards whose titles genuinely agree while pointing at the wrong record.

That leaves the album axis — and the dominant surface is exactly where it was discarded. The #782 album-dropped fallback ran *because* the album-constrained pass rejected every candidate, then re-admitted a winner on a vacuous artist plus a generic title. The discriminating signal existed and said no.

## What

`va_artist_axis_is_uninformative`, beside the #719/#721 subset guards, applied **per-candidate** in `_select_best_track_candidate` when `norm_album is None` — one clause covering both the #782 fallback pass and album-less-from-the-start queries. Per-candidate rather than a winner veto, so a non-V/A candidate stays eligible on its own merits and the artist axis keeps `token_set_ratio` for leader→ensemble subsets — **paired with a no-promotion rule**, because a bare per-candidate skip is not safe on its own. Striking the top-ranked candidate would hand the win to whatever scored next, and against a V/A query the only realistic survivor is the trailing-credit class (`CAGAYANO VARIOUS ARTISTS`, 81.08) that this PR's own tests document as a measured false positive — so the guard would have traded wrong-URL-A for wrong-URL-B rather than withholding a link. `_select_best_track_candidate` therefore remembers the best struck score and suppresses the pass when it matched or beat the best surviving candidate. Removals are allowed; promotions are not, which is what makes the "a MISSING Apple link, never a wrong one" claim actually true. V/A matching survives through the album-constrained lane and through the untouched album path (all 5 services), where the same 30d window measured **zero** V/A false positives.

## Two corrections to the issue's proposed mechanism

Both verified against the code and the installed crate, not argued:

1. **The guard keys on the NORMALIZED artist, not the raw string** (the issue says raw). `to_match_form` does not strip punctuation, so the `v/a` / `v.a` prefixes survive it — normalizing preserves every positive *and* every negative in the labeled matrix, and additionally catches `"  Various Artists"`, `"Various  Artists - Blues"` and `"Vàrious Artists"`, all of which are `False` raw. Decisively, the normalized query artist **is the L1 cache key**, so the guarded key space and the purged key space are the same space by construction rather than by coincidence. This also removes the proposed `raw_artist` parameter and both call-site edits.

2. **The purge scopes to `apple_music_track`, not `apple_music`.** The latter can never exist in the table — `TRACK_CACHE_KEYS` maps `APPLE_MUSIC -> "apple_music_track"`, pinned by a named CHECK — so the issue's `DELETE` would have removed **zero rows while reporting a clean zero**, silently leaving the exact ordering hazard the issue itself flags.

## Cache purge (same deploy, not after)

`track_streaming_url_cache` (#893) is hit-only, TTL-less, and peeked *before* the live probe, so without the purge pre-guard wrong deep-links serve forever on exactly the repeat-play traffic that recurs most — the guard would never get to re-adjudicate them.

`scripts/purge_va_apple_track_cache.py` uses the coarse-net + pure-arbiter idiom from `audit_va_writeback_pollution.py`, with the donor's regex **widened** to `v[./]\s*a\.?` after measurement showed `v.a`, `v.a - jazz` and `v.a 1998` are arbiter-positive but donor-net-negative — reused verbatim it would not have been a superset. Dry-run by default; `--limit`/`--after` waves; and a **recovery CSV written before every DELETE** on both paths, because the purge knowingly over-deletes and #904's ~56% probe-null rate means purged keys re-null rather than re-fill until the rate ceiling is raised.

## Telemetry

Splits `surface="track_album_less"` off `"track"`, so `"track"` now means strictly "the album-constrained pass won" and the album-less population this guard *also* strikes becomes measurable from deploy. Review flagged that population as wider than the evidence base; rather than re-scope the ticket, this makes it observable. **Any before/after comparison spanning this change must query `surface IN ('track','track_album_less','track_album_fallback')`** — `"track"` alone is now a strictly smaller population.

## Testing

5258 unit tests pass (5231 baseline + 27 new); `ruff check` + `ruff format --check` clean.

- Both real FP pairs pinned as rejections (`I'm On My Way` / `Aguas de Março`).
- A V/A pair whose **album clears** still resolves and stays `album_verified=True` — the lane V/A matching survives through.
- The album-less `Jombang Jet` pair pinned as a **deliberate kill**, with its accepted recall cost documented inline.
- `CAGAYANO VARIOUS ARTISTS` (81.08) pinned as the boundary: `is_compilation_artist` is leading-anchored, so this one-sided credit is *not* struck. It is also the only realistic non-V/A candidate that clears the artist floor against a V/A query at all — a real artist name scores 17–24 and never reaches the guard. This assertion is what fails if the AND is ever loosened to an OR.
- Purge tests assert the **superset property in both directions**, which is what catches a regex/arbiter gap like the dotless-`v.a` one above.

## Deviation from the Bug Fix Protocol

No integration test is added, and that is deliberate rather than an omission: CI provisions only `DISCOGS_TOKEN`, and Apple Music needs an ES256 developer token it does not hold, so an `external_api`-marked Apple test would never execute. The unit tests already mock only `httpx` — they drive the real `find_track_metadata` → `_select_best_track_candidate` → `matching.py` → real `is_compilation_artist` path end to end, so a mocked file under `tests/integration/` would be a near-duplicate for zero added coverage. Rationale recorded in `docs/plans/lml-1139-va-track-guard.md`.

## Deploy order

1. Merge + deploy the guard.
2. Roll `LML_APPLE_MUSIC_RATE_PER_MIN` 60 → 300 → 600 via `set-railway-var.yml` and **confirm it took** (a bare var-set can return `SKIPPED`).
3. Purge, staged in waves, staging first, recovery CSVs kept.
4. Only then unblock WXYC/Backend-Service#2000, handing it the deploy SHA.


jakebromberg:
## Review findings addressed (commit `c82a7f0`)

**1. Guard promotion — fixed.** The per-candidate `continue` really could hand the win to a worse candidate; `test_guard_is_per_candidate_not_a_winner_veto` demonstrated it against its own sibling test (struck candidate 93.59 combined, surviving documented-FP candidate 90.54). `_select_best_track_candidate` now tracks the best struck score and suppresses the pass when it matched or beat the best surviving candidate — compared against `best_overall_score` so it stays independent of the orthogonal artwork preference, and ties suppress since the loop breaks them by response order. Split into two tests: one pinning that a strike never promotes, one pinning that a candidate which already outscored every struck one still wins (thin 89.74-vs-90.54 margin, so it fails if the rule silently becomes "any strike suppresses"). Docstring, `docs/architecture.md` and the PR body updated — the "never a wrong one" claim is now true rather than aspirational.

**2. Recovery CSV clobbering — fixed.** `--out` defaults to a timestamped per-invocation path, and `write_recovery_csv` opens with mode `"x"`, so an explicit path reused across waves aborts instead of truncating.

**3. SELECT/DELETE race — fixed.** The execute path is one `DELETE … RETURNING` inside a single `pg.acquire()` transaction, with the CSV written before commit — so rows can only be gone once they are recorded, and a crash rolls the DELETE back. The dry-run path keeps its read-only SELECT (no gap to race).

**7. Deleted-row count — fixed by the same change.** The logged count is now the DELETE's own row count, which is what `docs/scripts.md` asks the operator to record for wave reconciliation.

**4. False pg-coverage claim — made true.** Added `tests/integration/test_purge_va_apple_track_cache_pg.py` (6 tests, `@pytest.mark.pg`), covering the `~` regex operator, the `ANY($2)` array bind, `DELETE … RETURNING`, the `LIMIT $4` + cursor wave composition, dry-run inertness, and an all-rejected wave. Registered in `_LML_CACHE_FIXTURE_FILES` so the guard-adoption meta-test binds it. Plan text updated to name the actual file.

**5. Donor `SUPERSET_REGEX` — fixed here rather than deferred.** Widened to `v[./]\s*a\.?`, matching the new script. Widening a coarse net whose sole arbiter is `is_compilation_artist` can only surface more true positives, so it is safe. Added a parametrized superset test to `test_audit_va_writeback_pollution.py`; the old pattern fails it on `v.a`, `v.a - jazz` and `v.a 1998`.

**8. Negative `--limit` — fixed.** `type=_non_negative_int` rejects it at parse time, before any connection opens.

**6. Terminal `absent` — filed as #1144, not settled here.** You're right that a strike is a refusal to adjudicate rather than evidence of absence, and BS freezes the null. Mapping it to `unresolved` needs a signal channel out of `find_track_url` (which returns `str | None`), and it has a real cost: the strike happens *after* the Apple call, so every re-ask would pay a probe and then refuse. The likely right shape is a zero-probe pre-check on the query artist, which is a design decision rather than a mapping tweak. `docs/architecture.md` now records the downstream permanence and points at the issue.

Local gate before push: `ruff check`, `ruff format --check`, `mypy`, `pytest -n auto` (5,618 passed), `pytest -m pg` (565 passed), `scripts/check_plan_links.sh`. Rebased onto `origin/main`.

Issue #1103, opened 2026-08-04
YouTube Music joins the streaming-URL post-process: registry entry, cache read-fill + warm, deferred search-URL fallback


jakebromberg:
## Parent

Sub-issue of #1100 (probe-engine sub-epic), under #830. The response-path follow-up the #833 GO spike called for.

> Code links are pinned to [`204a86b`](https://github.com/WXYC/library-metadata-lookup/commit/204a86bae6f043d4c44aed37d38e8dd35d82230d) (main as of 2026-08-03).

## Problem

YouTube Music has a working verified source but no response-path integration. The client ([`clients/streaming/youtube_music.py`](https://github.com/WXYC/library-metadata-lookup/blob/204a86bae6f043d4c44aed37d38e8dd35d82230d/clients/streaming/youtube_music.py), built for the #1056 drain after the #833 GO spike) honors the `find_album_match(artist, title) -> SourceMatch | None` contract and resolves direct `music.youtube.com/browse/<MPREb_…>` album links unauthenticated — but only the offline drain scripts consume it. On the response path:

- YTM is not in `STREAMING_URL_CACHE_CONFIG` ([`streaming_url_postprocess.py:274`](https://github.com/WXYC/library-metadata-lookup/blob/204a86bae6f043d4c44aed37d38e8dd35d82230d/lookup/streaming_url_postprocess.py#L274)) — so lookups get neither PG cache read-fill nor a background warm, even though the #1056/#1070 drain populated `lml_cache.album_streaming_url_cache` rows for it.
- `StreamingService.YOUTUBE_MUSIC.album_cache_key` is `None` ([`streaming/service.py:87-96`](https://github.com/WXYC/library-metadata-lookup/blob/204a86bae6f043d4c44aed37d38e8dd35d82230d/streaming/service.py#L87-L96)).
- `enrich_one` builds YTM's `music.youtube.com/search?q=` fallback **before** the post-process (unlike Bandcamp's deferred fallback) — so even a cached verified URL couldn't win today; see the normalization comment at [`item.py:598-606`](https://github.com/WXYC/library-metadata-lookup/blob/204a86bae6f043d4c44aed37d38e8dd35d82230d/lookup/enrichment/item.py#L598-L606).

Net: every persisted `youtube_music_url` that lacks a librarian `streaming_links` override is a synthesized search URL.

## Desired end state

[...]

Issue #1098, opened 2026-08-04
Bandcamp live probe on the enrichment path: synchronous, bounded, cache-first — model on the Apple probe (fills the flowsheet row first-time)


jakebromberg:
## Parent

Sub-issue of [#1100](https://github.com/WXYC/library-metadata-lookup/issues/1100) (probe-engine sub-epic), under [#830](https://github.com/WXYC/library-metadata-lookup/issues/830) (Epic: Comprehensive verified streaming-link coverage). In #1100's ship-dag DAG this is the root node: it blocks #1101 (probe-engine extraction), which blocks #1103. Sibling of [#1052](https://github.com/WXYC/library-metadata-lookup/issues/1052) (Spotify parity on the bulk path) — this is the **Bandcamp** counterpart, but via a **synchronous live probe** rather than a warm, because Bandcamp has no off-path warm coverage on the enrichment path at all.

> Code links are pinned to [`204a86b`](https://github.com/WXYC/library-metadata-lookup/commit/204a86bae6f043d4c44aed37d38e8dd35d82230d) (main as of 2026-08-03).

> **Split, 2026-08-04.** The client-and-cache-layer plumbing — Bandcamp `fail_fast` mode, the live-probe saturation breaker, and the `resolve_streaming_url_with_cache` pass-through (guardrails 1 and 2 below) — moved to **#1106**, which now blocks this issue. This issue retains the enrichment-path probe itself: `enrich_one` wiring, the `router.py:1044` injection gate, `bandcamp_probe_status`, the `lml_bandcamp_live_probe` flag, the timeout knob, and the `entity.release_identity` mint decision. The split follows a first implementation pass that landed the client layer and ran out of room before reaching the probe.
>
> One design point moved with it: #1106 has the fail-fast 429 **raise** `BandcampRateLimitedError` rather than return `None` as guardrail 1 sketches, so a rate-limit shed stays distinguishable from a genuine no-match and can never be written as a negative-cache row. Guardrail 1's intent is unchanged; only the signalling mechanism is sharper.

## Problem

[...]

PR #1146, opened 2026-08-05
Per-service status maps and an extracted Apple probe module (LML#1101)

### LINKED ISSUE #1101: Extract a config-driven inline probe engine from enrich_one — Apple + Bandcamp blocks become registry rows; per-service probe-status map
## Parent

Sub-issue of #1100 (probe-engine sub-epic), under #830.

> Code links are pinned to [`204a86b`](https://github.com/WXYC/library-metadata-lookup/commit/204a86bae6f043d4c44aed37d38e8dd35d82230d) (main as of 2026-08-03).

## Problem

`enrich_one` resolves in-response streaming links through per-service bespoke blocks. The Apple inline probe is ~70 lines ([`lookup/enrichment/item.py:289-356`](https://github.com/WXYC/library-metadata-lookup/blob/204a86bae6f043d4c44aed37d38e8dd35d82230d/lookup/enrichment/item.py#L289-L356)) of flag gates, cache-first peek, `asyncio.wait_for` ceiling with the #930 deadline clamp (including the self-inflicted-timeout distinction), L1 track-cache write-back (#893), and #1053 status mapping. #1098 adds a structurally-parallel Bandcamp block plus two guardrails Apple doesn't need (429 fail-fast, breaker + shed counter). `resolve_streaming_status` ([`lookup/enrichment/streaming_status.py`](https://github.com/WXYC/library-metadata-lookup/blob/204a86bae6f043d4c44aed37d38e8dd35d82230d/lookup/enrichment/streaming_status.py)) takes service-specific positional parameters (`apple_music_override`, `apple_probe_status`) and derives Spotify/Bandcamp verdicts from URL presence — every new probing service extends its signature. A third instance (Spotify, if #1052 adopts the probe shape; YouTube Music later) makes `enrich_one` unmaintainable, and the module-budget guardrail (`tests/unit/test_module_budgets.py`) will force an extraction under pressure rather than by design.

## Desired end state

One generic inline-probe loop in `enrich_one`, driven by a per-service probe config keyed by `StreamingService` — either new fields on `StreamingUrlCacheConfig` ([`lookup/streaming_url_postprocess.py:274`](https://github.com/WXYC/library-metadata-lookup/blob/204a86bae6f043d4c44aed37d38e8dd35d82230d/lookup/streaming_url_postprocess.py#L274)) or a sibling `INLINE_PROBE_CONFIG` registry (implementer's choice; keep the import-time drift-guard convention of [`streaming_url_postprocess.py:307`](https://github.com/WXYC/library-metadata-lookup/blob/204a86bae6f043d4c44aed37d38e8dd35d82230d/lookup/streaming_url_postprocess.py#L307)). Per-service config covers: enable-flag setting name, timeout env var + default (`lookup/timeouts.py` `probe_timeout_s_from_env`), #930 deadline clamp, probe method (track-level `find_track_url`/`find_track_metadata` for Apple vs album-level `find_album_match` for Bandcamp), L1 track-cache eligibility (Apple-only), breaker + fail-fast mode (Bandcamp), and #1053 status mapping. `resolve_streaming_status` accepts a per-service probe-status map instead of growing another positional parameter per service. Apple and Bandcamp become two registry rows; observable behavior is byte-identical.

## Constraints

- **Pure refactor — no behavior change.** Characterization tests first (the `tests/unit/test_bandcamp_retry_characterization.py` approach), then move code under them. Existing Apple semantics (#444 degrade-to-null, #930 clamp + self-inflicted-timeout distinction, #1053 verdicts, #893 L1 write-back, the #477 override gate) and #1098's Bandcamp semantics keep byte-identical outcomes.
- The interactive-vs-bulk client-injection seam in `lookup/router.py` is untouched: which services probe on which path stays a wiring decision, not an engine decision.
- Respect module budgets — the engine likely lands in its own module under `lookup/enrichment/`.

## Dependencies (ship-dag)

- **Blocked by #1098** — the engine is extracted from two working instances (Apple + Bandcamp), not designed speculatively from one. Do not start until #1098 merges.
- **Blocks #1103** (YouTube Music registry entry) — YTM's status reporting lands on the generalized probe-status map, not a third bespoke signature extension.

## Acceptance criteria

- [ ] Characterization tests pin the Apple + Bandcamp probe behavior before any code moves; they pass unchanged after.
- [ ] Apple + Bandcamp probes are config rows; no per-service `if` blocks remain in `enrich_one`'s probe section.
- [ ] `resolve_streaming_status` takes a per-service probe-status map; adding a probing service requires no signature change.
- [ ] Import-time drift guard between the probe registry and `ALBUM_CACHED_SERVICES` / `STREAMING_URL_CACHE_CONFIG`.
- [ ] Module budgets, mypy, ruff, and the full unit suite green.

## Related

WXYC/library-metadata-lookup#1100 (parent sub-epic) · #1098 (second probe instance, the blocker) · #595 (streaming-cache polymorphism — the tier-2/4 precedent for config-driven service loops) · #1037 (one `StreamingService` enum + registry) · #903 (interactive-path Apple probe removal — same seam; coordinate if in flight) · #1052 (Spotify shape decision — a probe shape would be this engine's third row)

jakebromberg:
Closes #1101.

## What landed

**1. `resolve_streaming_status` takes per-service maps.** It grew two positional parameters per probing service (`apple_music_override` + `apple_probe_status`, then `bandcamp_url` + `bandcamp_probe_status`); #1103's YouTube Music leg would have made eight. It now takes `verified_urls` + `probe_status` keyed by `StreamingService`, so adding a probing service is a new key at the call site and no signature change here. This is the acceptance criterion that actually unblocks #1103.

It also drops a service `StreamingResolution` does not model rather than raising on it. The model is generated from wxyc-shared's `api.yaml`, so `StreamingService` can carry a member whose wire field has not shipped — exactly `youtube_music` today. A not-yet-modelled verdict degrades to an omitted key instead of a 500 on the lookup path, matching the LML#444 posture the rest of `enrich_one` holds.

**2. The Apple probe moves to `lookup/enrichment/apple_probe.py`**, sibling of #1098's `bandcamp_probe.py`. It returns an `AppleProbeResult` five-tuple: the URL slot, the LML#1053 verdict, the raw `AppleMusicTrackMatch` the LML#505 sibling gate reads, artwork, and release year. Apple was the last probe still written longhand in `enrich_one`, and `item.py` had reached **exactly** its 850-line ceiling.

| Module | Before | After | Budget |
|---|---:|---:|---|
| `lookup/enrichment/item.py` | 850 | **677** | 850 → 700 (recalibrated down) |
| `lookup/enrichment/apple_probe.py` | — | **309** | 400 (new, 1.3×) |
| `lookup/enrichment/streaming_status.py` | 76 | **100** | 100 → 150 (was exactly at ceiling) |

## What this deliberately does not build

#1101 asks for a config-driven engine turning Apple and Bandcamp into registry rows. **That premise is stale.** The issue was filed 2026-08-03 against `204a86b` and predicted #1098 would add a structurally-parallel Bandcamp *block* to `enrich_one`. #1098 extracted to its own module instead — so there were never two parallel blocks in `enrich_one` for an engine to collapse. The AC "no per-service `if` blocks remain in `enrich_one`'s probe section" is satisfied by this PR through extraction, which is the posture the repo already takes (`streaming_status.py`, `bandcamp_probe.py`, `streaming_warm.py`, `streaming_warm_admission.py`).

Read side by side, the two probes share a skeleton, not a body:

| Axis | Apple | Bandcamp |
|---|---|---|
| Cache substrate | **track** cache, keyed `(service, artist, album, song)` | **album** cache |
| Call | client-direct (`find_track_url` / `find_track_metadata`) | `resolve_streaming_url_with_cache(fail_fast=True)` |
| Return | rich `AppleMusicTrackMatch` → three downstream consumers | bare URL |
| Path gating | client injection — interactive **and** bulk | `should_suppress_streaming_warm()` + `is_top1` — bulk only |
| Error taxonomy | clamped-vs-genuine timeout split + 4 Sentry keys | breaker/429/transport → PostHog shed counter |
| Mint | none | own `_mint_identity` under its own ceiling |

Roughly twenty lines are genuinely common. A config carrying the rest needs about a dozen fields of which most serve exactly one row — a dispatch table wearing a registry's clothes.

The registry is real, but its family is **`resolve_streaming_url_with_cache`-shaped album probes**: Bandcamp today, YTM via #1103, Spotify via #1052. Apple is the outlier on every axis. That family has one member right now, so the extraction becomes correct at #1103 with three parallel rows — and Apple stays bespoke because it genuinely is. The rationale is recorded in `apple_probe.py`'s module docstring so it does not get re-litigated.

## Verification

- **Near-pure motion, bounded by AST.** Comparing the moved block against `origin/main` with comments and docstrings excluded yields exactly **three** deltas, all deliberate: `settings = get_settings()` and the `skip_happy_probe` expression both moved to the caller, and the deadline-clamp `IfExp` became a call to the new shared `clamp_probe_timeout_for` helper. Nothing else differs. (The first revision of this PR was pure motion bar the `get_settings()` line; the other two deltas come from the xhigh review — see the review-response comment.)
- **Characterization untouched.** `test_enrichment.py`, `test_enrichment_track_cache.py` and `test_streaming_status.py` are the net; no assertion in them changed. 18 stale patch targets and 2 `logger` patches were repointed to the new module — that is the whole test-side delta on those files.
- **New direct surfaces.** `test_streaming_status_map.py` (16 tests) pins the map contract including unknown-key dropping and `StrEnum`/bare-string key interop; `test_apple_probe.py` (17) pins the `AppleProbeResult` tuple, which end-to-end tests cannot catch a silently-dropped field in.
- **5367 unit tests pass**; ruff, ruff format, and mypy clean.

## Note on `row_artist` / `search_term`

Typed `str`, not `str | None`: `enrich_one` derives both as `... or ""`, so "missing" arrives as the empty string and the falsiness checks are what gate the probe — same as at the pre-#1101 call site. mypy caught the discrepancy in the first draft of the signature.

---

## Post-review addendum

An xhigh review raised 12 findings, all addressed in `9275741` — see the [review-response comment](https://github.com/WXYC/library-metadata-lookup/pull/1146#issuecomment-5198553697). The one that mattered: `verified_urls` was a trap for exactly the extension it advertises, because YouTube Music and SoundCloud slots already carry templated **search URLs** by the time `resolve_streaming_status` is called (Bandcamp is protected only because LML#573 PR-3 defers its fallback past that call). Following this PR's own "just add a key" recipe at #1103 would have reported a search page as `verified`, and BS/iOS would render an available-badge for a link that resolves nothing. The map is now built from a documented `_RESOLUTION_PROVING_URL_SERVICES` constant with tripwire tests.

Also landed from the review: six stale `item.py` references repointed at `apple_probe.py`; the duplicated deadline-clamp idiom extracted to `clamp_probe_timeout_for` in `spine_deadline.py` (it was the one genuinely-common piece this PR named and then left copy-pasted, and #1103 would have made a third copy); the probe now takes a computed `skip_happy_probe` instead of the override URL; `StreamingResolution.model_fields` pinned against `StreamingService` so a wxyc-shared rename can't silently drop every Apple verdict; and the budget arithmetic recomputed from the real measurement.


jakebromberg:
## Review response — 12 findings, 12 addressed

Thanks; the first one was a real trap and I'd have shipped it.

### The load-bearing one

**`verified_urls` was a trap for the extension it advertises.** Confirmed: `youtube_music_url` / `soundcloud_url` are filled with `_build_streaming_search_url` fallbacks at `item.py:404-412`, *before* the `resolve_streaming_status` call — and Bandcamp is protected only because LML#573 PR-3 deliberately defers its fallback past that call. Following this PR's own "just add a key" recipe at #1103 would have reported a search page as `verified`.

Fixed structurally rather than with a comment: the map is now built from a documented `_RESOLUTION_PROVING_URL_SERVICES` constant whose membership rule is "slots that cannot yet hold a search URL", plus three tripwire tests that fail loudly and say what to do instead (defer the YTM fallback past the call the way Bandcamp's already is, *then* add the key).

### The rest

| # | Finding | Response |
|---|---|---|
| 3 | Six stale `item.py` references | Repointed: `spine_deadline.py:165`, `entity/track_streaming_url_cache.py:30`, `clients/streaming/apple_music.py:403`, `docs/env-vars.md:56`, `docs/scripts.md:234`, `scripts/purge_va_apple_track_cache.py:4`. You were right that `52e4aa5` is the precedent. (`apple_music.py:78` stays — the LML#505 block really is still in `item.py`.) |
| 7 | Clamp idiom duplicated | Extracted to `clamp_probe_timeout_for` in `spine_deadline.py`, which already owns the concept. Both probes use it. This was the one piece of the "~20 common lines" I named and then left copy-pasted. |
| 10 | `apple_music_override` leaks override precedence | Now takes a computed `skip_happy_probe: bool`. The probe no longer knows overrides exist. |
| 6 | Positional unpack | Bound by name via `apple.apple_music_url` etc. |
| 5 | `SimpleNamespace` instead of `EnrichmentContext` | Real `EnrichmentContext`, `AsyncMock(spec=PgSource)`, mirroring `test_bandcamp_live_probe.py`. |
| 2 | Silent drop, no observability | DEBUG log naming the dropped services. |
| 9 | Inert `type: ignore` | Both removed — verified with `--warn-unused-ignores`. You were right that the `dict-item` one asserted the opposite of the contract its test pins. |
| 8 | Budget arithmetic wrong | Measurement was stale (309, now 312). Recomputed at the stated 1.3× rule: `apple_probe.py` → 450, `streaming_status.py` → 200, `item.py` → 750. |
| 11 | Empty-map return shape change | Documented in the docstring, including why the omitted-key shape is the correct one and why it's unreachable today. |
| 12 | `architecture.md` "track-cache-first" | Qualified — the L1 peek is happy-path-only and additionally needs a `song`, a PG handle and both kill-switches. |

### On #4, partially

You're right about the drift risk and I've pinned it, but not by adding a tuple to `streaming/service.py` — a hand-maintained wire-field list there is the free-floating duplication LML#1037 exists to retire, and the question this function asks genuinely is "what does the wire model carry?". Instead `test_streaming_service.py` now pins `StreamingResolution.model_fields` three ways: every field is a `StreamingService` catalog key, the set is exactly `{spotify, apple_music, bandcamp}`, and `_MODELLED_SERVICES` derives from the same set. A rename upstream now fails loudly instead of silently dropping every Apple verdict through pydantic's `extra='ignore'`.

### The pure-motion claim needs qualifying

It no longer holds, and the PR description above has been updated. The AST diff against `origin/main` (comments and docstrings excluded) now shows **exactly three** deltas, all deliberate: `settings = get_settings()` and the `skip_happy_probe` expression both moved to the caller, and the clamp `IfExp` became the shared helper call. Nothing else differs.

5373 unit tests pass; ruff, ruff format and mypy clean.

Issue #1021, opened 2026-07-30
[V/A][§3–5] Populate kind:compilation tracks[] in bulk-resolve (CDC consumer)


jakebromberg:
Sub-issue of WXYC/library-metadata-lookup#271 (§3–5). Blocked by the §2 matcher child (needs the table populated). This is the **CDC / bulk-resolve** consumer of the per-track index (the interactive `/lookup` consumer is a sibling child).

## Problem

`identity/bulk_resolve.py` already returns `kind: compilation` for V/A rows, but with an empty `tracks: []` / `provenance: []` stub — see [`compilation_result()` at `identity/bulk_resolve.py:371`](https://github.com/WXYC/library-metadata-lookup/blob/main/identity/bulk_resolve.py) and its `TODO(#271 follow-up)`. Because the array is always empty, the BS consumer (`jobs/library-identity-consumer`) never gets a resolved signal for V/A rows: it stamps `unresolved_attempted_at` (BS migration 0130) and **re-asks every V/A row every `UNRESOLVED_RETRY_DAYS` (30 days), indefinitely** — the churn [WXYC/Backend-Service#801](https://github.com/WXYC/Backend-Service/issues/801) is filed to stop. Populating `tracks[]` is what breaks that loop.

## Desired end state

`compilation_result()` reads the §1 table and returns real `tracks[]` with per-track `sources[]` (source, `external_id`, `confidence`, `method`) and composed per-track identity, instead of `[]`. Contract already shipped — the generated classes `BulkResolveTrackIdentity`, `BulkResolveProvenanceEntry`, `BulkResolveResult` exist in `generated/api_models.py`; this fills the stub, no `api.yaml` change.

**Note — detection already solved:** #271 §3 proposes detecting V/A via `code_volume_letters LIKE 'Z%' OR EXISTS compilation_track_artist…`. The shipped endpoint already detects via `wxyc_etl.text.is_compilation_artist` ([`bulk_resolve.py:21`](https://github.com/WXYC/library-metadata-lookup/blob/main/identity/bulk_resolve.py)). Do **not** rebuild detection — reuse it.

## Per-track composition (from #271 §4–5)

The §3.4.1.1 composition rules apply per-track:

- Cross-source agreement boost (Rule 2): if Discogs and MB agree on a track artist via wikidata-cache cross-ref, boost that track's confidence. `cross_ref_present(s1, s2)` runs per-track.
[...]

PR #1152, opened 2026-08-07
feat(#1020): per-track compilation identity matcher + backfill

### LINKED ISSUE #1020: [V/A][§2] Per-track compilation matcher + backfill (retire match_compilations.py / merge_cta.py)
Sub-issue of WXYC/library-metadata-lookup#271 (§2 + §6). Blocked by the §1 storage child (needs the table to write into).

## Problem

Per-track compilation identity is produced today by `scripts/match_compilations.py` (writes JSON to disk) and pushed to tubafrenzy MySQL by `scripts/merge_cta.py` (SSH write). Both retire under the [architecture pivot](https://github.com/WXYC/Backend-Service/blob/main/plans/library-hook-canonicalization/architecture-pivot-2026-05-09.md)'s "no cross-service writes" rule. LML needs an *internal* per-track matcher that populates the §1 table, replacing those scripts.

## Desired end state

A per-track matcher that, given a V/A / `Soundtracks - *` library row, resolves each track's credited artist against external sources (Discogs + MusicBrainz at minimum) and UPSERTs `(library_id, track_position, source, external_id, confidence, method)` into the §1 table. A backfill drains the existing V/A + soundtrack population once. `scripts/match_compilations.py` / `scripts/merge_cta.py` are retired (or `match_compilations.py` refactored to write into the §1 table instead of JSON; the `merge_cta.py` SSH-to-tubafrenzy path goes away regardless).

## Matcher concerns (from #271 §2)

Reuse the artist-level cascade where it transfers (`exact_match`, `name_variation`, `alias_match`, `trigram`, `cross_source_agreement`), but:

- **No album-artist anchor.** The V/A `library_id`'s `album_artist` is a genre/shelf bucket (`Various Artists - Hiphop`, `Soundtracks - L`), not a real artist. Match only on the track's own `artist_name`. Confirmed: `is_compilation_artist("Various Artists - Hiphop")` and `is_compilation_artist("Soundtracks - L")` are both `True` — **include the `Soundtracks - *` buckets in the population set, not just `Various Artists - *`.**
- **Features / collabs / joint credits.** Track artists like `"Mira Calix ft. Tom Skinner"` or the live-repro's `"Brian Reitzell And Roger J. Manning, Jr."` need splitting on `ft.` / `feat.` / `&` / `,` (resolve each side), or accepting the join as one canonical entity when the source has it that way.
- **Stricter confidence floors than album-level.** Track artists are less curated; wrong-confidence false positives are easier. Do a confidence-matrix review in the PR.

## Performance (load-bearing — see #271)

- The matcher makes many Discogs / MusicBrainz calls across the whole V/A population. It **must** go through the shared Discogs rate bucket / global limiter (see [LML#879 `discogs_rate_gate_fail_open` counter](https://github.com/WXYC/library-metadata-lookup/blob/main/docs/observability-rowless-flag.md)) — do not open a second uncoordinated Discogs client. A flowsheet-backfill-style flood is the failure mode to avoid.
- This is a batch/offline job, **not** on the `/lookup` request path. Keep it that way — no synchronous live-Discogs work leaks into interactive latency.
- `risk:prod-write`: the backfill writes prod identity rows. Only reset/retry failed rows, never overwrite successfully resolved ones (org data-safety rule).

## Acceptance criteria

- [ ] Per-track matcher resolves Discogs + MB identity for V/A + `Soundtracks - *` rows and UPSERTs into the §1 table.
- [ ] Feature/collab splitting handled with tests covering `ft.`/`feat.`/`&`/`,` and the joint-credit live-repro case.
- [ ] All external calls routed through the shared Discogs rate bucket; a test/inspection confirms no second client.
- [ ] Backfill drains the existing population; only failed rows are retryable, successes are never overwritten.
- [ ] `scripts/match_compilations.py` retired or refactored to write the §1 table; `scripts/merge_cta.py` removed; `docs/scripts.md` updated.
- [ ] TDD: unit tests for the matcher cascade + a `pg`-marked backfill test on fixtures.

## Related

- WXYC/library-metadata-lookup#271 — parent.
- #210, #237 — existing V/A LML tickets; their fixes inform the per-track matcher.
- Existing scripts being retired: `scripts/match_compilations.py`, `scripts/merge_cta.py`.

## Amendments (2026-08-06 — from WXYC/Backend-Service#801's decision-record addendum; read before implementing)

**1. The proposed storage key inherits the exact defect BS#1989 killed one repo over — fix it before DDL.** This body says the matcher "UPSERTs `(library_id, track_position, source, external_id, confidence, method)`" — but its inputs are CTA rows, and **78% of them have no `track_position`** (BS#1989 measured 113,382/144,778 NULL; the NULL fraction is upstream in MySQL too, 77.2%, so no sync fixes it). A position-keyed upsert is dead on arrival for most of the population, exactly as it was in Backend. The key must be credit-shaped — `(library_id, artist_name, track_title, source)`-normalized-form — matching the table's real grain (one row per credited artist per track; a 66-performer track legitimately shares one position).

**2. Store attempt outcomes, not just matches.** When the matcher visits a release, write a row for **every** CTA credit — misses get NULL `external_id` (the same "the leg ran" convention `BulkResolveProvenanceEntry` already documents). This is load-bearing twice: (a) it is this script's own resumability/only-retry-failures discipline (org data-safety rule — never re-attempt successes, only misses on demand); (b) it is what makes `tracks[]` in #1021 an **implicit three-state signal** — empty = unvisited, non-empty = attempted — which is the agreed "resolved" semantics for the BS consumer (BS#1991: non-empty ⇒ resolved, exits the 30-day sweep). Without attempt rows, the resolved signal cannot be built.

**3. Position recovery is an explicit deliverable, not a nice-to-have.** BS's position-backfill rider (BS#1991, per #801 D7) fills CTA's NULL positions from the `track_position` this repo sends — which only exists where the matcher has joined credit-matches against the **release tracklist** (the #1019 `compilation_track_location` index / discogs-cache track tables). Where the join can't recover a position, send null (`track_position` goes nullable in WXYC/wxyc-shared#297's amendment). If this join isn't built, 78% of `tracks[]` entries carry null positions and the BS rider delivers nothing.

**AC deltas:** add — attempt rows written for misses (NULL `external_id`), asserted by test; storage key survives NULL positions (credit-shaped), with a test importing a position-less fixture compilation; position recovery via the release-tracklist join, with a fixture proving a CTA row with NULL position gains one in the stored/echoed output; re-runs only retry misses (existing org rule, now testable via the attempt rows).

**4. Autonomous-run boundary (added 2026-08-06): the backfill *code* is the deliverable; the backfill *execution* is HITL.** The drain writes prod identity rows and drives Discogs traffic through the shared rate bucket (`risk:prod-write` above). An autonomous implementation run delivers the matcher, the backfill script, and their tests — it must **not** execute the drain against prod. The prod drain is a human-triggered post-merge step, same discipline as WXYC/Backend-Service#1994/#1996. AC clarification: "Backfill drains the existing population" is satisfied at implementation time by the script plus a fixture-scale `pg`-marked test; the production drain itself happens after merge, by hand.

jakebromberg:
Closes #1020

Sub-issue of #271 §2 + §6. Implements the plan at `docs/plans/lml-1020-per-track-identity-matcher.md` (six review rounds, committed in this branch's first commit) in full: an internal per-track **identity** matcher — resolving each `library.db` compilation-track credit (Discogs + MusicBrainz), storing per-source attempt rows in `lml_cache.compilation_track_identity` — plus the backfill script that drains the population, and the retirement of `scripts/match_compilations.py` / `scripts/merge_cta.py`.

Blocks #1021, which blocks WXYC/Backend-Service#1991 (S2 of Backend-Service#801).

## What this delivers

Layer 2 of #271's Option B split. Layer 1 (#1019's `lml_cache.compilation_track_location`) is a *recall/location* index with no external-ID resolution. Layer 2 is what `bulk-resolve`'s future `tracks[]` needs: per-source external IDs, method, and confidence, keyed on the CTA credit itself.

## F1–F4 corrections the plan made to the ticket as filed

The plan's grounding pass found four things that changed stated deliverables (all verified against code, not inferred — see the plan doc for full detail):

- **F1** — `library.db`'s `compilation_track_artist` table has **no position column at all** (three columns: `library_release_id`, `artist_name`, `track_title`). 100% of positions come from a local join against the #1019 recall index (D5), never from CTA. The shape isn't verified anywhere else in the repo (`library/db.py` only checks the table exists), so this backfill fails fast at startup (`validate_cta_shape` / `CtaShapeError`) rather than mid-enumeration.
- **F2** — `library.db`'s `library.id` is the legacy MySQL `LIBRARY_RELEASE_ID`, **not** Backend's `wxyc_schema.library.id`. Documented in the table DDL comment, the generated `.sql` sidecar, and pinned by a dedicated schema test — this plan names the trap for #1021 rather than solving it (that's #1021's read-side problem, bridged by the already-committed `CatalogCompilationTrackRow.legacy_release_id`).
- **F3** — the shared-Discogs-rate-bucket AC is **not** satisfied by holding a `DiscogsService` in-process (the shared PG rate bucket defaults off). The Discogs leg is reached only over `POST /api/v1/artists/resolve/bulk` against the running LML service — never a direct import of `DiscogsService`/`DiscogsCacheService`. Pinned by a static AST-based test (`test_module_makes_no_direct_discogs_calls`) that inspects the module's actual import statements.
- **F4** — the Discogs leg already exists as `BareNameArtistResolver` (LML#759's verify-before-mint cascade). A compilation credit is a bare artist name with no album anchor — exactly that resolver's premise — so this plan reuses it wholesale rather than growing a second matcher.

## Slices (one PR, six commits, in order)

1. **Schema** (`e29835f`) — `lml_cache.compilation_track_identity`: credit-shaped PK `(library_id, track_artist, track_title, source)`, a verdict-coherence CHECK that deliberately excludes `resolved_artist_name`, a partial retry-cohort index. Registered across all four sidecar surfaces (DDL module, generated `.sql`, `main.py` lifespan bootstrap, `regenerate_lml_cache_sql.py`). Also closes a small gap slice 1 left in `test_pg_fixture_guard_adoption.py`'s roster (its own schema test file was undetected by the guard-adoption sweep).
2. **Store read/write** (`765d148`) — the D1 two-statement writer as two functions (`write_compilation_track_identity_verdict` / `write_compilation_track_identity_position`), split because a verdict and its position arrive on different schedules; the `WHERE external_id IS NULL` retry-cohort reader; the per-`library_id` reader #1021 will consume. Neither writer swallows a PG failure — a genuine outage must abort the caller, not get recorded as a mass of false misses.
3. **Matcher core** (`706c73f`, plus a small prerequisite extraction in `ad2ce0e`) — CTA extraction, the D4 whole-then-split credit cascade, the Discogs leg's three-outcome failure contract (`InvalidNameError` → miss row, `escalation_unavailable` → no row at all, a genuine 503 propagates), the D3 method→confidence mapping, and the MusicBrainz leg (similarity floor + deterministic tie-break + ambiguity band). `ad2ce0e` extracts `fetch_mb_artist_candidates` out of `lookup/external_search.py` so the interactive fallback path and this matcher share one query instead of two copies of the same `f_unaccent`-symmetric SQL.
4. **Position recovery** (`a08e791`) — the D5 `LATERAL` join against the #1019 recall index, with the fan-out-collapse determinism a mock can't exercise (Postgres's own `ORDER BY ... LIMIT 1`). Reports recall-index coverage and CTA↔Discogs credit-string agreement as two separate stats.
5. **Fan-back + CLI** (`ae0ad7c`) — `group_credits_by_artist` maps one resolved credit string back to every `(library_id, track_title)` row that carries it; `select_credit_population` implements `--incremental`/`--full`/`--retry-misses` at the `(library_id, track_artist, track_title, source)` grain the plan requires (pinned by a test draining MB-unconfigured then MB-configured); the CLI mirrors the two named precedents (`build_compilation_track_location.py` for the mode flags, `artist_resolve_drain/__main__.py` for `--live`/`--base-url`/`--api-key`/graceful shutdown).
6. **Retirement + docs** (`bac20a4`) — `scripts/merge_cta.py` deleted outright (SSH-wrote to tubafrenzy MySQL, forbidden by the pivot); `scripts/match_compilations.py`'s matching cascade moved via `git mv` to `scripts/_lib/release_matching.py` (now under full mypy enforcement, dropped from `pyproject.toml`'s override list) with its four call sites updated and its standalone CLI deleted; `docs/architecture.md`, `docs/scripts.md`, `CLAUDE.md`'s router line, and a superseding note on `docs/plans/streaming-availability-pg-migration.md`'s PR-E line.

## A deliberate scope decision, flagged for review

The plan's slice 5 names `run_drain` (JSONL-logged resume + escalation-cooldown retry) as an import candidate alongside `make_post_batch`/`resolve_batch`/`PAGE_SIZE`. What this branch actually imports from the sibling drain: `make_post_batch` (the HTTP seam), `PAGE_SIZE` + `chunk` (both resolve passes page to the endpoint cap), `_TERMINAL_REASONS` (the D3 allowlist), and `report.sample_spot_check` (the `--live` gate's review table). `resolve_batch` is reached transitively inside `make_post_batch` rather than imported directly. This branch does **not** drive `run_drain`'s own outer loop — reasoning is in the module docstring: D2 already gives this backfill a stronger, cross-process resumability mechanism than a local JSONL file (every attempt is a durable PG row), so a crash loses at most one in-flight page (≤25 credits) and the next `--incremental`/`--retry-misses` invocation picks up exactly where the process left off, on a different machine if need be. The property this trades away is same-session cooldown-retry for `escalation_unavailable`; an escalated credit still gets no row (D3) and is picked up by the next `--incremental` pass instead of a same-session backoff. Flagging this explicitly since the plan named the import — happy to wire `run_drain` in if review disagrees with the trade-off.

## Measurement pass (read-only; run against what's locally available)

- **Distinct-credit-string count against `library.db`** — could not be measured. None of the `library.db` snapshots available on this machine (newest: a 2026-07-19 prod cutover snapshot) carry the `compilation_track_artist` table at all; per this session's own prior-session notes, CTA export was only restored in discogs-etl on 2026-07-28, after every snapshot on this machine was captured. The current *production* `library.db` almost certainly has it (that's what unblocked this plan), but downloading it wasn't in scope for a read-only local check. **Recommend re-running this count against a post-07-28 `library.db` before sizing a real drain**, per the plan's own sizing-is-a-prerequisite note (`PAGE_SIZE=25` × the full distinct-credit-string count, at ~30s/page against the shared 50/min budget).
- **MB expression-trigram-index check** — could not be measured either: `DATABASE_URL_MUSICBRAINZ` is not configured in this local environment and no local MusicBrainz-cache PG is reachable. Per D3 (and this task's own fallback instruction), noting this as a **musicbrainz-cache prerequisite** rather than blocking: before any real MB-leg drain, confirm expression-trigram indexes exist on `lower(f_unaccent(mb_artist.name))` and `lower(f_unaccent(mb_artist_alias.name))` (`lookup/external_search.py:48-52`'s existing warning) — without them the `%` predicate seq-scans, which one interactive lookup can absorb but a full drain cannot.

## Test evidence

All local CI green (this machine: PostgreSQL 16 via homebrew, no Docker):

- `ruff check .` / `ruff format --check .` — clean.
- `mypy . --ignore-missing-imports` — clean, including `scripts/_lib/release_matching.py` and `scripts/backfill_compilation_track_identity.py` now under full enforcement (no override entries).
- `pytest -m 'not pg and not external_api'` — 5805 passed, 2 skipped, 2 xfailed.
- `pytest -m pg` — 598 passed, 1 failed. The one failure (`test_streaming_catalog.py::TestSchemaBootstrap::test_album_delete_restricted_by_fk_even_with_opt_in`) is **pre-existing and unrelated**: confirmed via `git stash` against this branch's own baseline commit (before any LML#1020 slice-2-6 work), and it's in `entity/streaming_catalog.py`'s test suite, a module this PR never touches. It's specific to this long-lived local, non-Docker PG cluster (not a fresh container), most likely residual state from a prior local session.

New `pg`-marked coverage added by this PR (all green): schema (11), store (11 including the coherence-CHECK end-to-end insert), position recovery (6), full fixture-scale drain (5) — 33 new `pg` tests total, plus 129 new unit tests across the matcher, orchestration, and the `fetch_mb_artist_candidates` extraction.

## Hard boundaries respected

- No live Discogs API call anywhere in this branch or its tests — pinned by a static AST-based import check (F3).
- No minting into `entity.identity` — `--live` defaults off, and every test that exercises the HTTP boundary uses a fake `post_batch`.
- No `entity.*` DDL — `lml_cache.compilation_track_identity` is entirely LML-owned, lifespan-bootstrapped, no discogs-cache coordination.
- Nothing executed against production or any non-fixture population — every `pg` test runs against a local, throwaway schema; the production drain remains a human-triggered post-merge step (`risk:prod-write`), exactly as scoped.

## Not merging this PR

Per the run's instructions, this PR is left open for review — not merged, and auto-merge is not enabled.

## Review round 2 (2026-08-07)

A maximum-scrutiny review pass over the full diff surfaced six confirmed correctness defects plus efficiency/simplification/coverage findings; all fifteen are fixed in the two `fix(#1020)` commits, each within the plan's stated design decisions:

- **Split pass chunked to the endpoint cap** (was: a page of joint credits that all miss whole-string unions >25 names, 413s, and deterministically wedges the drain at the same page on every restart).
- **Page-level persistence** (was: every verdict held in memory until both legs finished the whole population — a crash discarded the entire session's paid Discogs budget). `httpx.HTTPError` now degrades at the page boundary with prior pages persisted, mirroring `run_drain`'s posture; the module docstring's "at most one in-flight page" loss bound is now true.
- **`--recover-positions`** standalone mode (was: position recovery unreachable once a population was fully attempted — the only route to new recall-index positions was a `--full` re-POST of ~140K credits for a join that costs zero external calls).
- **D5 stats de-conflated** — the agreement denominator is scoped to covered comps (`covered_candidate_rows`), so a recall-index coverage problem can no longer read as a CTA↔Discogs string-agreement problem.
- **Split-derived verdicts re-stamped** as `member_group`/0.75 instead of carrying the winning part's `exact_match`/1.00 — a part-match is now distinguishable from (and strictly below) every whole-string tier.
- **Spot-check wired** — every run ends by emitting `sample_spot_check` over this run's `api_search` resolutions, so the `--live` gate's documented human review has an artifact to review.
- Also: the D3 outcome contract is an allowlist over `_TERMINAL_REASONS` (unknown reasons re-ask + warn, never durable negative evidence); MB confidence goes through the single method→confidence table (comparable across `sources[]`) and its ambiguity band only vetoes on runners-up that cleared the floor; `resolve_discogs_credits` defaults to the non-minting mode; `--limit` caps selected work per source (not CTA rows read); the MB leg pre-filters invalid credits into deterministic miss rows; the miss cohort has one definition (the store helper); the per-row position UPDATE fan-out is one set-based statement; and the test suite now covers multi-credit pages, multi-page persistence-under-abort, drained-population recovery, discriminating tie-breaks, and F2's verbatim id-space fixture proof.


jakebromberg:
### Code review

Five-lens verification pass (conventions adherence, shallow bug scan, git-history invariants, prior-PR guidance, code-comment consistency) over the full diff. Found 2 issues, both fixed in e48ff9a:

1. The page-boundary `except httpx.HTTPError` swallowed 4xx responses, contradicting the policy PR #780 established in the sibling drain ("a 4xx is a misconfig, not 'couldn't ask' — re-raise it"): a bad `--api-key` would log one WARNING on the first page and exit 0 having resolved nothing. The catch now re-raises sub-500 `HTTPStatusError`s, pinned by a test injecting a 401 on page 2 that also asserts page 1's rows persisted first.

https://github.com/WXYC/library-metadata-lookup/blob/e48ff9a3e9527c57020e8f3d26d877e541a2517d/scripts/backfill_compilation_track_identity.py#L857-L873

2. The module docstring claimed `make_post_batch`/`resolve_batch`/`PAGE_SIZE` are all imported from the sibling drain; `resolve_batch` never is (it is reached transitively inside `make_post_batch`). Corrected to the actual import list.

https://github.com/WXYC/library-metadata-lookup/blob/e48ff9a3e9527c57020e8f3d26d877e541a2517d/scripts/backfill_compilation_track_identity.py#L30-L36

Clean on everything else checked: `lml_cache`/`entity.*` ownership boundary, lifespan bootstrap ordering, the mypy-override removals (both modules verified clean as moved), sidecar regeneration parity (byte-identical), the `match_compilations.py` cascade move (every historical fix present verbatim, no orphaned importers), fixture-data conventions, and the docs-router updates.

Issue #1117, opened 2026-08-04
Codegen Freshness red on every PR: generated/api_models.py lags wxyc-shared api.yaml 1.29.0 (BS#1965 CTA export)


jakebromberg:
## Problem

The **Codegen Freshness** CI job fails on every open PR, blocking merges for changes that have nothing to do with the API contract. `generated/api_models.py` on `main` lags the `wxyc-shared` `api.yaml` it is generated from.

Confirmed red on at least WXYC/library-metadata-lookup#1113 and WXYC/library-metadata-lookup#1110 — two unrelated PRs, both otherwise fully green (Default Tests, External API Tests, Lint & Format, PG Tests, Type Check, CI Marker Sync all passing). The job is not flaky and is not detecting a real defect in either PR; it is reporting a stale committed artifact on `main`.

## Trigger

`wxyc-shared` commit [`cae6030`](https://github.com/WXYC/wxyc-shared/commit/cae60305766d1b37d3e17178f0a737f6a8eaf15e) — *"feat(api): extend CatalogExportRow + add CTA export for the library.db producer (BS#1965)"*, merged 2026-08-03 16:13 PDT, taking `api.yaml` to version **1.29.0**.

The `codegen-check` job downloads `api.yaml` from `wxyc-shared` `main` at run time ([`scripts/generate_api_models.sh`](https://github.com/WXYC/library-metadata-lookup/blob/545484a40442a3f817aad29a80e9ef668f7ab50e/scripts/generate_api_models.sh) falls back to `https://raw.githubusercontent.com/WXYC/wxyc-shared/main/api.yaml` when no sibling checkout is present, which is always the case in CI). So the moment that spec change merged, every subsequent LML PR started failing — no LML commit caused it and no LML PR can fix it except by regenerating.

## Drift observed

From the CI log on WXYC/library-metadata-lookup#1113, at minimum:

- `CatalogExportRow.label` gains a description — `label: str | None = None` becomes a `Field(None, description="Record label (library.label), projected as its real value...")`.
- A new `CatalogCompilationTrackRow` model appears, with `legacy_release_id`, `artist_name` (`constr(min_length=1, max_length=255)`), and `track_title` (`constr(max_length=255) | None`).

Confirmed absent from the committed file: `grep -c CatalogCompilationTrackRow generated/api_models.py` returns `0` on `main`. Treat the above as a floor, not an exhaustive list — regenerate and read the real diff.

## Desired end state

[...]

PR #1185, opened 2026-08-13
fix(lookup): stop a row-less compilation hit erasing the artist's shelved rows

### LINKED ISSUE #1184: Row-less compilation hit erases the artist's shelved library rows — the step-3b merge-back gate is empty by construction
## Problem

When `/lookup` resolves a track to a compilation that WXYC **does not** shelve, the row-less result *replaces* the artist's own shelved albums in `results` instead of joining them. The caller receives `found_on_compilation: true`, `song_not_found: false`, a context message reading `Found "<song>" by <artist> on:`, and a `results` list whose only member is an `id: 0` row pointing at a release nobody can pull off the shelf. The albums WXYC actually holds by that artist are gone from the response.

Reproduced against **prod LML** on 2026-08-13 (`POST /api/v1/lookup`, artist `Arabian Prince`):

| `song` | `search_type` | `song_not_found` | `found_on_compilation` | `results` |
|---|---|---|---|---|
| `Strange Live` (typo) | `compilation` | `false` | `true` | 1 × `id=0` — Arabian Prince / *Chrome Children Vol. 2* |
| `Strange Life` (correct) | `compilation` | `false` | `true` | 1 × `id=0` — Arabian Prince / *Chrome Children Vol. 2* |
| `Qqzzx Nonexistent Track` | `compilation` | `true` | `false` | `id=99` *Brother Arab*, `id=54515` *Innovative Life: The Anthology 1984-1989* |
| *(artist only, no song)* | `fallback` | `true` | `false` | `id=99`, `id=54515` |

The nonsense-track row is the diagnostic: when Discogs finds **nothing**, the artist fallback works and both shelved albums come back with the correct `"…is not on any album in the library, but here are some albums by Arabian Prince:"` framing. It is the *successful* compilation match that destroys the result. Both spellings behave identically, so fuzzy track matching is working correctly and the typo is not a factor.

The two rows that should have survived are real and shelved:

| library id | artist | title | call number | format |
|---|---|---|---|---|
| 99 | Arabian Prince | Brother Arab | Ar 1/1 | vinyl |
| 54515 | Arabian Prince | Innovative Life: The Anthology 1984-1989 | Ar 1/2 | cd |

(WXYC shelves *Peanut Butter Wolf Presents Chrome Children* — Vol. 1, library id 50018, filed under `Various Artists - Hiphop` — but not Vol. 2, which is why the matched comp resolves row-less.)

## Root cause

1. `ARTIST_PLUS_ALBUM` misses on artist+song, falls through to its artist-only leg, finds both shelved rows, and returns `Outcome.artist_fallback(items)` — `state.results` = the two rows, `song_not_found = True`.
2. That flag gates `TRACK_ON_COMPILATION`, which resolves the track to *Chrome Children Vol. 2*. No WXYC catalog row exists for it, so it is surfaced in the LML#631 / LML#628 row-less shape (`id: 0`, empty `library_url`, `call_number: "(external)"`).
3. [`_apply`](https://github.com/WXYC/library-metadata-lookup/blob/bec8457a9267acf6fdd81e3807f9ed5e596a79d1/core/search.py#L623-L642) stashes the two real rows into `state.artist_fallback_results` (correct — `preserve_prior_results_as_fallback` fires because `state.results` and `state.song_not_found` both hold), then **overwrites** `state.results` with the single row-less item and sets `found_on_compilation = True`, `song_not_found = False`.
4. The stash is never redeemed. [`_step_validate_tracks`](https://github.com/WXYC/library-metadata-lookup/blob/bec8457a9267acf6fdd81e3807f9ed5e596a79d1/lookup/orchestrator.py#L750-L759) computes `real_results = [r for r in state.library_results if r.id != 0]` and gates on it:

```python
real_results = [r for r in state.library_results if r.id != 0]
if state.song_not_found and not state.found_on_compilation:
    widened = [r for r in state.unranked_fallback_candidates if r.id != 0]
    if len(widened) > len(real_results):
        real_results = widened
should_validate = bool(real_results and parsed.song and parsed.artist) and (
    not state.found_on_compilation or bool(state.artist_fallback_results)
)
if not should_validate:
    return
```

On this path `library_results` holds exactly one item and that item has `id == 0`, so `real_results` is empty **by construction**, `should_validate` is `False`, and the step returns before reaching [`apply_track_validation_cascade`](https://github.com/WXYC/library-metadata-lookup/blob/bec8457a9267acf6fdd81e3807f9ed5e596a79d1/lookup/validation.py#L421-L431). The compilation tier of that cascade — the branch written precisely to validate `artist_fallback_results` and prepend confirmed matches ahead of the compilation results — is unreachable whenever the compilation hit is row-less. The widen on line 751 can't rescue it either: it is gated on `not state.found_on_compilation`, which `Outcome.compilation` has already cleared.

Note the second clause of `should_validate` (`not state.found_on_compilation or bool(state.artist_fallback_results)`) was clearly written to admit exactly this case. The first clause silently vetoes it.

## Desired end state

A track that resolves to an unshelved compilation must not erase the artist's shelved albums. The response should carry both the row-less compilation match *and* the artist's own library rows, ordered so the rows a DJ can actually pull are legible, with `context_message` / `found_on_compilation` / `song_not_found` honestly describing that mix.

If the decision is that a row-less comp match should never displace real rows at all, that is also an acceptable end state — but it must be a decision recorded on this issue, not an accident of a gate expression.

## Where

- [`lookup/orchestrator.py#L750-L759`](https://github.com/WXYC/library-metadata-lookup/blob/bec8457a9267acf6fdd81e3807f9ed5e596a79d1/lookup/orchestrator.py#L750-L759) — `_step_validate_tracks`, the gate that is empty by construction.
- [`core/search.py#L623-L642`](https://github.com/WXYC/library-metadata-lookup/blob/bec8457a9267acf6fdd81e3807f9ed5e596a79d1/core/search.py#L623-L642) — `_apply`, the replace-not-merge write site.
- [`lookup/validation.py#L421-L431`](https://github.com/WXYC/library-metadata-lookup/blob/bec8457a9267acf6fdd81e3807f9ed5e596a79d1/lookup/validation.py#L421-L431) — `apply_track_validation_cascade`'s compilation tier, the unreachable merge-back.
- [`lookup/strategies/track_on_compilation.py`](https://github.com/WXYC/library-metadata-lookup/blob/bec8457a9267acf6fdd81e3807f9ed5e596a79d1/lookup/strategies/track_on_compilation.py) — where the row-less item originates.

## Constraints

- **`lookup/orchestrator.py` is frozen for *features* under [Post-launch service hardening (org project #32)](https://github.com/orgs/WXYC/projects/32).** This is a correctness bug in a live consumer-facing path, not a feature, so it is in scope — but keep the diff to the gate/merge semantics and resist adjacent cleanup in that file.
- Must not regress WXYC/library-metadata-lookup#1018 / #1026: when the artist's own release **is** the primary match, folded shelf locations still append *after* the primaries (Case A), and the Case-B prepend ordering in `_fold_locations_into_results` must keep working.
- Must not reintroduce the WXYC/request-o-matic#173 leak: row-less items still need to be identifiable as such (`id == 0`) so downstream callers can filter them from a DJ-facing feed.
- The fix must hold when the artist fallback rows *cannot* be track-validated against Discogs (the typo case above): a failed validation of the artist's own albums is not a reason to return zero rows to a caller who asked about an artist WXYC demonstrably stocks.

## Suggested approach

Two candidate shapes; pick one and record why on the issue.

1. **Relax the gate (smaller diff).** Let `_step_validate_tracks` proceed when `state.found_on_compilation and state.artist_fallback_results` even if `real_results` is empty, so the compilation tier of `apply_track_validation_cascade` runs. This alone is not sufficient — that tier calls `filter_results_by_track_validation(artist_fallback_results, …)`, which will reject *Brother Arab* / *Innovative Life* because Discogs won't confirm the track on them (certainly for the typo, and the anthology's tracklist would need to confirm for the correct spelling). Pair it with a fallback that appends the unvalidated artist rows after the comp row rather than dropping them.
2. **Don't let a row-less result displace real rows (more principled).** In `_apply`, when incoming `outcome.items` are all `id == 0` and `state.results` holds real library rows, merge instead of replace. This fixes the class of bug rather than the one gate, and makes the `should_validate` clause behave as its author evidently intended.

Either way the response signals need to end up self-consistent: a mixed list should not read as a bare `Found "<song>" by <artist> on:` with unshelved-only content above the shelved rows.

## Acceptance criteria

- [ ] `POST /api/v1/lookup` with `artist: "Arabian Prince"`, `song: "Strange Life"` returns the row-less *Chrome Children Vol. 2* match **and** library rows 99 + 54515.
- [ ] The same request with the misspelled `"Strange Live"` also returns rows 99 + 54515 (validation failure must not zero the response).
- [ ] `context_message`, `found_on_compilation`, and `song_not_found` agree with the returned mix — no header claiming a match over a list that contains none of the claimed thing.
- [ ] Regression test at the `execute_search_pipeline` / `perform_lookup` seam covering "row-less comp hit + non-empty `artist_fallback_results`", asserting the real rows survive. The strategies package is constructed with injectable execute funcs (`lookup/strategies/__init__.py::build_strategies`), so this is drivable with `AsyncMock`s and needs no live Discogs.
- [ ] Existing #1018 / #1026 Case-A and Case-B ordering tests still pass.
- [ ] Verified against prod after deploy with the four probes in the table above.

## Related

- WXYC/request-o-matic#256 — the caller-side half: request-o-matic strips `id <= 0` rows but passes `found_on_compilation` / `context_message` through unchanged, so this bug surfaces to DJs as a header with nothing under it.
- WXYC/library-metadata-lookup#1018 — the mirror-image case (the artist's own release hides the compilation copy), resolved via the location union. This issue is the inverse: the compilation copy hides the artist's own releases.
- WXYC/library-metadata-lookup#1026 — the location-union fold and its Case-A/Case-B ordering contract, which any fix here must preserve.
- WXYC/library-metadata-lookup#631, #628, #627 — origin and contract of the row-less result shape.
- WXYC/library-metadata-lookup#959 — adjacent: TRACK_ON_COMPILATION surfacing rows whose specific WXYC pressing lacks the track.

jakebromberg:
## Problem

When `/lookup` resolves a track to a compilation WXYC does not shelve, the row-less result *replaced* the artist's own shelved albums instead of joining them. The caller got `found_on_compilation: true`, a `Found "<song>" by <artist> on:` context, and a single `id: 0` row pointing at a release nobody can pull.

Prod, 2026-08-13, artist `Arabian Prince`:

| `song` | `found_on_compilation` | `results` |
|---|---|---|
| `Strange Live` (typo) | `true` | 1 × `id=0` — *Chrome Children Vol. 2* |
| `Strange Life` (correct) | `true` | 1 × `id=0` — *Chrome Children Vol. 2* |
| `Qqzzx Nonexistent Track` | `false` | `id=99` *Brother Arab*, `id=54515` *Innovative Life: The Anthology 1984-1989* |

The nonsense-track row is the diagnostic: when Discogs finds nothing, the artist fallback works. It was the **successful** compilation match that destroyed the result. Both spellings behave identically — fuzzy track matching was fine and the typo was never the issue.

## Root cause

`_apply` stashes the artist rows into `artist_fallback_results` (correct), then overwrites `state.results` with the row-less item. Step 3b is where that stash is meant to be redeemed, but its gate read:

```python
real_results = [r for r in state.library_results if r.id != 0]
should_validate = bool(real_results and parsed.song and parsed.artist) and (
    not state.found_on_compilation or bool(state.artist_fallback_results)
)
```

On this path `library_results` holds exactly one item and its id is `0`, so `real_results` is empty **by construction**, `should_validate` is `False`, and the compilation tier of `apply_track_validation_cascade` — the branch written precisely to redeem the stash — is unreachable. The second clause was clearly written to admit this case; the first silently vetoed it.

## Fix

The compilation tier validates `artist_fallback_results`, not `library_results`, so it never needed a real row in the latter to run. This gates that shape explicitly, and when validation confirms nothing on a row-less comp, appends the stashed rows behind it rather than dropping them.

Two details worth review attention:

- **The append is capped at `MAX_SEARCH_RESULTS`.** The stash is un-truncated — `search_library_with_fallback` fetches at `_FETCH_LIMIT` (10× the cap) — and every appended row costs a per-item artwork fetch, streaming probe and identity resolve in steps 3c/4/6. A prolific artist would otherwise have blown the response and the tail fan-out open. The confirmed branch inherits the same cap from `filter_results_by_track_validation`; this one has to apply it. Covered by `test_appended_rows_respect_the_result_cap`.
- **The cascade can now overturn the compilation verdict.** `Step3bResult` gains `found_on_compilation: bool | None` (`None` = "leave alone", the same convention `Outcome.song_not_found_after` uses). Only the row-less append sets it, to `False`: the sole row carrying the track is unshelved, so the shelf-facing reading of `Found "<song>" on:` is false. The response adopts the framing this same request already produces when Discogs finds nothing at all — `"<song>" is not on any album in the library, but here are some albums by <artist>:` — now with the artist's albums under it.

Scoped to the row-less case on purpose: when the matched compilation is itself shelved the DJ can pull it, the response is already actionable, and an unconfirmed fallback stays suppressed exactly as before (pinned by `test_shelved_compilation_with_unconfirmed_fallback_is_untouched`).

The row-less comp keeps position 0 so top-1 artwork still binds to the release that actually carries the track (LML#628/#630).

## Tests

New `tests/unit/test_rowless_compilation_artist_rows.py` — 7 tests across both layers (cascade tier + orchestrator gate). Written red first; the 4 that describe the new behavior failed against `main` before the fix. The cap test was independently verified to fail with the cap line removed.

Local: 5,966 passed / 1 skipped / 670 deselected, `ruff check` clean, `ruff format --check` clean, `mypy . --ignore-missing-imports` clean, `scripts/check_plan_links.sh` clean.

`docs/architecture.md`'s step-3b entry is updated — its gate description ("real, non-rowless results present") was the thing this PR falsifies.

## Not in scope

The caller-side half is WXYC/request-o-matic#256: request-o-matic strips `id <= 0` rows but passes `found_on_compilation` / `context_message` through unchanged, so an all-row-less response still renders a header over an empty list there. Independent fix; neither blocks the other.

## Verification

Acceptance criteria 1–2 on the issue are prod probes; they'll be run after this deploys to staging and prod.

Closes #1184


jakebromberg:
CI note: **Codegen Freshness** is red on this branch and is not caused by this change. It is the repo-wide drift tracked in #1117 — CI regenerates `generated/api_models.py` against wxyc-shared `main` at run time, and upstream has since added `AppConfig.donateUrl` / `AppConfig.donateEnabled` (the donation-button work) that the committed snapshot predates. The diff CI reports is exactly those two fields and nothing from this PR, which touches no generated file and no contract.

Every other job is green: Default Tests, Lint & Format, Type Check, PG Tests, External API Tests, CI Marker Sync.

Regenerating here would fold an unrelated contract bump into a bug fix, so it is deliberately left to #1117.

PR #1207, opened 2026-08-17
fix(lookup): let the library-miss 80/80 floor clear a Discogs disambiguation suffix

### LINKED ISSUE #1206: Discogs disambiguation suffix defeats search recall: bare-name queries never retrieve a 'Name (N)' credit (MAVI / 'Mavi (12)')
## Problem

A Discogs **disambiguation suffix** on the cached artist credit defeats search recall. When Discogs credits an artist as `Name (N)` — the suffix it assigns whenever multiple artists share a name — a caller who types the bare name resolves nothing, even though the release is present in discogs-cache as a main credit.

Measured against prod LML (`/api/v1/lookup/bulk?allow_release_resolution_fallback=true`, the enrichment worker's exact configuration) on 2026-08-17:

| Query | `search_type` | results |
|---|---|---|
| `MAVI` / *The Pilot* / `31 Days` | `compilation` | **0** |
| `Mavi` / *The Pilot* / `31 Days` | `compilation` | **0** |
| **`Mavi (12)`** / *The Pilot* / `31 Days` | `compilation` | **1** ✅ |

Case is not the variable — both `MAVI` and `Mavi` miss. The suffix is.

The release is in prod discogs-cache and reachable:

```sql
SELECT r.id, r.title, ra.artist_name
FROM release r JOIN release_artist ra ON ra.release_id = r.id AND ra.extra = 0
WHERE r.id = 37193856;
--  37193856 | The Pilot | Mavi (12)
```

### Real-world impact

Two playcuts in the 2026-08-16 `metadata-no-match-digest` are directly caused by this, both from the same rotation album:

- Backend-Service `flowsheet` #5308909 — `MAVI — 31 Days · The Pilot · Loma Vista Recordings`
- Backend-Service `flowsheet` #5308853 — `MAVI — G-ANNIS FREESTYLE · The Pilot · Loma Vista Recordings`

Both wrote `metadata_status = 'enriched_no_match'`, which is terminal in Backend-Service — so they stay unenriched permanently (see WXYC/Backend-Service#2176). This is a rotation album, so the miss is DJ-visible on the flowsheet and in the iOS playlist.

Because the trigger is "Discogs disambiguated this artist name", the blast radius is every common or colliding artist name in the WXYC library — not a one-off.

## Why this is new

WXYC/library-metadata-lookup#784 enumerated the free-text Discogs recall categories (multi-artist `A & B` strings, V/A-comp features, typos, generic `S/T` titles) and was closed. **The disambiguation suffix is not one of them.** WXYC/library-metadata-lookup#1063's audit concluded the residual recall gap was "cache coverage, not matcher code" — that verdict holds for most of the 2026-08-16/17 digest rows, but not this one: the release *is* cached, and the correct suffixed form resolves it on demand.

## Where

- [`discogs/matching.py#L41-L63`](https://github.com/WXYC/library-metadata-lookup/blob/49655c2/discogs/matching.py#L41-L63) — `strip_discogs_suffix`, delegating to `wxyc_etl.text.strip_discogs_disambiguation(name, broad=False)` (adopted in WXYC/library-metadata-lookup#1042). This primitive exists and is correct; it is applied at **scoring/validation**, after candidates are already in hand.
- [`core/search.py#L1046-L1077`](https://github.com/WXYC/library-metadata-lookup/blob/49655c2/core/search.py#L1046-L1077) — `get_search_type_from_state`; the `ARTIST_PLUS_ALBUM` strategy is the path these queries take.
- [`lookup/orchestrator.py`](https://github.com/WXYC/library-metadata-lookup/blob/49655c2/lookup/orchestrator.py) — step 3a `_library_miss_discogs_search` (the LML#583/#784 library-miss probe).

**These are candidate loci, not a diagnosis.** The recall path was not fully traced — what is established is the input/output behavior in the table above. The implementer should confirm where the caller's artist string is matched against the cached/API credit before choosing a fix site.

## Suggested approach

The asymmetry to close: `strip_discogs_disambiguation` is applied to the **candidate** side but the **query** side never tries the suffixed form, so a bare-name query never retrieves the suffixed row in the first place. Options, roughly in order of preference:

1. Make the cache/API lookup match on a **folded** artist key (suffix stripped on both sides at retrieval time), rather than stripping only when scoring an already-retrieved candidate. This is the symmetric fix and mirrors the treatment described in `lookup/artist_resolution.py#L100-L124`, which already strips both sides.
2. If retrieval must stay exact-keyed, add a second retrieval attempt against `artist_name ILIKE '<name> (%'` when the bare-name attempt returns nothing.

Prefer whichever keeps the `broad=False` (numeric-only) semantics at retrieval — `broad=True` also strips country/qualifier suffixes and is a wider behavior change than this ticket needs.

## Acceptance criteria

- [ ] `MAVI` / *The Pilot* / `31 Days` resolves to Discogs release `37193856` through `/api/v1/lookup` **and** `/api/v1/lookup/bulk?allow_release_resolution_fallback=true`.
- [ ] The same holds for a lowercase (`mavi`) and an uppercase (`MAVI`) caller form.
- [ ] Regression test covering the bare-name → `Name (N)` retrieval path, parameterized over at least a numeric suffix and one non-colliding control (a bare name whose cached credit has **no** suffix must not regress).
- [ ] No measurable recall regression on the existing matcher-quality suite; confirm the fix does not let a bare name match the *wrong* numbered artist (e.g. `Mavi (12)` vs a hypothetical `Mavi (3)`) — if the cache holds multiple suffixed variants of one bare name, the ambiguity must resolve conservatively (album title must still confirm) rather than picking the first.
- [ ] Re-run the 2026-08-16 digest's Section A rows; #5308909 and #5308853 resolve.

## Notes for implementer

- Reproduce with `POST /api/v1/lookup` and body `{"artist": "MAVI", "album": "The Pilot", "song": "31 Days"}` against prod, then swap `artist` to `Mavi (12)` — no auth beyond the standard `Authorization: Bearer $LML_API_KEY`.
- Prod discogs-cache is WXYC-filtered (138,496 releases), so "absent from that cache" never proves "absent from Discogs". This ticket concerns only names that **are** present.
- Unrelated observation surfaced by the same replay, filed separately for triage rather than bundled here: `search_type` comes back as `compilation` on **zero-result** responses (all 17 Section A rows), where `api.yaml`'s documented default is `none`. Harmless to Backend-Service today only because its `extractArtwork` walks the `results` array instead of trusting `search_type` alone. Not in scope for this fix — do not change it here.

## Related

- WXYC/library-metadata-lookup#784 — predecessor recall-gap ticket; this is a category it did not enumerate
- WXYC/library-metadata-lookup#1063 — the audit whose "cache coverage, not matcher code" verdict this amends for one row class
- WXYC/library-metadata-lookup#1042 — adopted the `wxyc_etl` disambig-strip primitive this ticket asks to apply at retrieval
- WXYC/library-metadata-lookup#1181, WXYC/library-metadata-lookup#1182, WXYC/library-metadata-lookup#1187 — sibling `concern:matcher-quality` recall/scoring-axis bugs
- WXYC/Backend-Service#2176 — why a miss here is permanent rather than self-healing

jakebromberg:
Closes #1206

## Problem

`/api/v1/lookup` returns **0 results** for `MAVI` / *The Pilot* / `31 Days`, but **1 result** for `Mavi (12)` / *The Pilot* / `31 Days`. `Mavi (12)` is how Discogs credits the artist — the `(N)` disambiguation suffix Discogs assigns whenever several artists share a name. Case is not the variable; both `MAVI` and `Mavi` miss. The suffix is.

Two real WXYC playcuts were lost to this, both from the rotation album *The Pilot*: Backend-Service `flowsheet` #5308909 (`MAVI — 31 Days`) and #5308853 (`MAVI — G-ANNIS FREESTYLE`). Both wrote `metadata_status = 'enriched_no_match'`, which is terminal in Backend-Service (WXYC/Backend-Service#2176), so they stay unenriched permanently. Because the trigger is "Discogs disambiguated this artist name", the blast radius is every common or colliding artist name in the WXYC library, not a one-off.

## History: the retrieval framing was measured and refuted

**Read this before the diff.** The issue's own `## Suggested approach` section is superseded — it proposed closing an assumed *retrieval* asymmetry (fold the artist key at retrieval time, or fire a second `ILIKE '<name> (%'` query when the bare-name attempt returns nothing). Both options address a gap that does not exist. Measured against the prod discogs-cache (recorded in full in [the correction comment on #1206](https://github.com/WXYC/library-metadata-lookup/issues/1206#issuecomment-5318365906)):

The **unmodified** `_SEARCH_BY_TRACK_ARTIST_SQL` CTE (`discogs/cache_service.py`) already returns the right row for the bare name:

| query artist | limit | rows | release 37193856 present | window full? |
|---|---|---|---|---|
| `Mavi` | 20 | 1 | yes | no |
| `Mavi` | 200 | 1 | yes | no |
| `MAVI` | 20 | 1 | yes | no |
| `Mavi (12)` | 20 | 1 | yes | no |

The window never fills at either limit, so candidate crowding is not the mechanism. `search_releases`' artist+album query (the `ARTIST_PLUS_ALBUM` path) likewise returns the release for the bare name:

| query artist | rows | release 37193856 | score |
|---|---|---|---|
| `MAVI` | 1 | yes | **0.812** |
| `Mavi` | 1 | yes | **0.812** |
| `Mavi (12)` | 1 | yes | **1.000** |

The `pg_trgm` threshold is 0.3 and `similarity('mavi','mavi (12)') = 0.625`, so the bare name clears the trigram floor easily. **Retrieval is ruled out at every SQL layer.** Nothing in the cache-query layer needed changing, and a retrieval-side fold or fallback query would have been dead code.

## Root cause

The discard happens one layer downstream, in scoring. `_floor_best()` in `lookup/strategies/library_miss.py` scores the query artist against the **raw, unstripped** Discogs credit via `score_match` (`clients/streaming/matching.py`, plain `fuzz.token_sort_ratio` over normalized forms), and `is_acceptable_match` requires **both** axes ≥ `SCORE_MATCH_ACCEPTANCE_FLOOR = 80.0`. Verified by running the repo's own code:

```
SCORE_MATCH_ACCEPTANCE_FLOOR = 80.0
score_match('MAVI',      'Mavi (12)') =  61.538  -> REJECT
score_match('Mavi',      'Mavi (12)') =  61.538  -> REJECT
score_match('mavi',      'Mavi (12)') =  61.538  -> REJECT
score_match('Mavi (12)', 'Mavi (12)') = 100.000  -> PASS
```

61.5 fails the floor; 100 passes. That is the entire bug, and it explains the exact prod signature: the candidate was present at score 0.812 and the response was still empty. It also explains why every arm behaves identically — the PG cache arm and the live API arm both surface the same raw suffixed credit and floor-fail on it the same way.

## What changed

`_artist_variants_with_stripped_suffix()` widens the **candidate-side** artist variant list passed to `find_best_typed_match`'s `artist_fn` with each raw variant's disambiguation-suffix-stripped form, via `discogs.matching.strip_discogs_suffix` (`broad=False`, numeric-only). The stripped form is added as an *extra* variant, never a replacement, so a caller who already types the suffixed form keeps matching the raw variant at 100 exactly as before. The artist axis takes the max across variants, so the bare name now clears via its own exact match.

`broad=False` is deliberate and preserved: non-numeric qualifier suffixes are never widened. Confirmed directly — `score_match('MAVI', 'Mavi (UK)')` stays **61.538**, i.e. `(UK)` continues to be treated as ordinary title content, deliberately narrower than `clients.streaming.matching.strip_discogs_disambig` (`broad=True`, used elsewhere for canonical-artist resolution).

**The title axis is untouched.** This can never mint a match on the artist axis alone. When the cache holds multiple suffixed variants of one bare name, every variant's artist axis can clear via its own stripped form, but only the one whose album also matches the query clears `is_acceptable_match` — the existing album confirmation stays the conservative gate, which is what the ticket's conservative-resolution acceptance criterion asks for. `tests/unit/test_library_miss_discogs.py::TestDisambiguationSuffixRecall::test_conservative_multi_variant_resolution` pins exactly that with the ticket's hypothetical `Mavi (3)` on an unrelated album.

Tests added: a parameterized bare-name reproduction over `MAVI` / `Mavi` / `mavi`, a no-suffix control (a bare name whose cached credit has no suffix must not regress), the multi-variant conservative-resolution case, a `(UK)` qualifier-suffix negative, an end-to-end replay through the real `perform_lookup` spine, and a live-API companion in `tests/integration/test_library_miss_recall_live.py` (`external_api`-marked).

## Verification

```
uv run ruff check .            -> All checks passed!
uv run ruff format --check .   -> 587 files already formatted
uv run pytest -n auto -q       -> 6422 passed, 1 skipped, 2 xfailed
git diff --stat origin/main    -> 3 files changed, 217 insertions(+), 1 deletion(-)
```

Local `-m pg` noise (14 failures across `artist_candidate_sets`, `artist_resolve_bulk`, `entity_resolution`) is a local Homebrew PG@16 missing `wxyc_unaccent.rules`. Confirmed pre-existing by stashing this branch and re-running against unmodified `origin/main` — **not** caused by this change. The `external_api`-marked live replay self-skips without `DISCOGS_TOKEN`.

## Open question for the reviewer (scope, not a defect)

`score_match` / `find_best_typed_match` in `clients/streaming/matching.py` are shared infrastructure, reused by callers well outside this path — Bandcamp, Apple Music, YouTube Music, artwork matching, and Wikipedia URL picking. Each of those plausibly hits the same raw-credit gap, since none of them strips a Discogs disambiguation suffix before scoring either.

This PR deliberately does **not** sweep them. The fix is applied at the one call site where the bug was measured, keeping the blast radius to the library-miss path and leaving the shared scorer's behavior unchanged for every other caller. The alternative — pushing the strip down into `score_match` itself — would fix all callers at once but silently changes scoring for paths that were never measured, and would need its own evidence.

Worth a reviewer's opinion on whether that restraint is right, or whether a follow-up ticket should sweep the other callers deliberately. Flagging it rather than deciding it unilaterally.

### DIFF
--- a/lookup/strategies/library_miss.py
+++ b/lookup/strategies/library_miss.py
@@ -14,6 +14,7 @@ Strategy-adjacent, so it lives in this package (LML#727).
 import logging
 
 from clients.streaming.matching import find_best_typed_match
+from discogs.matching import strip_discogs_suffix
 from discogs.models import DiscogsSearchRequest, DiscogsSearchResponse, DiscogsSearchResult
 from discogs.service import DiscogsService
 from library.models import LibraryItem
@@ -22,6 +23,40 @@ from lookup.strategies.va_rescue import find_va_comp_match
 from services.parser import ParsedRequest
 
 logger = logging.getLogger(__name__)
+
+
+def _artist_variants_with_stripped_suffix(result: DiscogsSearchResult) -> list[str | None]:
+    """Candidate-side artist scoring variants, widened with each raw variant's
+    Discogs-disambiguation-suffix-stripped form (LML#1206).
+
+    Discogs assigns a trailing ``(N)`` suffix to a cached credit whenever
+    multiple Discogs artists share a name (``"Mavi (12)"``). The 80/80
+    floor's fuzzy comparison (``score_match`` -> ``fuzz.token_sort_ratio``)
+    treats that suffix as ordinary title content: ``score_match("MAVI",
+    "Mavi (12)")`` measures 61.5, under the 80.0 acceptance floor, even
+    though retrieval already had the right release in hand (every arm --
+    PG-cache and live API alike -- surfaces the same raw suffixed credit and
+    floor-fails identically). A bare-name query for such a credit returned no
+    results on every arm.
+
+    Adding the numeric-only (``broad=False``) stripped form as an *extra*
+    variant lets a bare-name query clear via its own exact match, without
+    disturbing the raw variant (a caller who already types the suffixed form
+    keeps matching that too) and without widening to non-numeric qualifier
+    suffixes like ``"(UK)"`` -- those score exactly as before, deliberately
+    narrower than ``clients.streaming.matching.strip_discogs_disambig``
+    (``broad=True``, used elsewhere for canonical-artist resolution).
+
+    The title axis is untouched, so this can never mint a match on the
+    artist axis alone: when the cache holds multiple suffixed variants of
+    one bare name, every variant's artist axis can clear via its own
+    stripped form, but only the one whose album also matches the query
+    clears ``is_acceptable_match`` -- the existing album confirmation stays
+    the conservative gate, not this variant list.
+    """
+    raw_variants = result.artist_variants()
+    stripped_variants = [strip_discogs_suffix(v) for v in raw_variants if v]
+    return [*raw_variants, *stripped_variants]
 
 
 async def _library_miss_discogs_search(
@@ -90,7 +125,9 @@ async def _library_miss_discogs_search(
             response.results,
             query_artist=artist,
             query_title=album,
-            artist_fn=lambda r: r.artist_variants(),
+            # LML#1206: widened with the suffix-stripped form of each
+            # candidate variant -- see the function's own docstring.
+            artist_fn=_artist_variants_with_stripped_suffix,
             title_fn=lambda r: r.album,
             # LML#1097: deterministic tie-break by release_id, mirroring
             # release_resolution.py's (-score, release_id) sort key.
--- a/tests/integration/test_library_miss_recall_live.py
+++ b/tests/integration/test_library_miss_recall_live.py
@@ -123,15 +123,38 @@ async def test_tasquier_comp_resolves_via_tracklist_credit(service):
 
 @pytest.mark.asyncio
 async def test_mavi_disambiguation_suffix_resolves(service):
-    """LML#1206: the bare name ``"MAVI"`` must resolve *The Pilot* (release
-    37193856), cached under Discogs' disambiguation-suffixed credit
-    ``"Mavi (12)"`` — the live-API companion to the mocked reproduction in
+    """LML#1206: the bare name ``"MAVI"`` must resolve *The Pilot*, credited
+    under Discogs' disambiguation-suffixed form ``"Mavi (12)"`` — the live-API
+    companion to the mocked reproduction in
     ``tests/unit/test_library_miss_discogs.py::TestDisambiguationSuffixRecall``.
+
+    Both *The Pilot* pressings are correct and share master 4206252: 35940307
+    is the 2025-11-25 digital release, 37193856 the 2026-02-27 vinyl LP. Their
+    tracklists are identical and both carry the ticket's two victim tracks
+    (``31 Days`` and ``G-Annis Freestyle``), so either satisfies the enrichment
+    the ticket is about. The API arm this module exercises (no PG cache
+    attached) surfaces both at a 100/100 tie, which LML#1097 breaks
+    deterministically by ascending ``release_id`` — making 35940307 the
+    guaranteed pick here, while the ticket's prod measurement named 37193856
+    because the PG cache arm holds that pressing. Pinning the set rather than
+    one id mirrors ``test_anadol_and_marie_klock_resolves_either_pressing`` and
+    ``test_matmos_self_titled_placeholder_resolves`` above.
+
+    The credit assertion is what carries this test's protocol value: the
+    resolved candidate must be the *suffixed* credit, which is precisely what
+    could not clear the 80/80 floor before this fix
+    (``score_match("MAVI", "Mavi (12)")`` = 61.5).
     """
     result = await _library_miss_discogs_search(
         _parsed("MAVI", "The Pilot"),
         discogs_service=service,
     )
-    assert result is not None, "expected release 37193856; got no floor-clearing candidate"
+    assert result is not None, "expected a The Pilot pressing; got no floor-clearing candidate"
     _, best = result
-    assert best.release_id == 37193856, f"resolved wrong release {best.release_id}"
+    assert best.release_id in {35940307, 37193856}, f"resolved wrong release {best.release_id}"
+    assert best.artist == "Mavi (12)", (
+        f"expected the disambiguation-suffixed credit this fix targets; got {best.artist!r}"
+    )
+    assert (best.album or "").strip().lower() == "the pilot", (
+        f"resolved the wrong album {best.album!r}"
+    )
--- a/tests/integration/test_library_miss_recall_live.py
+++ b/tests/integration/test_library_miss_recall_live.py
@@ -119,3 +119,19 @@ async def test_tasquier_comp_resolves_via_tracklist_credit(service):
     assert result is not None, "expected release 27518829; got no candidate past floor or rescue"
     _, best = result
     assert best.release_id == 27518829, f"resolved wrong release {best.release_id}"
+
+
+@pytest.mark.asyncio
+async def test_mavi_disambiguation_suffix_resolves(service):
+    """LML#1206: the bare name ``"MAVI"`` must resolve *The Pilot* (release
+    37193856), cached under Discogs' disambiguation-suffixed credit
+    ``"Mavi (12)"`` — the live-API companion to the mocked reproduction in
+    ``tests/unit/test_library_miss_discogs.py::TestDisambiguationSuffixRecall``.
+    """
+    result = await _library_miss_discogs_search(
+        _parsed("MAVI", "The Pilot"),
+        discogs_service=service,
+    )
+    assert result is not None, "expected release 37193856; got no floor-clearing candidate"
+    _, best = result
+    assert best.release_id == 37193856, f"resolved wrong release {best.release_id}"
--- a/tests/unit/test_library_miss_discogs.py
+++ b/tests/unit/test_library_miss_discogs.py
@@ -1420,3 +1420,166 @@ class TestLibraryMissNoPoison:
         )
         assert result is None
         assert mock_discogs_service.search.call_count == 2
+
+
+# ---------------------------------------------------------------------------
+# LML#1206 — Discogs disambiguation suffix defeats the 80/80 floor
+# ---------------------------------------------------------------------------
+
+
+class TestDisambiguationSuffixRecall:
+    """A Discogs disambiguation suffix on the cached artist credit
+    (``"Mavi (12)"``, the ``(N)`` Discogs appends whenever multiple artists
+    share a name) defeated the 80/80 floor for a bare-name query. Measured
+    against prod: ``MAVI`` / ``Mavi`` / *The Pilot* / ``31 Days`` returned
+    zero results, while ``Mavi (12)`` / *The Pilot* / ``31 Days`` returned one
+    (release 37193856).
+
+    Root cause traced to ``_floor_best``'s call into
+    ``find_best_typed_match``: the artist axis scores
+    ``score_match(query_artist, r.artist_variants())`` — plain
+    ``fuzz.token_sort_ratio`` over the RAW candidate credit, with no
+    disambiguation-suffix strip on either side. ``score_match("MAVI", "Mavi
+    (12)")`` measures 61.5, well under the shared 80.0
+    ``SCORE_MATCH_ACCEPTANCE_FLOOR`` (``clients/streaming/matching.py``), so
+    ``is_acceptable_match`` rejects the candidate outright — retrieval
+    already had the right release in hand (the PG cache's own
+    trigram-similarity ordering clears its floor comfortably at this pair),
+    but this scoring gate discarded it before it ever reached the response.
+    ``score_match("Mavi (12)", "Mavi (12)")`` measures 100, which is why the
+    exact suffixed query worked.
+    """
+
+    RELEASE_ID = 37193856
+    ALBUM = "The Pilot"
+    SUFFIXED_ARTIST = "Mavi (12)"
+
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize("query_artist", ["MAVI", "Mavi", "mavi"])
+    async def test_bare_name_resolves_suffixed_credit(self, mock_discogs_service, query_artist):
+        """Every case form of the bare name must retrieve the suffix-credited
+        release — case was never the variable in the ticket's prod measurement,
+        the suffix was."""
+        mock_discogs_service.search.return_value = DiscogsSearchResponse(
+            results=[
+                make_discogs_result(
+                    release_id=self.RELEASE_ID,
+                    artist=self.SUFFIXED_ARTIST,
+                    album=self.ALBUM,
+                )
+            ]
+        )
+        result = await _library_miss_discogs_search(
+            _parsed(query_artist, self.ALBUM), discogs_service=mock_discogs_service
+        )
+        assert result is not None, (
+            f"query_artist={query_artist!r} did not resolve the "
+            f"{self.SUFFIXED_ARTIST!r}-credited release despite the exact album match"
+        )
+        _, discogs_result = result
+        assert discogs_result.release_id == self.RELEASE_ID
+
+    @pytest.mark.asyncio
+    async def test_no_suffix_control_unaffected(self, mock_discogs_service):
+        """AC — a bare name whose cached credit has NO suffix must not regress."""
+        mock_discogs_service.search.return_value = DiscogsSearchResponse(
+            results=[
+                make_discogs_result(release_id=self.RELEASE_ID, artist="Mavi", album=self.ALBUM)
+            ]
+        )
+        result = await _library_miss_discogs_search(
+            _parsed("Mavi", self.ALBUM), discogs_service=mock_discogs_service
+        )
+        assert result is not None
+        _, discogs_result = result
+        assert discogs_result.release_id == self.RELEASE_ID
+        assert discogs_result.artist == "Mavi"
+
+    @pytest.mark.asyncio
+    async def test_conservative_multi_variant_resolution(self, mock_discogs_service):
+        """AC — when the cache holds multiple suffixed variants of one bare
+        name, the album axis is what disambiguates: only the variant whose
+        album ALSO matches the query is accepted. The artist axis alone must
+        never mint a match — a hypothetical ``"Mavi (3)"`` on an unrelated
+        album must not win just because its stripped form also reads
+        "Mavi"."""
+        mock_discogs_service.search.return_value = DiscogsSearchResponse(
+            results=[
+                make_discogs_result(
+                    release_id=self.RELEASE_ID,
+                    artist=self.SUFFIXED_ARTIST,
+                    album=self.ALBUM,
+                ),
+                make_discogs_result(
+                    release_id=99999999,
+                    artist="Mavi (3)",
+                    album="A Completely Different Album",
+                ),
+            ]
+        )
+        result = await _library_miss_discogs_search(
+            _parsed("Mavi", self.ALBUM), discogs_service=mock_discogs_service
+        )
+        assert result is not None
+        _, discogs_result = result
+        assert discogs_result.release_id == self.RELEASE_ID, (
+            "the hypothetical Mavi (3) / unrelated-album candidate must not win — "
+            f"got {discogs_result.release_id}"
+        )
+
+    @pytest.mark.asyncio
+    async def test_qualifier_suffix_not_widened(self, mock_discogs_service):
+        """AC — ``broad=False`` (numeric-only) semantics: a non-numeric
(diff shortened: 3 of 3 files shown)


jakebromberg (commented):
## Review

Reviewed against `origin/main`, with each finding checked by executing the code rather than reading it. The fix does what it claims — the root-cause analysis is correct and the retrieval-vs-scoring correction in the description is the right call. Findings below are ranked; the first two are worth resolving before merge, the rest are follow-ups or notes.

### Blast radius is length-dependent, and that belongs in the record

Not a defect, but the single most useful thing I measured, because it bounds everything else. The suffix penalty is a function of how much of the string the suffix occupies:

| candidate credit | `score_match(bare, credit)` | verdict |
|---|---|---|
| `Mavi (12)` | 61.5 | FAIL |
| `Fust (12)` | 61.5 | FAIL |
| `Bright (12)` | 70.6 | FAIL |
| `Stereolab (12)` | 78.3 | FAIL |
| `Jessica Pratt (12)` | 83.9 | **PASS** |
| `Chuquimamani-Condori (12)` | 88.9 | **PASS** |

So this bug only ever bit **short artist names** — roughly ≤ 12 characters. Long names already cleared the floor on the raw credit. That explains why it survived until a 4-character name hit it, and it means the population this PR changes behavior for is much narrower than "every disambiguated artist". Worth stating in the code, because the next reader will otherwise assume the fix is universal.

### 1. HIGH — the docstring's central safety claim is false on the self-titled path

`_artist_variants_with_stripped_suffix`'s docstring says the album axis is an independent gate: *"The title axis is untouched, so this can never mint a match on the artist axis alone … the existing album confirmation stays the conservative gate."* Twenty lines below, at `library_miss.py:116-117`:

```python
if is_self_titled(album):
    album = artist
```

On that path the title axis **is a copy of the artist name**, so it confirms nothing independent. Confirmed by execution — query `artist="Bright"`, `album="s/t"` against a single candidate `("Bright (3)", "Bright")`:

```
after swap: query_artist='Bright' query_title='Bright'
artist raw=75.0  widened=100.0  title=100.0
PRE-FIX  winner: None
POST-FIX winner: 500
```

`(3)` is Discogs stating positively that this is a *different artist*, and it is now accepted with no independent evidence.

**Severity nuance, from the control I ran second.** This collapse is **pre-existing**, not introduced here: an *unsuffixed* candidate `("Bright", "Bright")` already matched that same query on `origin/main` with equally zero independent confirmation. LML#784 category 4 accepted that tradeoff knowingly — the comment at 110-115 guards only against the narrower "candidate literally titled S/T" case. So this PR **widens the population reaching an already-unguarded path**; it does not create the hole. That is why I rank it HIGH rather than blocking-critical.

The defect that is unambiguously this PR's, though, is the **docstring**: it asserts a safety property the code does not have, and future readers will build on it. Minimum fix is to scope the claim honestly ("the album axis confirms except on the self-titled path, where it is a copy of the artist name"). The stronger fix — skipping the widening when `is_self_titled(parsed.album)`, since the suffix is the only remaining discriminator there — is a real behavior change and deserves its own evidence, so I would not fold it in silently.

### 2. MEDIUM — newly-reachable ranking flip when a bare and a suffixed credit share an album

Confirmed by execution. Candidates `[(900, "Mavi", "The Pilot"), (100, "Mavi (12)", "The Pilot")]`, query `("Mavi", "The Pilot")`:

```
PRE-FIX  winner: 900  ("Mavi", the exact-name credit)
POST-FIX winner: 100  ("Mavi (12)")
```

Both now tie at 100/100, so LML#1097's ascending-`release_id` tie-break decides — not artist correctness. An exact bare-name credit can now lose to a different, numerically-disambiguated artist purely on an id comparison.

`TestDisambiguationSuffixRecall::test_conservative_multi_variant_resolution` does not cover this: its second variant carries an *unrelated* album, so the album axis still does the work there. The uncovered case is the same-album one, which is precisely the colliding case. Worth either a tie-break preference for the exact (unstripped) credit match, or an explicit test pinning the accepted behavior.

### 3. MEDIUM — the raw suffixed credit propagates to the caller

`library_miss.py` returns:

```python
library_item = LibraryItem(id=0, artist=best.artist or artist, title=best.album or album)
```

`best.artist` is the raw `"Mavi (12)"`, which flows through `to_catalog_item()` → `LibraryCatalogItem.artist` and `to_match_result()` → `EnrichedDiscogsMatchResult.artist`. The new live test now asserts exactly this (`best.artist == "Mavi (12)"`). For a bare-name query this string was previously unreachable on this path; it is now the intended common case.

Backend-Service persists that into flowsheet metadata, so the fix that rescues flowsheet #5308909 would label it artist **`Mavi (12)`** on the flowsheet and in the iOS playlist — a Discogs-internal disambiguator surfacing as a DJ-visible artist name. That partially undercuts the user-facing goal. Either strip the suffix on the outbound `artist` (keeping the raw form for scoring only), or confirm with Backend-Service that the suffixed display form is acceptable. This is a product decision, not a code defect, but it should be made deliberately rather than inherited.

### 4. MEDIUM — the same gap is left in `lookup/artwork.py`, on the more common path

`lookup/artwork.py:~333` has the identical shape — `find_best_typed_match(..., artist_fn=lambda r: r.artist_variants(), title_fn=lambda r: r.album)` against unstripped candidate credits. That is the path for library **hits**, which is where a rotation album like *The Pilot* spends most of its life, and it sits squarely in the post-launch-hardening artwork-outage class.

**One correction to flag.** The nearby comment claims *"Numeric disambigs clear via `token_sort_ratio`'s tolerance"*, and it would be easy to read this PR's 61.5 measurement as refuting it. Measured, it does not — that comment is scoped to the compilation branch, where the base string is long:

```
score_match('Various Artists', 'Various Artists (2)')  = 88.2  PASS
score_match('Various Artists', 'Various Artists (12)') = 85.7  PASS
score_match('Various',         'Various Artists (2)')  = 53.8  FAIL
```

So the comment is **correct within the case it describes**, and per the length-dependence table above. The gap in `artwork.py` is real, but it bites short library artist names, not the "Various Artists" case the comment is about. Worth a follow-up ticket rather than expanding this PR — and worth naming `artwork.py` specifically, since the PR's stated open question is scoped to Bandcamp / Apple Music / YouTube Music (different scorers, genuinely more risk) and does not mention the same function one file over.

### 5. LOW — no dedup; the common case pays double

`return [*raw_variants, *stripped_variants]` never dedups, and for the overwhelming majority of credits (no suffix) the two halves are byte-identical. Measured on a 15-credit candidate with no suffixes: 15 variants become 30, of which 15 are provably redundant, and `find_best_typed_match`'s `max(score_match(q, v) for q in ... for v in ...)` scores each twice.

Cost is genuinely small in absolute terms — variant construction went 0.08 µs → 2.22 µs per candidate, and a full artist-axis pass roughly doubles — so this is not a hot-path hazard at a 20-candidate window. But it is free to fix, and `clients/streaming/matching.py::normalize_artist_credit` already models this exact "raw + numeric-strip" generation with a `seen` set. `dict.fromkeys` over the concatenation would preserve order and drop the redundancy.

### Verified as sound (no action)

These are the properties I set out to break and could not:

- **`broad=False` is genuinely preserved.** `(UK)`, `(US)`, `(Band)` are never stripped; only ASCII-numeric suffixes are. `score_match("MAVI", "Mavi (UK)")` stays 61.5, so a qualifier suffix is not promoted. `"Mavi (12) (UK)"` is also left intact (the numeric group is not terminal).
- **The album axis does confirm on the normal path.** `("Mavi (3)", "A Completely Different Album")` alone against query `("Mavi", "The Pilot")` returns `None`; with the real candidate present, 37193856 wins. The adversarial `Mavi (12)` / `Mavi (3)` pair resolves correctly whenever the albums differ.
- **No-regression control holds exactly.** For an unsuffixed credit the stripped form equals the raw form, so the max is unchanged at 100 — behavior is bit-identical, only recomputed (see finding 5).
- **Near-miss names still rejected.** `score_match("Mavi", "Davi")` = 75.0, below the floor; a `("Davi (12)", "The Pilot")` candidate does not match.
- **Degenerate inputs are safe.** `None` and empty variants are filtered before scoring; a pathological credit of `"(12)"` strips to `""` and is dropped by `find_best_typed_match`'s own empty-variant guard rather than scoring 100 against anything.

### Notes

- The live-API failure this branch hit (`test_mavi_disambiguation_suffix_resolves` resolving 35940307 rather than 37193856) was a correct resolution, not a bug: both pressings share master `4206252`, carry identical tracklists including both victim tracks, and differ only as the 2025-11-25 digital release vs the 2026-02-27 vinyl LP. Fixed by pinning the acceptable set, matching the file's two existing precedents. Worth noting the module self-skips without `DISCOGS_TOKEN`, so that class of failure is only ever reachable in CI.
- `lookup/strategies/library_miss.py` is now 193 lines against the 200-line budget in `tests/unit/test_module_budgets.py:236` — 7 lines of headroom.
- The fix is one-directional: a *suffixed* query against a *bare* candidate still floor-fails. Fine given queries originate from DJ input, but the docstring reads as though symmetry were restored.

jakebromberg:
## Closing out the "raw credit propagates outward" finding, for the Backend-Service consumer at least

The review flagged that `artist=best.artist` carries the raw `"Mavi (12)"` credit into the response, and asked whether a rescued playcut would end up displaying a Discogs-internal disambiguator as the DJ-visible artist name. Checked against the consumer these victim rows actually flow through, and **it does not materialize there**.

`Backend-Service`'s enrichment worker (`apps/enrichment-worker/enrich.ts`) writes only these fields from an LML response: `artwork_url`, `discogs_url`, `release_year`, the four streaming URLs, `artist_bio`, and `artist_wikipedia_url`. It never writes an artist *name* from the response into `album_metadata` or `flowsheet` — the DJ-visible artist stays the flowsheet row's own `artist_name`, which for the two victim rows (#5308909, #5308853) comes from rotation autofill and already reads `MAVI`.

Backend-Service is in fact already defensive about precisely this suffix on its own side: `shared/lml-client/src/index.ts:1658` does `release.artist.toLowerCase().split('(')[0].trim()` before comparing.

So for the flowsheet enrichment path this ticket is about, the finding is not a merge blocker and not a product decision — it is a no-op. **Scoping that claim honestly:** this was verified for the Backend-Service enrichment consumer only. `request-o-matic` also consumes `LookupResponse` and posts to Slack, and I did not check whether it surfaces `artist` directly. If it does, the cosmetic concern stands for that surface and is worth its own ticket — but it is pre-existing there too, since any lookup that already resolved a suffixed credit by its exact name has always returned the raw form.

The other findings are untouched by this and still stand on their own merits — in particular the docstring that asserts a title-axis confirmation the self-titled path does not provide, which is the one unambiguous defect in the set.

jakebromberg:
## Response to the review, in commit 5b567b6

**1. HIGH — self-titled docstring claim.** Fixed. `artist_variants_with_stripped_suffix`'s docstring (moved to `lookup/matching.py`, see below) no longer claims the title axis independently confirms every match — it now says plainly that the confirmation is NOT independent on the caller's self-titled path (`if is_self_titled(album): album = artist`), names the collapse as pre-existing (the `Bright`/`Bright` control from the review), and points at the follow-up. Filed WXYC/library-metadata-lookup#1208 for the collapse itself, scoped to the library-miss path with a note that `lookup/artwork.py` carries the identical shape.

**2. MEDIUM — ranking flip.** Fixed with a new `artist_variant_tie_break_key(query_artist, result)` in `lookup/matching.py`, wired as `_floor_best`'s `key_fn` ahead of the existing LML#1097 release_id ordering. It returns `(0, release_id)` when the candidate's raw credit alone already reaches its widened max score, `(1, release_id)` when only the suffix-stripped widening got it there — so an exact bare-name credit now always outranks a suffix-widened one at equal combined score, and release_id only breaks a tie within each group. Added `test_exact_credit_wins_tie_over_suffix_stripped_widening` (`tests/unit/test_library_miss_discogs.py`) covering exactly the `900`/`100` same-album collision from the review, plus direct unit coverage on the key function itself in `tests/unit/test_matching.py::TestArtistVariantTieBreakKey`. Confirmed the new test failed red against pre-fix code (picked `100` instead of `900`) before implementing.

**3. LOW — no dedup.** Fixed: `artist_variants_with_stripped_suffix` now returns `list(dict.fromkeys([*raw_variants, *stripped_variants]))`, so an unsuffixed credit (the common case) produces one variant instead of two identical ones. Covered by `tests/unit/test_matching.py::TestArtistVariantsWithStrippedSuffix`.

**4. MEDIUM — `lookup/artwork.py` gap.** Filed as a follow-up rather than folded in: WXYC/library-metadata-lookup#1209. Two reasons, both from the brief rather than a new call: the fix needs its own before/after measurement the way #1206 needed one (same restraint the PR description already applied to the shared `score_match`/`find_best_typed_match`), and `lookup/artwork.py` is a frozen surface for the Post-launch service hardening initiative's duration ([org project #32](https://github.com/orgs/WXYC/projects/32)) — a drive-by change there doesn't belong on this PR regardless of size. The issue also folds in the correction that the "Numeric disambigs clear via `token_sort_ratio`'s tolerance" comment is right for the compilation case it describes (`score_match('Various Artists', 'Various Artists (2)') = 88.2`), and carries forward the shared-scorer sweep question (Bandcamp/Apple Music/YouTube Music/Wikipedia) so it isn't lost.

**Blast-radius note (length-dependence table).** Added as context in the moved function's docstring rather than as a standalone claim — the docstring now states the widening's effect is scoped to short/collision-prone credits via the mechanism, without re-deriving the specific score table (that lives in this PR's history and in #1206/#1209 for reference).

Housekeeping: the variant-widening + tie-break helpers moved from `lookup/strategies/library_miss.py` into `lookup/matching.py` (its sole caller stays `_library_miss_discogs_search`) — partly because that module is where the rest of this file's pure leaf predicates already live, partly because the fixes above would have pushed `library_miss.py` past its 200-line module budget. `library_miss.py` is now 165 lines; `lookup/matching.py` is 459 against its 550 budget.

Verification: `uv run ruff check .` and `uv run ruff format --check .` both clean; `uv run pytest -n auto -q` → 6429 passed, 1 skipped, 2 xfailed (baseline 6422 + 7 new tests). CI green on push, including the live Discogs `test_mavi_disambiguation_suffix_resolves` companion.

